PR #2, opened 2021-07-19
EVM storage and StarkNet testing

BitBaseBit:
This PR adds evm storage and StarkNet testing. The StarkNet test implementation can be found in scripts/starknet. It works as follows:
First we gather all the opcode tests into two opcode files. We then transpile them into two StarkNet contracts. During the transpilation phase we create a json storage table file for each opcode file, these two storage table files contain a json object where the key is the name of the opcode test, and the value is another object that contains the storage address it's result is stored in , and the expected result, here's an example:
```
{
    "1_sub.opcode": {
        "storage_addr_low": 0,
        "gold_res_low": 78, 
        "gold_res_high": 0
    },  
    "1_too-big-for-uint128.opcode": {
        "storage_addr_low": 1,
        "gold_res_low": 113427455640312821154458202477256070485,
        "gold_res_high": 1
    },  
    "1_xor.opcode": {
        "storage_addr_low": 2,
        "gold_res_low": 49, 
        "gold_res_high": 0
    },  
    "add.opcode": {
        "storage_addr_low": 3,
        "gold_res_low": 21, 
        "gold_res_high": 0
    }
}
```
After compiling those contracts, we deploy both of them concurrently and wait for the deployment transactions for each contract to be ACCEPTED_ONCHAIN. Once that happens, we invoke the only external function called `main`, in both contracts. This function executes the opcode tests and stores the result in a storage mapping. Again this function is invoked concurrently for both contracts, and waited on to be ACCEPTED_ONCHAIN. Once that happens, we concurrently invoke the view function for each contract, getting the values at all the aforementioned storage addresses at once. This result is them compared with the expected result.

You can run the test with `python scripts/starknet/starknet_test.py GATEWAY_URL`


scripts/starknet/testing_utils.py
```diff
@@ -0,0 +1,158 @@
...
+def get_compiled_contracts(contract_list):
+    a = []
+    for contract in contract_list:
+        a.append(f"{contract[:-6]}_compiled.json")
+    return a
+
+
+def get_abis(contract_list):
+    a = []
+    count = 1
```
murcake, line 154: Seems to be unused

src/EvmToCairo.py
```diff
@@ -38,15 +39,14 @@
    local stack0 : StackItem
    assert stack0 = StackItem(value=Uint256(-1, 0), next=&stack0)  # Points to itself.
 
-   let (local res) = run_from{{msize=msize, memory_dict=memory_dict}}(Uint256(0, 0), &stack0)
-   default_dict_finalize(memory_start, memory_dict, 0)
```
murcake, line 42: Why remove dict finalization?

src/Operations/Memory.py
```diff
@@ -63,7 +65,7 @@ def access_width(cls):
     def _do_memory_access(self, address, state):
         value = state.stack.pop()
         return [
-            f"let (byte, _) = extract_lowest_byte({value})",
+            f"let (local byte, _) = extract_lowest_byte({value})",
             f"mstore8(offset={address}, byte=byte)",
```
murcake, line 69: Does `byte` need to be `local`? It's used only once on the second line and can't be revoked

murcake (commented):
I understand that it might be hard to do at that point, but if possible, I would love to see storage, transpile-time evaluation and starknet-testing in separate MRs

murcake:
It is still a WIP, so sorry if I was carping to obvious things that are just not yet cleaned up

src/EvmToCairo.py
BitBaseBit: It doesn't work on StarkNet

scripts/starknet/testing_utils.py
BitBaseBit: It's being used

scripts/starknet/starknet_test.py
```diff
@@ -0,0 +1,452 @@
...
+
+def get_abis(contract_list):
+    a = []
+    count = 1
+    for contract in contract_list:
+        a.append(os.path.join(ARTIFACTS_DIR, f"{contract[:-6]}_abi.json"))
+    return a
+
+
+def starknet_compile(contracts_list):
```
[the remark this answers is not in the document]
BitBaseBit: Good point, I'll update

scripts/starknet/testing_utils.py
```diff
@@ -0,0 +1,158 @@
+import os
+import sys
+import math
+import ntpath
```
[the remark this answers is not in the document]
BitBaseBit: Oops, that's not meant to be there.

src/Operations/Memory.py
BitBaseBit: It prevents range_check_ptr from being revoked.

scripts/starknet/testing_utils.py
murcake: I might be overlooking something, but this function is 5 lines long and I can't find any mention of `count` anywhere

src/EvmToCairo.py
murcake: I think, Shahar has mentioned that they've whitelisted dict hints, so it should work now?..

BitBaseBit: default_dict_finalize isn't whitelisted yet.

PR #3, opened 2021-07-20
Implement return operation

murcake:
This PR instroduces the `RETURN` operation. It adds an additional return result to each `segment` and the `run_to` function. It's called an `output`. Per EVM specification, as soon as the output is requested, the contract should return a slice of memory and the execution should stop. 

The output is implemented as a 128-bit packed big-endian felt-array that is serialized to the output (for now) to enable verification of it.

Other changes include tracking test files to properly regenerate bats.


src/Operations/Halting.py
```diff
@@ -1,13 +1,16 @@
...
 class Stop(Operation):
     def proceed(self, state):
-        return ["if 1 == 1:", "return (stack, Uint256(-1,0))", "end"]
+        state.unreachable = True
+        return state.make_return_instructions(Uint256(0), EMPTY_OUTPUT)
 
 
 """
-In the future, when StarkWare has implemented
+In the future, when StarkWare has implementent
```
Scooletz, line 13:
```suggestion
In the future, when StarkWare implements
```

might be better

tests/transpilation/and.opcode.gold.cairo
```diff
@@ -8,26 +9,28 @@ from starkware.cairo.common.serialize import serialize_word
...
+        if output.active == 1:
+            return (stack, output)
+        end
         return run_from(evm_pc=evm_pc, stack=stack)
     end
     let (immediate) = uint256_eq(evm_pc, Uint256(-1, 0))
     if immediate == 1:
-        let stack0 : StackItem* = stack
-        return (res=stack0.value)
+        return (stack, Output(1, cast(0, felt*), 0))
```
Scooletz, line 33: I do understand what is happening in here, that it provides a null array for input, still `cast (0, felt*)`. Still have some kind of mental friction to read it as a `null`.

Scooletz (commented):
I was going through the changes and I noticed that a lot of them is for bringing Cairo `0.3.0` on the board. I can see that it's a single commit. Would it be possible to split these two and have a PR for 0.3 first?

Is the description of the PR full? It looks like it's missing some words maybe?

The approach with a flag and a check makes it easy to reason about and to follow tests. 👍

cairo-src/evm/utils.cairo
```diff
@@ -1,5 +1,6 @@
 from starkware.cairo.common.math import unsigned_div_rem
 from starkware.cairo.common.math_cmp import is_le
+from starkware.cairo.common.serialize import serialize_word
```
BitBaseBit, line 3: Why are we importing serialize_word here? I'm pretty sure it doesn't exist on StarkNet

murcake: That's for the time being, while we still aren't quite on starknet (although close). I need it to test that the output is correct.

murcake:
@Scooletz 

> I was going through the changes and I noticed that a lot of them is for bringing Cairo `0.3.0` on the board. I can see that it's a single commit. Would it be possible to split these two and have a PR for 0.3 first?

Will do

> Is the description of the PR full? It looks like it's missing some words maybe?

I think it is, but my English skills are far from perfect =). What makes it seem like something is missing? I'll try to fix it

cairo-src/evm/utils.cairo
BitBaseBit: Ah yes, makes sense.

tests/transpilation/and.opcode.gold.cairo
Scooletz: I think I have none 😞  This is probably language related comment.

BitBaseBit:
LGTM

Scooletz:
@murcake I amended the description, if needed GH keeps previous versions of it so we can revert if needed.

murcake:
> @murcake I amended the description, if needed GH keeps previous versions of it so we can revert if needed.

Ah, I see. It's just that I copy descriptions from my editor which I ask to trim lines at ~80-90 symbols. I'll untrim them for MR descriptions in the future

PR #7, opened 2021-07-23
Log operations

cicr99:
Implemented log opcodes behavior regarding:
- popping things from stack
- updating the memory consumption counter


tests/transpilation/log_zero_length.opcode
```diff
@@ -0,0 +1,4 @@
+PUSH1 0x00   # [0]
+PUSH1 0x80   # [128, 0]
```
murcake, line 2: Cool, it's a nice idea to test the corner case

murcake (approved):
LGTM

cairo-src/evm/utils.cairo
```diff
@@ -25,6 +25,15 @@ func ceil_div{range_check_ptr}(a, b) -> (res):
     end
 end
 
+func memory_expansion_for_range{range_check_ptr}(s, f, l) -> (res):
```
murcake, line 28: Perhaps, `update_msize` is a more suitable name? Together with more descriptive parameter names, it would make the function crystally clear

PR #8, opened 2021-07-23
add `request_fresh_name` method for state

murcake:
the method guarantees to return a new unique name every time it's
    called (on the same object)

Removes the need to remember to increase `n_locals` by one each time.


BitBaseBit (commented):
LGTM

PR #9, opened 2021-07-26
added immediate evaluation when possible

BitBaseBit:
Often, when processing an opcode, the values that the opcodes are available as immediates, or as temporary variables from previous calculations (within the same segment of course). This means that instead of doing all the heavy uint256 math in Cairo, we can just evaluate the operation in python and assign the result to a constant variable in Cairo.


src/Operations/Bitwise.py
```diff
@@ -1,74 +1,177 @@
...
+                f"local {res} : Uint256 = Uint256({res_low}, {res_high})",
+                True,
+                evaluated,
+            )
+        else:
+            return (
+                f"let (local {res} : Uint256) = uint256_and({op1}, {op2})",
+                False,
+                0,
+            )
```
murcake, line 23: It seems like almost all of the changes related to operation classes follow the same pattern. The only thing that differs is the evaluation function in python and the evaluation function in cairo. And there is a lot of such classes. Perhaps it would make sense to extract the common logic in the parent class?

murcake (commented):
I don't quite understand why the `evaluated_locals` is needed. If we can fully evaluate some computation, we don't even need to create a reference. We can just push the value. To check that evaluation is possible, we just check whether all operands are `Uint256`. It cuts locals usage even further!
P.S. now that I think about it, the comment about locals usage is probably irrelevant, because `ap` would advance just as far

BitBaseBit:
> I don't quite understand why the `evaluated_locals` is needed. If we can fully evaluate some computation, we don't even need to create a reference. We can just push the value. To check that evaluation is possible, we just check whether all operands are `Uint256`. It cuts locals usage even further!
> P.S. now that I think about it, the comment about locals usage is probably irrelevant, because `ap` would advance just as far

Ah yes this is a very good point! It will drastically cut down locals usage

src/Operations/Binary.py
```diff
@@ -1,16 +1,40 @@
...
+            return [instruction]
+        else:
+            state.stack.push_ref(res_ref_name)
+            return [instruction]
+
+    def can_eval(self, op1, op2) -> (str, str, bool):
+        op1_imm = isinstance(op1, Uint256)
+        op2_imm = isinstance(op2, Uint256)
+        if isinstance(op1, str) or isinstance(op2, str):
+            return op1, op2, False
```
murcake, line 31: I think this `if` is unneeded, you check for it in the next line

PR #10, opened 2021-07-27
Sha3 operation

natdurlik:
Implemented sha3 operation.


tests/transpilation/sha3.opcode
```diff
@@ -0,0 +1,3 @@
+PUSH1 0x20
+PUSH1 0x0
+SHA3
```
murcake, line 3: Is it true that you compute keccak of 0 here?

src/Operations/SHA3.py
```diff
@@ -0,0 +1,43 @@
...
+                let (local msize) = update_msize(msize, {offset}, {length})
+                let (local memval) = create_from_memory({offset}, {length})
+                local high
+                local low
+                %{{
+                    from Crypto.Hash import keccak
+
+                    keccak_hash = keccak.new(digest_bits=256)
+                    arr = []
+                    arr_length = {length}//16 if {length} % 16 == 0 else {length}//16+1
```
murcake, line 24: Just a note: I don't know if you consider it more readable, but ceil division can be expressed as `({length} - 1) // 16 + 1`

murcake (commented):
LGTM. I would suggest though to squash all commits into one, since they basically fix each other

tests/transpilation/sha3.opcode
natdurlik: Yes

PR #11, opened 2021-07-28
Enforce stack manipulation

murcake:
Many operations could benefit from automatic stack parameters and outputs handling. It makes their implementation more declarative and much shorter, and reduces chances to make a mistake.

Furthermore, this PR updates golden tests to work with starknet contracts. Running is out of question for now, but we do check them against golden files and check that they compile.


BitBaseBit:
Looks good to me

PR #12, opened 2021-07-28
Implemented the warp cli app and direct transpliation

BitBaseBit:
The warp cli can be installed by running `python setup.py install`. It can then be used to transpile a solidity/vyper contract to Cairo with `warp transpile contract.sol`.


warp/cli/opcode_tables/op_to_str.json
```diff
@@ -0,0 +1,147 @@
...
+    "57": "JUMPI",
+    "58": "PC",
+    "59": "MSIZE",
+    "09": "MULMOD",
+    "5a": "GAS",
+    "5b": "JUMPDEST",
+    "60": "PUSH1",
+    "61": "PUSH2",
+    "62": "PUSH3",
+    "63": "PUSH4"
```
murcake, line 146: Would it be a bad idea to not move this whole table into a python file as a dictionary, and just import it whenever we need it? It seems like it would reduce the need for that big go utility and the two json files

PR #13, opened 2021-07-29
implement CALLDATA{COPY,SIZE,LOAD}

murcake:
implement CALLDATA{COPY,SIZE,LOAD}

the implementation simply reads usual 128-bit packed big-endian felts
    from array to memory/whatever

this also introduces execution environment to all cairo segments, for
     all call related and constant info (as per yellow paper)

----------

There is a bit of a problem with the joint testing approach. When we
     join all tests together, they share memory. It means that memory
     parts that we previously assumed to be zeroed out, are
     not. Furthermore, MSIZE tests also lose relevance.

For now, I've just deleted MSIZE tests and moved memory accesses far
     apart from each other in various tests. However, since network
     latency is apparently very low now, maybe we can again return to
     separate tests? Transactions have ability to be deployed in
     paralell, so it would also cut down waiting times. Perhaps, we
     can discuss doing something like that in the future. A local
     testing environment would be really handy, hopefully we'll get it
     at some point.


BitBaseBit:
LGTM

cicr99 (approved):
Seems fine to me

src/EvmToCairo.py
```diff
@@ -16,21 +16,31 @@
...
+    "evm.stack": {"StackItem"},
     "evm.output": {"Output"},
+    "evm.exec_env": {"ExecutionEnvironment"},
 }
 
-MAIN_BODY = f"""
+MAIN = """
+@external
+func main{storage_ptr: Storage*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+   unused_bits, payload_len, payload: felt*,
```
murcake, line 34: Rename `unused_bits` to `unused_bytes`

src/EvmToCairo.py
```diff
@@ -16,21 +16,31 @@
...
+   ):
    alloc_locals
    let (local __fp__, _) = get_fp_and_pc()
 
+   local exec_env: ExecutionEnvironment = ExecutionEnvironment(
+      payload_len = payload_len,
+      payload = payload,
+   )
```
murcake, line 42: Change to `input_len = payload_len * 16 - unused_bytes, input=payload`

PR #23, opened 2021-08-11
Binary Exponentiation method for Exp Operation

cicr99:
The `Exp` class was using the `pow` operator to do the evaluation, but this could be very inefficient if the numbers are too large. Using Binary Exponentiation algorithm fixes that problem.


murcake (approved):
Good!

PR #25, opened 2021-08-16
Implement shahars suggestions

murcake:


scripts/utils/mload.py
```diff
@@ -52,15 +56,14 @@ def parse_memory(mem: str) -> dict:
             i += 1
         elif mode == "m8":
             k = int(words[i])
-            sv = words[i + 1]
-            v = int(sv, 16) if sv.startswith("0x") else int(sv)
+            v = read_int(sv, words[i + 1])
```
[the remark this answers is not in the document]
murcake: Oh, right, it's a bug

BitBaseBit (commented):
There is something in here that breaks the ERC20 example, where we are always hitting an `assert 0 = 1` when we invoke any contract function.

warp/transpiler/Operations/Bitwise.py
```diff
@@ -32,44 +32,47 @@ def __init__(self):
...
 
+    @classmethod
     def required_imports(cls):
         return {UINT256_MODULE: {"uint256_shr"}}
 
 
-def sar(a, b):
-    b = uint256_to_int256(b)  # per yellow paper, only b is treated as signed
-    return int256_to_uint256(a >> b)
+def sar(value, shift):
```
natdurlik, line 59: Why isn't sar function in Sar class?

murcake: It doesn't need the class state and semantically it is just an operation on two numbers

murcake:
> There is something in here that breaks the ERC20 example, where we are always hitting an `assert 0 = 1` when we invoke any contract function.

@BitBaseBit it should work now. The issue was that `is_gt` worked as "<"

BitBaseBit:
> > There is something in here that breaks the ERC20 example, where we are always hitting an `assert 0 = 1` when we invoke any contract function.
> 
> @BitBaseBit it should work now. The issue was that `is_gt` worked as "<"

But the GT opcode should work as "<" as per the yellow paper no?

murcake:
> > > There is something in here that breaks the ERC20 example, where we are always hitting an `assert 0 = 1` when we invoke any contract function.
> > 
> > 
> > @BitBaseBit it should work now. The issue was that `is_gt` worked as "<"
> 
> But the GT opcode should work as "<" as per the yellow paper no?

First of all, I misspelled, it used to work as "<=". Second, "<" is a less than sign. The yellow paper specifies that GT checks that the top of the stack is greater than the item second from the top of the stack. So, the way that we use `is_gt` means that it should work as ">", not as "<".

BitBaseBit:
> > > > There is something in here that breaks the ERC20 example, where we are always hitting an `assert 0 = 1` when we invoke any contract function.
> > > 
> > > 
> > > @BitBaseBit it should work now. The issue was that `is_gt` worked as "<"
> > 
> > 
> > But the GT opcode should work as "<" as per the yellow paper no?
> 
> First of all, I misspelled, it used to work as "<=". Second, "<" is a less than sign. The yellow paper specifies that GT checks that the top of the stack is greater than the item second from the top of the stack. So, the way that we use `is_gt` means that it should work as ">", not as "<".

First of all, that makes sense, I thought you were referring to strict comparison. Second, "<" alone is neither less than nor greater than, i.e a < b or b < a.

BitBaseBit (approved):
LGTM

PR #31, opened 2021-09-03
add translation from yul builtin -> cairo library function name

BitBaseBit:


murcake (approved):
It feels to me like this change would fit better in the MangleNamesVisitor, since it also has to do with names' adaptation.

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,218 @@
...
+    "uint256_mul",
+    "uint256_unsigned_div_rem",
+    "uint256_signed_div_rem",
+]
 
 class ToCairoVisitor(AstVisitor):
     def __init__(self):
         super().__init__()
         self.repr_stack: list[str] = []
+        self.preabmle: bool = False
```
cicr99, line 38: this is just a typo

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,218 @@
...
+                    variables.append(var_repr)
+                else:
+                    var_repr += ": Uint256"
+                    variables.append(var_repr)
+            else:
+                if function_name in RETURNS_TWO:
+                    var_repr += ", _"
+                    variables.append(var_repr)
+                else:
+                    variables.append(var_repr)
```
cicr99, line 83:
```suggestion
            if "Uint256" not in var_repr:
                var_repr += ": Uint256"
            if function_name in RETURNS_TWO:
                var_repr += ", _"
            variables.append(var_repr)
```

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,209 @@
...
+                var_repr += ": Uint256"
+            if function_name in RETURNS_TWO:
+                var_repr += ", _"
+            variables.append(var_repr)
+        return ", ".join("local " + x for x in variables)
+
+    def visit_assignment(self, node: ast.Assignment) -> str:
+        self.preamble = False
+        value_repr: str = self.print(node.value)
+        function_name: str = value_repr[: value_repr.find("(")].strip()
```
murcake, line 80: We can get `function_name` without text manipulation. `function_name = node.value.function_name.name`. It should be done in the if-branch where we make sure that `isinstance(node.value, ast.FunctionCall)`.

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,209 @@
...
-            if not node.variable_names:
-                self.repr_stack.append(value_repr)
+            function_args: str = value_repr[
+                value_repr.find("(") + 1 : value_repr.rfind(")")
+            ]
+            if function_name in YUL_BUILTINS_MAP.keys():
+                builtin_to_cairo = YUL_BUILTINS_MAP[function_name](function_args)
+                merge_imports(self.imports, builtin_to_cairo.required_imports())
+                if not node.variable_names:
+                    self.preamble = True
```
murcake, line 90: `self.preamble` seems to be set to `True` in all branches of the outer-most if. Can it be moved up and set to True once?

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,209 @@
...
+
+    def visit_assignment(self, node: ast.Assignment) -> str:
+        self.preamble = False
+        value_repr: str = self.print(node.value)
+        function_name: str = value_repr[: value_repr.find("(")].strip()
+        ids_repr: str = self.generate_ids_typed(node.variable_names, function_name)
         if isinstance(node.value, ast.FunctionCall):
-            if not node.variable_names:
-                self.repr_stack.append(value_repr)
+            function_args: str = value_repr[
```
murcake, line 83: It also can be done without text manipulation, we have an ability to structurally introspect AST nodes and their properties without parsing text.

examples/ERC20/ERC20.sol
```diff
@@ -4,24 +4,24 @@ contract WARP {
     uint8  public decimals    = 18;
     uint256 public totalSupply= 100000000000000000000000000000000000;
 
-    mapping (address => uint)                       public  balanceOf;
-    mapping (address => mapping (address => uint))  public  allowance;
+    mapping (uint => uint)                       public  balanceOf;
```
murcake, line 7: Just curious, how does it help?

warp/yul/BuiltinHandler.py
```diff
@@ -0,0 +1,470 @@
...
+            function_name="uint256_shl",
+            function_args=function_args,
+        )
+
+
+class Shr(BuiltinHandler):  # ARG ORDER NOT CONVENTIONAL
+    def __init__(self, function_args: str):
+        super().__init__(
+            module=UINT256_MODULE,
+            function_name="uint256_shl",
```
murcake, line 175: Typo: should be "shr"

murcake (commented):
(the issue might be that i'm using old `gen-yul-json-ast`)
For some reason, some functions in if-expressions end up on a newline. That breaks parsing: 

`python yul/main.py ERC20.sol` results in 

``` 
starkware.cairo.lang.compiler.parser_transformer.ParserError: <string>:518:4: Unexpected token Token(_NEWLINE, '\n'). Expected: expression.
if 
   ^```

File: warp/yul/BuiltinHandler.py [hunk 8]
```diff
@@ -0,0 +1,476 @@
...
+            f"local memory_dict : DictAccess* = memory_dict"
+            f"local msize = msize"
+            f"let (local byte, _) = extract_lowest_byte({self.value})",
+            ref_copy="local memory_dict : DictAccess* = memory_dict",
+        )
+
+    def generate_cairo(self):
+        return f"""
+{self.preamble}
+mstore8(offset={self.address}, byte=byte)
```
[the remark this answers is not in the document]
BitBaseBit: byte is in the preamble

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,209 @@
...
+    "starkware.starknet.common.storage": {"Storage"},
+    "evm.utils": {"update_msize"},
+}
+
+RETURNS_TWO = [
+    "uint256_add",
+    "uint256_mul",
+    "uint256_unsigned_div_rem",
+    "uint256_signed_div_rem",
+]
```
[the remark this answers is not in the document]
BitBaseBit: Yep will do

examples/ERC20/ERC20.sol
BitBaseBit: It removes the code that does the unnecessary shifting to get the first 20 bytes.

warp/yul/BuiltinHandler.py
```diff
@@ -0,0 +1,476 @@
...
+        self.function_name = function_name
+        self.function_args = function_args
+        self.preamble = preamble
+        self.ref_copy = ref_copy
+        self.function_call = f"{self.function_name}({self.function_args})"
+        self.generated_cairo = self.generate_cairo()
+
+    def required_imports(self):
+        if self.module == "":
+            return {"": {""}}
```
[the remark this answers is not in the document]
BitBaseBit: Yeah, done

warp/yul/main.py
```diff
@@ -14,23 +14,23 @@
 from yul.ToCairoVisitor import ToCairoVisitor
 from yul.parse import parse_node
 
-AST_GENERATOR = "gen-yul-json-ast"
+AST_GENERATOR = "./gen-yul-json-ast"
```
murcake, line 17: Maybe we should rather rely on the executable being in PATH?

warp/yul/main.py
```diff
@@ -14,23 +14,23 @@
...
 
 MAIN_PREAMBLE = """%lang starknet
 %builtins pedersen range_check
 """
 
 
 def main(argv):
-    if len(argv) != 2:
+    if len(argv) != 3:
         sys.exit(f"Usage: python {argv[0]} SOLIDITY-FILE")
```
murcake, line 26: Could you please also update the usage message?

warp/yul/BuiltinHandler.py
```diff
@@ -356,17 +357,17 @@ def required_imports(self):
...
-        self.value: str = function_args.split(",")[1].strip()
+        self.offset: str = get_low_bits(function_args.split(",")[0].strip())
         super().__init__(
-            module="evm.uint256",
-            function_name="uint256_signextend",
+            module="evm.memory",
+            function_name="mload",
             function_args=function_args,
-            preamble=f"let (local msize) = update_msize{{range_check_ptr=range_check_ptr}}(msize, {self.address}, 1)"
+            preamble=f"let (local msize) = update_msize{{range_check_ptr=range_check_ptr}}(msize, {self.address}, 32)"
```
murcake, line 365: `self.address` should be `self.offset`, it fails otherwise

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,195 @@
...
+            self.preamble = True
+            return f"{vars_repr} = {value_repr}"
 
-    def visit_block(self, node: ast.Block):
-        stmts_repr = "\n".join(self.print(x) for x in node.statements)
-        self.repr_stack.append(stmts_repr)
+    def visit_block(self, node: ast.Block) -> str:
+        try:
+            stmts_repr = "\n".join(self.print(x) for x in node.statements)
+        except TypeError:
```
murcake, line 165: Why do we catch `TypeError` here? If something throws `TypeError` it is the underlying issue that should be solved.

In that case, I have a few suggestions:
- `visit_function_definition` instead of returning appends code to `cairo_code`. You can have it return, and just remove self.cairo_code, it's not used anywhere else
- `visit_break/continue/leave` don't return anything. For now, they can return an empty string
- something else, I don't quite know what yet

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,195 @@
...
-            "alloc_locals\n"
-            f"{body_repr}\n"
-            f"return ({return_names})\n"
-            "end\n"
-        )
-
-    def visit_if(self, node: ast.If):
+        if "abi_" in node.name or "checked_" in node.name:
+            return
+        self.cairo_code += f"""
```
murcake, line 176: You can just return here

warp/yul/ToCairoVisitor.py
```diff
@@ -1,114 +1,195 @@
...
+                if not node.variables:
+                    self.preamble = True
+                    return value_repr
+                else:
+                    self.preamble = True
+                    return f"let ({vars_repr}) = {value_repr}"
         else:
-            self.repr_stack.append(f"{vars_repr} = {value_repr}")
+            vars_repr = self.generate_ids_typed(node.variables)
+            self.preamble = True
```
murcake, line 159: This preamble is set to True in all branches, can't it be set once in a branch above?

warp/yul/ToCairoVisitor.py
BitBaseBit: I've tried that method but with the current structure this seems like the easiest way. Also , we only care about function definitions, because the only other top level declarations are storage variables, which we are working on getting using a different method.

BitBaseBit: The only reason for the typeError is because `visit_functino_definition` doesn't return anything.

murcake: Unfortunately, it masks many other kinds of type errors too. Not just when a visitor doesn't return. We've had this except block 

```python
       try:
            return getattr(self, method_name)(node, *args, **kwargs)
        except AttributeError:
            return self.common_visit(node, *args, **kwargs)
```

in yul_ast.py, and it made error discovery quite unintuitive in several cases during the short life of this project. So I had to get rid of it.

Blocks contain not only top level declarations but all possible statements of the program. So each block with `break` gets replaced by nothing.

I've applied my suggestions from the above, and I still got `None`s in the list. It might be another bug and another missing block.

PR #33, opened 2021-09-07
replace breaks with leaves

murcake:
After this transformation there are no for-loops, breaks or continues in a Yul AST. In all places, that would require a `return` in Cairo, Yul `leave`'s are placed


BitBaseBit (approved):
LGTM

PR #35, opened 2021-09-08
Yul testing

cicr99:
Implemented scripts for testing yul transpilation:
- `transpile_test.py` to compare the generated cairo code from a solidity contract with the expected cairo code for it.
- `compilation_test.py` to check the correct compilation of these cairo codes.

The steps for transpiling the solidity code to cairo code were moved into the function `sol_to_cairo` in another module different from `yul.main`, so this function could also be called from the scripts


Makefile
```diff
@@ -23,6 +23,11 @@ test_starknet: warp
 	python scripts/starknet/starknet_test.py "${STARKNET_ENV}"
 .PHONY: test_starknet
 
+test_yul:
```
murcake, line 26: It makes sense to also make target `test` depend on `test_yul`

murcake:
Can you please also add yul tests to the CI?

For me, the if-flattening tests doesn't pass. The reason is the same weird error with extra symbols at the end of `if-flattening.sol` on lines that don't even exit. I'm investigating this

murcake:
I just can't, it doesn't reproduce again

murcake:
Could you please, rebase onto main instead of merging if feasible and add the tests to the CI. Other than that, LGTM

BitBaseBit (approved):
LGTM

PR #36, opened 2021-09-08
split expressions

murcake:
split expressions to get rid of nested function calls

this closer models cairo, where functions can't be nested, and
     function calls are not allowed in if conditions (this PR takes
     care of this as well)


BitBaseBit (approved):
LGTM

PR #39, opened 2021-09-09
Replace yul caller() with get_caller_data_uint265()

swapnilraj:
Before this change generated cairo code for solidity contracts which used `msg.sender`, looked like this:
```
let (local a: Uint256) = caller()
```
After the changes:
```
from evm.calls import get_caller_address_uint256
# ...
let (local a: Uint256) = get_caller_address_uint256()
```

### DIFF
--- a/warp/yul/BuiltinHandler.py
+++ b/warp/yul/BuiltinHandler.py
@@ -431,6 +431,16 @@ class SHA3(BuiltinHandler):
     def required_imports(self):
         return {"evm.sha3": {"sha"}}
 
+# ============ Call Data ============
+class Caller(BuiltinHandler):
+    def __init__(self, function_args: str):
+        super().__init__(
+            function_name="get_caller_address",
+            function_args=function_args,
+        )
+
+    def required_imports(self):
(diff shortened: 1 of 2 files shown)


warp/yul/BuiltinHandler.py
```diff
@@ -431,6 +431,17 @@ def __init__(self, function_args: str):
     def required_imports(self):
         return {"evm.sha3": {"sha"}}
 
+# ============ Call Data ============
+class Caller(BuiltinHandler):
+    def __init__(self, function_args: str):
+        super().__init__(
+            module="",
```
murcake, line 438: I think you can specify module here as "evm.calls" and then you won't have to override `required_imports`

murcake (approved):

warp/yul/BuiltinHandler.py
swapnilraj: Yea that seems like a better way of doing it, I have updated the PR.

BitBaseBit (approved):
LGTM

PR #41, opened 2021-09-10
remove starknet tests of the old warp transpiler

murcake:
also remove from the Makefile

Apparently our test environment has been disabled. It couldn't have
     been used by warp users other than the nethermind development
     team.

Also, now StarkNet has an official way to do unit testing.


BitBaseBit (approved):
LGTM

PR #42, opened 2021-09-11
fix non compiling contracts

murcake:
Some contracts in the `main` didn't compile with the last version of gen-yul-json-ast, but did compile with an older version. The reason is that the last version disables inlining of functions (otherwise, public functions get inlined). Uninlined functions display the fact that we don't copy references on non-builtin functions.

There were several more issues with the transpiler, each one fixed in a separate commit.


warp/yul/ToCairoVisitor.py
```diff
@@ -118,88 +120,29 @@ def visit_literal(self, node: ast.Literal) -> str:
...
         args_repr = ", ".join(self.print(x) for x in node.arguments)
-        if fun_repr in self.public_functions:
-            args_repr = self.to_high_low_felt(args_repr)
         if fun_repr == "revert":
-            return "assert 0 = 1"
+            return "assert 0 = 1\njmp rel 0"
         elif fun_repr.startswith("checked_add"):
             merge_imports(self.imports, {"evm.uint256": {"u256_add"}})
             return f"u256_add({args_repr})"
-        elif fun_repr.startswith("checked_sub"):
```
BitBaseBit, line 188: Why are we no longer checking for checked_sub?

warp/yul/ToCairoVisitor.py
```diff
@@ -210,181 +153,65 @@ def visit_expression_statement(self, node: ast.ExpressionStatement) -> str:
...
+                stmt_reprs.append(self.visit(stmt))
+                if self.last_needs_ref_copy:
+                    stmt_reprs.append(gen_ref_copy())
+        return "\n".join(stmt_reprs)
 
     def visit_function_definition(self, node: ast.FunctionDefinition):
         self.last_function = node
-        self.set_ret_vars_len(node)
-        taboos = ["abi_", "checked_", "getter_", "panic_error"]
+        taboos = ["abi_", "checked_add", "getter_"]
```
BitBaseBit, line 183: Why did we remove panic_error

warp/yul/ToCairoVisitor.py
murcake: `checked_sub` doesn't act like `uint256_sub` sometimes. Sometimes, it takes only one argument and subtracts some hard-coded number from it. I suspect that it's also the case with `checked_add` in some cases but we'll see.

warp/yul/ToCairoVisitor.py
murcake: `RevertNormalizer` automatically removes all "reverting" functions and turns calls to them to reverts. "reverting" functions are functions such that each execution path of them leads to a revert

murcake: So, that automatically removes "panic_error"

BitBaseBit (approved):
LGTM

PR #43, opened 2021-09-13
Configure pre-commit-hook to format using black

swapnilraj:


swapnilraj:
Staged files should be re-formatted and committed instead of aborting the commit

PR #46, opened 2021-09-15
Implement transpilation for built-in calldatacopy

swapnilraj:


warp/yul/BuiltinHandler.py
```diff
@@ -412,6 +412,22 @@ def __init__(self, function_args: str):
         self.function_call = "Uint256(exec_env.calldata_size, 0)"
 
 
+class CallDataCopy(BuiltinHandler):
+    def __init__(self, function_args: str):
+        super().__init__(
+            module="evm.calls",
+            function_name="calldatacopy_",
+            function_args=function_args,
+            call_implicits=["range_check_ptr", "exec_env"],
```
murcake, line 421: Why are call implicits and used implicits different?

warp/yul/BuiltinHandler.py
```diff
@@ -412,6 +412,22 @@ def __init__(self, function_args: str):
...
+            used_implicits=("range_check_ptr", "exec_env", "memory_dict", "msize"),
+        )
+        self.function_call = (
+            f"calldatacopy_{{range_check_ptr=range_check_ptr, exec_env=exec_env}}("
+            f"{self.function_args})\n"
+            f"local exec_env : ExecutionEnvironment = exec_env"
```
murcake, line 427: Wouldn't this copy be performed automatically because "exec_env" is in the list of `used_implicits`?

warp/yul/ExecEnv.py
```diff
@@ -24,30 +24,26 @@ def __init__(self):
...
-        else:
-            for func in calledFuncs:
+    def traverse_callgraph(self, callgraph):
+        for k, v in callgraph.items():
+            stack = []
+            stack.append(k)
+
+            while len(stack):
+                func = stack[-1]
+                stack.pop()
```
murcake, line 34:
```suggestion
                func = stack.pop()
```

warp/yul/ExecEnv.py
```diff
@@ -24,30 +24,26 @@ def __init__(self):
...
+                    if callgraph[func]["calldataOp"]:
                         self.needs_exec_env.append(func)
-                        self.traverse_inner(self.callgraph[func]["calledFunctions"])
-                    except KeyError:
-                        continue
-            else:
-                continue
+                    self.visited.append(func)
+                    stack.extend(v["calledFunctions"])
+                except KeyError:
```
murcake, line 44: What does that `KeyError` refer to? `callgraph[func]`, `callgraph[func]["calldataOp"]` or `v["calledFunctions"]`? Would it be more clear to use `dict.get` instead of []?

murcake (commented):
Maybe it makes sense to also make a transaction test? Like the ERC20_test.py and ERC20_storage.py

murcake:
I feel that I in general don't quite understand the need for the `NeedsExececutionEnvironment` class. Am I correct that it marks all functions that need to be passed the `ExecEnv` implicit? How is it different from all other implicits? We have a mechanism for marking them introduced in a87191d0594f503a81201ef6a43b4df2a24cff3e. In short, we just have to add "exec_env" to `BuiltinHandler` operations that require it and functions that call them will have "exec_env" automatically

swapnilraj:
> I feel that I in general don't quite understand the need for the `NeedsExececutionEnvironment` class. Am I correct that it marks all functions that need to be passed the `ExecEnv` implicit? How is it different from all other implicits? We have a mechanism for marking them introduced in [a87191d](https://github.com/NethermindEth/warp/commit/a87191d0594f503a81201ef6a43b4df2a24cff3e). In short, we just have to add "exec_env" to `BuiltinHandler` operations that require it and functions that call them will have "exec_env" automatically

As far I understand, `ExecEnv` traverses the callgraph and marks the functions that use calldata builtins. We use this to generate a list of functions which need the implicit `ExecutionEnvironment`. I used it because `calldataload` is already using it, but it seems that your earlier commit should handle this case as long as it knows that calldata builtins need `ExecutionEnvironment`, am I right?

murcake:
> > I feel that I in general don't quite understand the need for the `NeedsExececutionEnvironment` class. Am I correct that it marks all functions that need to be passed the `ExecEnv` implicit? How is it different from all other implicits? We have a mechanism for marking them introduced in [a87191d](https://github.com/NethermindEth/warp/commit/a87191d0594f503a81201ef6a43b4df2a24cff3e). In short, we just have to add "exec_env" to `BuiltinHandler` operations that require it and functions that call them will have "exec_env" automatically
> 
> As far I understand, `ExecEnv` traverses the callgraph and marks the functions that use calldata builtins. We use this to generate a list of functions which need the implicit `ExecutionEnvironment`. I used it because `calldataload` is already using it, but it seems that your earlier commit should handle this case as long as it knows that calldata builtins need `ExecutionEnvironment`, am I right?

Yeah, if you add to `used_implicits` of appropriate builtins

warp/yul/ExecEnv.py
swapnilraj: The `KeyError` refers to `callgraph[func]` so it seems easier to use `KeyError` unless you know something equivalent to [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) from JavaScript for Python

murcake: Unfortunately, there isn't builtin optional chaining, AFAIK. 

There are some options though:
1. `callgraph.get(func, {}).get("calldataOp")` or 
2. Assign and check: `info = callgraph.get(func); if not info: continue`
3. If we were able to use assignment expressions from python3.8 `if info := callgraph.get(func)`
4. Catch `KeyError` as you did

All these options have a difference balance of terseness and clarity. It's your call)

swapnilraj:
> > > I feel that I in general don't quite understand the need for the `NeedsExececutionEnvironment` class. Am I correct that it marks all functions that need to be passed the `ExecEnv` implicit? How is it different from all other implicits? We have a mechanism for marking them introduced in [a87191d](https://github.com/NethermindEth/warp/commit/a87191d0594f503a81201ef6a43b4df2a24cff3e). In short, we just have to add "exec_env" to `BuiltinHandler` operations that require it and functions that call them will have "exec_env" automatically
> > 
> > 
> > As far I understand, `ExecEnv` traverses the callgraph and marks the functions that use calldata builtins. We use this to generate a list of functions which need the implicit `ExecutionEnvironment`. I used it because `calldataload` is already using it, but it seems that your earlier commit should handle this case as long as it knows that calldata builtins need `ExecutionEnvironment`, am I right?
> 
> Yeah, if you add `ExecutionEnvironment` to `used_implicits` of appropriate builtins

There is one more bit of functionality we use the list of exec_env functions. We use to see if an external function needs `ExecutionEnvironment` and in that case add `calldata_size, calldata_len, calldata` to the parameter list. I don't think the commit you mentioned can be used to implement this? Maybe it makes sense to refactor that in a different PR?

murcake:
> There is one more bit of functionality we use the list of exec_env functions. We use to see if an external function needs `ExecutionEnvironment` and in that case add `calldata_size, calldata_len, calldata` to the parameter list. I don't think the commit you mentioned can be used to implement this? Maybe it makes sense to refactor that in a different PR?

Isn't it simple checking if `exec_env` is in `functions_to_implicits[node.name]`?

warp/cairo-src/evm/calls.cairo
```diff
@@ -1,16 +1,31 @@
...
+    alloc_locals
+    let (local caller_address) = get_caller_address()
+    let (local high, local low) = split_felt(caller_address)
+    let caller_data = Uint256(low=low, high=high)
+
+    return (caller_data=caller_data)
+end
 
-  return (caller_data=caller_data)
+func calldatacopy_{
```
murcake, line 20: The naming with underscore for `mstore` and `mload` was introduced not to break compatibility with the EVM transpiler. Fortunately, it's not a concern here. Anyway, I'm just carping, it doesn't matter

murcake (commented):
If we are switching to the general mechanism of passing implicits, does it mean that we can remove ExecEnv?

swapnilraj:
> If we are switching to the general mechanism of passing implicits, does it mean that we can remove ExecEnv?

For this PR I think `ExecEnv` live although I have removed the dependence of translating calldatacopy on `ExecEnv`.
I plan to remove `ExecEnv` in a different PR, to make it easy to review.

PR #48, opened 2021-09-24
calldata implementation & cli integration

BitBaseBit:


warp/yul/BuiltinHandler.py
```diff
@@ -385,12 +385,41 @@ def __init__(self, function_args: str):
...
+        self.function_call = "Uint256(exec_env.calldata_size, 0)"
+
+
 YUL_BUILTINS_MAP = {
     "add": Add,
     "addmod": AddMod,
     "and": And,
     "byte": Byte,
     "caller": Caller,
+    "calldataload": CallDataLoad,
```
murcake, line 422: Why isn't calldatasize added?

warp/yul/ExecEnv.py
```diff
@@ -0,0 +1,233 @@
...
+"""
+
+
+class NeedsExececutionEnvironment(AstVisitor):
+    def __init__(self):
+        super().__init__()
+        self.external_functions: list[str] = []
+        self.imports = defaultdict(set)
+        self.last_function: Optional[ast.FunctionDefinition] = None
+        self.curr_func: str = []
```
murcake, line 39: Why does this class need all the fields above?

warp/yul/ExecEnv.py
```diff
@@ -0,0 +1,233 @@
+from __future__ import annotations
```
murcake, line 1: There is `CallGraphBuilder` (and `cleanup_callgraph`) in `top_sort.py`, did you look into whether it can be reused?

warp/yul/ExecEnv.py
```diff
@@ -0,0 +1,233 @@
...
+            return "assert 0 = 1\njmp rel 0"
+        if fun_repr == "revert":
+            return ""
+        if "return" in fun_repr:
+            return ""
+        if fun_repr == "checked_add_uint256":
+            fun_repr = "add"
+        if "callvalue" in fun_repr:
+            return "Uint256(0,0)"
+        result: str
```
murcake, line 118: It seems like a lot of string manipulation which is not needed, we never use string output of this class

warp/yul/ToCairoVisitor.py
```diff
@@ -38,24 +45,38 @@
...
     "syscall_ptr": "felt*",
+    "exec_env": "ExecutionEnvironment",
 }
 
 IMPLICITS_SET = set(IMPLICITS.keys())
 
 
 class ToCairoVisitor(AstVisitor):
-    def __init__(self, sol_source: str, public_functions: list[str]):
+    def __init__(self, sol_source: str, sol_src_path: str, main_contract: str):
```
murcake, line 56: This change seems to be unnecessary?.. We don't use `sol_src_path` and `main_contract` for anything other than computing `public_functions`

warp/yul/ToCairoVisitor.py
```diff
@@ -155,11 +181,9 @@ def visit_variable_declaration(self, node: ast.VariableDeclaration) -> str:
             return value_repr
         vars_repr = ", ".join(f"local {self.visit(x)}" for x in node.variables)
         if isinstance(node.value, ast.FunctionCall):
-            if (
-                value_repr == "Uint256(31597865,9284653)"
-                or value_repr == "Uint256(0,0)"
-            ):
-                return f"{vars_repr} = {value_repr}"
+            if "calldatasize" in node.value.function_name.name:
```
murcake, line 186: Shouldn't it be automatically translated via the `BultinHandler` mechanism? Why do we need to hardcode this?

murcake (commented):
Why is if-flattening.cairo deleted?

warp/yul/ToCairoVisitor.py
BitBaseBit: we write the name of main_contract to a file so we can use it later for cli actions like invoke.

warp/yul/ExecEnv.py
BitBaseBit: we use it for all the callvalue checks for non-payable functions

BitBaseBit:
> Why is if-flattening.cairo deleted?

It's still there.

murcake:
> > Why is if-flattening.cairo deleted?
> 
> It's still there.

Oh right, I got confused by the GH interface.

warp/yul/ExecEnv.py
murcake: Oh I see. Can't we just though have a flag `has_calldata_call` and set it when there is a function call that is "calldataload" or "calldatasize"? Set it to false at the beginning of each function. Then query at the end. It seems like it would eliminate duplication between this and ToCairoVisitor. It's your call though.

BitBaseBit: But we would still need to traverse the callgraph to see if a function needs exec_env, because it might call a function which calls a function that needs exec_env.

murcake: All functions that need calldata also get marked as calldata. It's the same as in RevertNormalizer, the order is guaranteed to be optimal, because the's been top-sorted

PR #50, opened 2021-09-29
Finalize the default dict

JorikSchellekens:
The default dict needs to be finalised before the end of the program, as explained here: https://www.notion.so/nethermind/bb52b38dd5244093a07d4a4fefda1c1a?v=9b578f69d11149c9a65c5cfbd06b6b61&p=e2317bbb3ace41b585d9eb89d645c553

The call to `default_dict_finalize` reliably revokes all references so I've had to store all implicits in local.


cicr99 (approved):
It seems good to me

PR #53, opened 2021-09-30
run black formatter check in CI instead of in a precommit hook

murcake:
To reformat everything locally run `black .` from the root directory
     of the project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,18 +12,23 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: install solc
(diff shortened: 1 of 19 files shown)


warp/starknet/compile.py
```diff
@@ -0,0 +1,47 @@
+from typing import List
```
cicr99, line 1: This file is no longer needed. It was deleted in the previous PR.

murcake: Oh sorry, must have been added by the automatic merge conflict resolution algorithm

BitBaseBit (approved):
LGTM

PR #58, opened 2021-10-04
Remove splitting of variable returns in external functions

swapnilraj:
With the new cairo-lang release (0.4.2), structs are supported as return variables for external functions.
This PR removes the splitting of return variable in external functions.

### DIFF
--- a/warp/yul/ToCairoVisitor.py
+++ b/warp/yul/ToCairoVisitor.py
@@ -106,12 +106,8 @@ class ToCairoVisitor(AstVisitor):
             f"Each node type should have a custom visit, but {type(node)} doesn't"
         )
 
-    def visit_typed_name(self, node: ast.TypedName, split: bool = False) -> str:
-        if not split:
-            return f"{node.name} : {node.type}"
-        assert node.type == "Uint256", "Can't split non Uin256 type"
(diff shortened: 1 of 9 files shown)


warp/yul/ToCairoVisitor.py
```diff
@@ -300,7 +300,7 @@ def visit_leave(self, _node: ast.Leave) -> str:
 
     def _make_external_function(self, node: ast.FunctionDefinition) -> str:
         params = ", ".join(self.print(x, split=False) for x in node.parameters)
-        returns = ", ".join(self.print(x, split=True) for x in node.return_variables)
+        returns = ", ".join(self.print(x, split=False) for x in node.return_variables)
```
murcake, line 303: I think the split parameter (and code connected to it) can now be removed from `visit_typed_name`

murcake (approved):

warp/yul/ToCairoVisitor.py
swapnilraj: Ack

murcake:
Change cairo-lang requirement in build.yml to 0.4.2

murcake:
Typo =) Will not be helpful when searching for version bumps through commits

PR #61, opened 2021-10-06
Annotate view and external functions

swapnilraj:
Use `stateMutability` information from the contract ABI to annotate functions as either `view` or `external`.


JorikSchellekens (changes requested):
Just need to make sure we don't expose private storage vars

warp/yul/ToCairoVisitor.py
```diff
@@ -391,6 +416,12 @@ def _try_make_storage_accessor_body(
         )
         return body
 
+    def _lookup_function_mutability(self, function_name: str) -> Optional[str]:
+        for f, mutability in self.function_mutabilities:
+            if f in function_name:
```
murcake, line 421: Why are we doing this?.. Is that because in yul some functions get appended "random" digits to the end? 
If so, it seems to be a dangerous way to handle it. Suppose, someone has a public function named "set". Now, every function that contains "set" (including variable setters) is public as well. 

Perhaps, you can use regexes, as in "storage_access.py". However, IMO it highlights once again the dangers and inconvenience of trying to transpile higher level concepts from yul

PR #67, opened 2021-10-12
Replace $ with _ in identifiers

swapnilraj:
Yul identifiers do not allow `$`, so we replace it with `_`


murcake:
> Could you please add an offending contract? I think that replacing "$" with "_" is a part of yul optimizations

Here it is: https://github.com/NethermindEth/solidity/blob/2f075d54d0c9626552c19a53c2d43a974db0a417/libyul/optimiser/NameSimplifier.cpp#L84

PR #84, opened 2021-10-25
pack all info required for warp operatorions into one output file

murcake:
Instead of writing various artifacts, we pack all necessary
    information into a json file.


BitBaseBit (approved):
LGTM

PR #92, opened 2021-10-28
Make ExecutionEnvironment a pointer

swapnilraj:


warp/yul/ToCairoVisitor.py
```diff
@@ -380,10 +382,12 @@ def _make_external_function(
             params += "calldata_size, calldata_len, calldata : felt*"
             init_exec_env = (
                 f"let (returndata_ptr : felt*) = alloc()\n"
-                f"local exec_env : ExecutionEnvironment ="
-                f"ExecutionEnvironment(calldata_size=calldata_size,"
+                f"let (local exec_env : ExecutionEnvironment*) = alloc()\n"
```
murcake, line 385: I think this line can be removed

murcake (approved):

warp/yul/ToCairoVisitor.py
JorikSchellekens: Removed
=

murcake: Probably will require regenerating files

swapnilraj:
I have updated the test expectations, the test should pass now! 🤞

murcake:
Let's squash the last two commits and merge

PR #99, opened 2021-11-01
Use a felt* for return data

JorikSchellekens:


murcake (approved):
Finally!

murcake:
Let's merge it, we need 0.5.1 to test deployment

PR #105, opened 2021-11-03
Murcake/remove external wrappers

murcake:


warp/yul/starknet_utils.py
```diff
@@ -0,0 +1,30 @@
...
+    starknet: StarknetState, program_info: dict, address: str, method: str, *args: list
+) -> TransactionExecutionInfo:
+    evm_calldata = get_evm_calldata(
+        program_info["sol_abi"],
+        program_info["sol_abi_original"],
+        program_info["sol_bytecode"],
+        method,
+        args,
+    )
+    print(evm_calldata)
```
JorikSchellekens, line 19: Can we get rid of these prints?

JorikSchellekens (approved):
Smaller code on both sides, win win

warp/yul/starknet_utils.py
murcake: Yes, definitely

PR #110, opened 2021-11-04
Add support for short strings (ascii strings of max length 31)

swapnilraj:
Co-authored by : Jorik Schellekens <joriksch@gmail.com>


JorikSchellekens (approved):
Looks good, maybe bring it up to the team that non ascii strings will be modified.

murcake (approved):
Don't we actually support strings of size 32? Each Uint256 contains two strings of 16-bytes max

warp/cairo-src/evm/memory.cairo
```diff
@@ -90,8 +90,9 @@ end
 
 func mstore8_{memory_dict : DictAccess*, range_check_ptr, msize}(offset, value : Uint256):
     # Extracts lower byte of 'value' and does 'mstore8' on it. Also updates 'msize'.
-    let (byte, _) = extract_lowest_byte(value)
-    let (msize) = update_msize(msize, offset, 1)
+    alloc_locals
+    let (local byte, _) = extract_lowest_byte(value)
+    let (local msize) = update_msize(msize, offset, 1)
```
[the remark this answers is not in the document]
swapnilraj: The reference for msize and byte were being revoked

swapnilraj:
> Don't we actually support strings of size 32? Each Uint256 contains two strings of 16-bytes max

I haven't phrased it properly but the max length 31 is a bit mis-leading. String literals that we generate will be of max 31 length because cairo only supports that, but the interesting point to note is that YUL splits strings into 32 byte chunks so if you have a string of size 64 bytes it will be split into two 32 bytes but since we generate string literals on the cairo side we would still drop once byte from both the splits.
If we encode the strings ourselves instead of relying on cairo to do it then we do not have the 31 characters limit, we do that in a later PR.

murcake:
> > Don't we actually support strings of size 32? Each Uint256 contains two strings of 16-bytes max
> 
> I haven't phrased it properly but the max length 31 is a bit mis-leading. String literals that we generate will be of max 31 length because cairo only supports that, but the interesting point to note is that YUL splits strings into 32 byte chunks so if you have a string of size 64 bytes it will be split into two 32 bytes but since we generate string literals on the cairo side we would still drop once byte from both the splits. If we encode the strings ourselves instead of relying on cairo to do it then we do not have the 31 characters limit, we do that in a later PR.

But we never generate string literals of length any other than 16, right? We split 32-byte string literals into two parts of 16 bytes. Thus, we support 32-bytes-sized string chunks of Yul and thus support solidity strings of any length. What am I missing?

swapnilraj:
> > > Don't we actually support strings of size 32? Each Uint256 contains two strings of 16-bytes max
> > 
> > 
> > I haven't phrased it properly but the max length 31 is a bit mis-leading. String literals that we generate will be of max 31 length because cairo only supports that, but the interesting point to note is that YUL splits strings into 32 byte chunks so if you have a string of size 64 bytes it will be split into two 32 bytes but since we generate string literals on the cairo side we would still drop once byte from both the splits. If we encode the strings ourselves instead of relying on cairo to do it then we do not have the 31 characters limit, we do that in a later PR.
> 
> But we never generate string literals of length any other than 16, right? We split 32-byte string literals into two parts of 16 bytes. Thus, we support 32-bytes-sized string chunks of Yul and thus support solidity strings of any length. What am I missing?

You're right actually, I'll fix the git message and CI then.

murcake (approved):

warp/yul/parse.py
```diff
@@ -75,7 +75,14 @@ def parse_literal(yul_ast) -> ast.Literal:
     elif kind == "bool":
         return ast.Literal(yul_ast["value"] == "true")
     elif kind == "string":
-        raise WarpException("string literals are not supported yet")
+        parsed_string = "".join([i if i.isascii() else "" for i in yul_ast["value"]])[
+            :32
```
JorikSchellekens, line 79: We don't need to do this anymore because YUL splits strings

JorikSchellekens (changes requested):

warp/yul/parse.py
swapnilraj: I'll keep the 32 just for sanity?

murcake: I think it very much interferes with sane formatting =)

swapnilraj: fair

murcake: Also, that `isascii` check isn't going to work at all AFAIU. Unicode symbols are not presented as-is but [as escapes](https://docs.soliditylang.org/en/latest/yul.html?highlight=optimization#literals)

murcake:
Can you please also update macos kudu versions?

murcake:
When I try to run `kudu` from this branch, it tells me that strings are not supported

murcake:
> > When I try to run `kudu` from this branch, it tells me that strings are not supported
> 
> Weird I got the linux binary from the CI on string_parsing branch, I'll try again

Sorry, I was a bit mistaken, it was the case when I rebased on main. If I don't rebase I just get InternalCompilerError

murcake:
Oh, I see, I added `    string public interesting        = "hoho";`. Storage variables aren't supported yet, I assume

JorikSchellekens (approved):

PR #119, opened 2021-11-12
Dead variable elimination

JorikSchellekens:
The expression splitter was mostly fine in its implementation (I tested on things like `g(g(g(g(g(g())))))` and `f(g(), g(f(), g()))`). You will notice however that there have been a few lines added where Uint256s have been hoisted up into a variable. This is because we can't destructure an unassigned argument in a function in cairo `f(Uint256(...).low)`. This isn't relevant now but comes into play when the expression splitter is disabled in kudu and variable inlining and constant folding is implemented.

Also:
Adds a fix to `extcodesize` and fix constant functions being passed the arguments of the original functioncall in the
builtinhandler


JorikSchellekens:
expecting this to fail because I'm waiting on the expressions splitter kudu build

murcake:
> This is because we can't destructure an unassigned argument in a function in cairo f(Uint256(...).low).
This has been fixed by #122, now all functions accept Uint256's. No need for hoisting up

murcake:
You should probably change the merge-base branch

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,171 @@
...
+    ) -> (LiveVars, ast.FunctionDefinition):
+        # We need to ignore the generated getters from kudu since their
+        # body is only instantiated in ToCairoVisitor
+        if node.name.startswith("getter"):
+            return (set(), node)
+        (live_vars, body) = self.visit(
+            node.body, return_vars=set([n.name for n in node.return_variables])
+        )
+        live_vars = live_vars - set([arg.name for arg in node.parameters])
+        # The warp ifs are the only place where the if branch has a return in each branch
```
murcake, line 78: Not really, we don't extract into `__warp_if` if-expressions where at least one branch results in return

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,171 @@
...
+                return_variables=node.return_variables,
+                body=body,
+            ),
+        )
+
+    def visit_if(
+        self, node: ast.If, live_vars: Set[str] = None, return_vars: Set[str] = None
+    ) -> Tuple[LiveVars, ast.If]:
+        if any(
+            [
```
murcake, line 95: No need for list literals

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,171 @@
...
+                isinstance(statement, ast.Leave)
+                for statement in node.body.statements
+                + (node.else_body.statements if node.else_body else ())
+            ]
+        ):
+            live_vars = return_vars | live_vars
```
murcake, line 101: Just carping: `live_vars |= return_vars`

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def visit_break(self, node: ast.Break) -> Tuple[LiveVars, ast.Break]:
+        # TODO: Implement these so we can run DeadCodeEliminator at any stage
+        assert False, "There should be no breaks, run ForLoopIliminator first"
+        return (set(), node)
+
+    def visit_continue(self, node: ast.Continue) -> Tuple[LiveVars, ast.Continue]:
+        # TODO: Implement these so we can run DeadCodeEliminator at any stage
+        assert False, "There should be no continues, run ForLoopIliminator first"
+        return (set(), node)
```
murcake, line 131: Its okay to assume that you have a simplified tree. It's a good idea to write in the docs which visitors should go before this one though

tests/yul/return-var-capturing.json
```diff
@@ -0,0 +1 @@
...ternalType": "uint256", "name": "", "type": "uint256"}], "stateMutability": "payable", "type": "function"}], "sol_bytecode": "608060405234801561001057600080fd5b506101e7806100206000396000f3fe60806040526004361061001e5760003560e01c806396b8d9a414610023575b600080fd5b61003d6004803603810190610038919061009c565b610053565b60405161004a91906100eb565b60405180910390f35b6000818311156100715760158361006a9190610106565b9050610081565b602a8261007e9190610106565b90505b92915050565b6000813590506100968161019a565b92915050565b600080604083850312156100b3576100b2610195565b5b60006100c185828601610087565b92505060206100d285828601610087565b9150509250929050565b6100e58161015c565b82525050565b600060208201905061010060008301846100dc565b92915050565b60006101118261015c565b915061011c8361015c565b9250827fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0382111561015157610150610166565b5b828201905092915050565b6000819050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fd5b600080fd5b6101a38161015c565b81146101ae57600080fd5b5056fea264697066735822122082d1d0258c8772d8814472b146cdb4e44bb8a64a8ce2f54637dbf6b7fa53328d64736f6c63430008060033", "dynamic_argument_functions": []}
```
murcake, line 1: This probably shouldn't be there

murcake (commented):
It seems that for most contracts that results in the size increase. Can we please split this commit into two separate ones: expression splitting and unused variables eliminator? It would be easier to study how contract sizes are affected

warp/yul/ExpressionSplitter.py
```diff
@@ -40,6 +40,19 @@ def visit_function_call(self, node: ast.FunctionCall):
...
+                False
+            ), "Literals shouldn't be at top level, these should be optimised away at this point"
+        parent = self.path[-2]
+        if isinstance(
+            parent, (ast.Assignment, ast.VariableDeclaration, ast.ExpressionStatement)
+        ):
+            return node
+        var_name = self.env.process_subexpr(node)
+        return ast.Identifier(var_name)
+
```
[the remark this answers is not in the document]
JorikSchellekens: Can I see this change?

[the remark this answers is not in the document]
murcake: https://github.com/NethermindEth/warp/pull/122/commits/8fff3df8b1e14631db5358559bc5309805051eaf

warp/yul/DeadcodeEliminator.py
JorikSchellekens: This is not a good idea here, it replaces live_vars in place and can easily cause unexpected sets to be passed into functions

warp/yul/DeadcodeEliminator.py
JorikSchellekens: Yes, but in this case we could usefully add the eliminator at any stage. I'll update the docs

warp/yul/ExpressionSplitter.py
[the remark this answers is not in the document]
murcake: Well, not quite. It can happen in `if-expressions` because we do `if x.low + x.high != 0`, I guess. If this doesn't come up with this PR though, let's not worry about it. It's better fixed in ToCairoVisitor, by simply checking if the condition is a literal

warp/yul/DeadcodeEliminator.py
murcake: Ah right, it's passed from somewhere else. 
In general, it seems that you never modify sets. Why not enforce that using `frozenset`? In that case `|=` is completely safe also.

`frozenset` might be bothersome to type and read, so we can shorten it by `fset = frozenset` somewhere above

warp/yul/ExpressionSplitter.py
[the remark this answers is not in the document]
JorikSchellekens: Fine, I'll remove it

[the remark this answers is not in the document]
murcake: It didn't work for me to just remove it, there was some error about top level uninstantiated variables in `return-var-capturing` and only there. I didn't see any variables that fit the descriptiong in that contract though

warp/yul/DeadcodeEliminator.py
JorikSchellekens: Thanks

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,172 @@
...
+        assert (
+            vs == set()
+        ), "There should be no uninstantiated variables in the top scope."
+        return node
+
+    def visit_list(
+        self, nodes: Sequence[ast.Node], *args, **kwargs
+    ) -> (LiveVars, List[ast.Node]):
+        if len(nodes) == 0:
+            return (set(), list())
```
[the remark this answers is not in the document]
JorikSchellekens: Turns out that we need it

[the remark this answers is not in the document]
murcake: Alright)

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,172 @@
...
+                return_variables=node.return_variables,
+                body=body,
+            ),
+        )
+
+    def visit_if(
+        self, node: ast.If, live_vars: Set[str] = None, return_vars: Set[str] = None
+    ) -> Tuple[LiveVars, ast.If]:
+        if any(
+            isinstance(statement, ast.Leave)
```
murcake, line 98: There can be nested blocks and nested ifs, so it's probably not enough of a check. I suggest storing `return_vars` as a class field in a way similar to the `ScopeFlattener._new_function` context manager. Would be nice to wrap it in a ReaderT, but alas we're in the python land.

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,172 @@
...
+        if live_vars is None:
+            live_vars = fset()
+        if return_vars is not None:
+            live_vars |= return_vars
+        statements = []
+        for n in reversed(node.statements):
+            keep = True
+            if isinstance(n, ast.If):
+                (current_live_vars, statement) = self.visit(n, live_vars, return_vars)
+                live_vars = current_live_vars
```
murcake, line 151: At first I though that the set of live vars shrinks here, but apparently it's extended because we pass pre-existing `live_vars` to `self.visit` right? Is there any reason that we should pass it as an argument? 
I in general don't understand why if's treatment differs from other nodes, could you please explain?
Ah, I see that is because of passing `return_vars`. If we use a contextmanager as I suggested in the other commen, maybe we can get rid of special treatment for if?

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,172 @@
...
+                statements.append(statement)
+                continue
+
+            (current_live_vars, statement) = self.visit(n)
+            if isinstance(statement, (ast.Assignment, ast.VariableDeclaration)):
+                var_list = _get_variables(statement)
+                if not isinstance(statement.value, ast.FunctionCall):
+                    # We know that there is only one var
```
murcake, line 159: let's assert JIC

warp/yul/DeadcodeEliminator.py
```diff
@@ -0,0 +1,172 @@
...
+        (live_vars, _) = self.visit(node.value)
+        return (live_vars, node)
+
+    def visit_function_definition(
+        self, node: ast.FunctionDefinition
+    ) -> (LiveVars, ast.FunctionDefinition):
+        # We need to ignore the generated getters from kudu since their
+        # body is only instantiated in ToCairoVisitor
+        if node.name.startswith("getter"):
+            return (fset(), node)
```
murcake, line 79: Good news! We are getting rid of getters! No need for this branch now, yay

warp/yul/DeadcodeEliminator.py
JorikSchellekens: Actually there is a subtle bug in how I've done this. I wanted to kill any variable (declared before the scope of an if) that was assigned to in both branches of the if statement. However I didn't differentiate between variable declaration and assignment so any variables which are declared in both branches will kill a variable declaration of the same name before. I'm going to modify this to take away the `special treatment` which will underaproximate the analysis. I don't think we hit that case often though with the current ssa.

warp/yul/DeadcodeEliminator.py
JorikSchellekens: Great idea thank you

swapnilraj (approved):

PR #120, opened 2021-11-12
remove py-solc-x

BitBaseBit:


warp/cairo-src/evm/calls.cairo
```diff
@@ -56,10 +56,10 @@ func calldata_load{range_check_ptr, exec_env : ExecutionEnvironment*}(offset) ->
     return (value=value)
 end
 
-func returndata_copy{range_check_ptr, exec_env : ExecutionEnvironment, memory_dict : DictAccess*}(
+func returndata_copy{range_check_ptr, exec_env : ExecutionEnvironment*, memory_dict : DictAccess*}(
         memory_pos : Uint256, returndata_pos : Uint256, length : Uint256):
     array_copy_to_memory(
-        exec_env.returndata_len, exec_env.returndata, returndata_pos, memory_pos, length)
+        exec_env.returndata_len, exec_env.returndata, returndata_pos.low, memory_pos.low, length.low)
```
murcake, line 62: Let's do it in a separate commit and add a test to ensure that desync doesn't happen again

warp/yul/main.py
```diff
@@ -86,7 +86,7 @@ def transpile_from_yul(
     yul_ast = SwitchToIfVisitor().map(yul_ast)
     yul_ast = ExpressionSplitter(name_gen).map(yul_ast)
     yul_ast = RevertNormalizer().map(yul_ast)
-    yul_ast = ScopeFlattener(name_gen).map(yul_ast)
+    # yul_ast = ScopeFlattener(name_gen).map(yul_ast)
```
murcake, line 89: Let's also remove ScopeFlattener in another commit

requirements.txt
```diff
@@ -6,7 +6,6 @@ sympy
 pytest==6.2.4
 pytest-parallel
```
murcake, line 7: Can't comment on binary files, so commenting here =(. Why is the linux/kudu binary moved from `./warp/bin` to `./bin`? I think setup.py expects it to be in the former directory?

murcake (approved):
It's nice to get rid of another dependency

murcake:
I see that some tests fail, because we attempt to take the low part of a Uint256 constructor. #122 should help with that

BitBaseBit:
> I see that some tests fail, because we attempt to take the low part of a Uint256 constructor. #122 should help with that

This is odd because this test passes locally.

murcake:
> > I see that some tests fail, because we attempt to take the low part of a Uint256 constructor. #122 should help with that
> 
> This is odd because this test passes locally.

Do you run make test or make test_yul? The latter doesn't run every test

BitBaseBit:
> > > I see that some tests fail, because we attempt to take the low part of a Uint256 constructor. #122 should help with that
> > 
> > 
> > This is odd because this test passes locally.
> 
> Do you run make test or make test_yul? The latter doesn't run every test

Ah okay , yeah the issue is there when I run test_yul

BitBaseBit:
> > > I see that some tests fail, because we attempt to take the low part of a Uint256 constructor. #122 should help with that
> > 
> > 
> > This is odd because this test passes locally.
> 
> Do you run make test or make test_yul? The latter doesn't run every test

Go ahead & merge #122

murcake:
(also, it seems that this version of kudu is expression-splitting and the pr for it is #119 AFAIU)

PR #123, opened 2021-11-13
Compile script and update .gitignore

JorikSchellekens:
Add script to compile tests/yul/*.sol
Ignore .warp-deps

### DIFF
--- a/scripts/yul/compile_sol_files_in_tests_yul.sh
+++ b/scripts/yul/compile_sol_files_in_tests_yul.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+for f in tests/yul/*.sol do
+  warp transpile $f WARP --cairo-output &
+done
+wait


scripts/yul/compile_sol_files_in_tests_yul.sh
```diff
@@ -0,0 +1,4 @@
+for f in $(ls tests/yul/*.sol); do
```
murcake, line 1: Just carping: why not `for f in tests/yul/*.sol;`?

.gitignore
```diff
@@ -5,6 +5,7 @@ vendor/cairo-lang/build
 
 # Secrets
 /.warp-activation-token
+/.warp-deps
```
murcake, line 8: What command generates that?

murcake (approved):

.gitignore
JorikSchellekens: Ah, something I'd added to the makefile a long time ago but seems like someone changed the makefile

murcake (approved):
I'd also add a shebang to the script

murcake:
We should probably remove the ", ignore .warp-deps" from the commit description.

Sorry, for telling this in bits, I didn't notice the last time.

PR #124, opened 2021-11-14
fix json literals in abi

BitBaseBit:


murcake:
Does that need to go to `main`?

swapnilraj (approved):
Looks good, can you just a small note about this change in the docstring

warp/yul/utils.py
```diff
@@ -197,4 +197,6 @@ def make_abi_StarkNet_encodable(abi):
     """
     abiStr = f"{abi}".replace("'address'", "'uint256'")
     abiStr = abiStr.replace("'", '"')
+    abiStr = abiStr.replace("True", "true")
+    abiStr = abiStr.replace("False", "false")
```
murcake, line 201: Why do we do that? Also I'm worried about it changing function names like `isTrue` to `istrue`, which would change the selector again

BitBaseBit: oh yeah, I'll fix it now

PR #129, opened 2021-11-16
add remaining builtins

BitBaseBit:
Adds all the remaining Yul builtiins. Throws an error if certain builtins are present in the generated Yul.

### DIFF
--- a/warp/yul/BuiltinHandler.py
+++ b/warp/yul/BuiltinHandler.py
@@ -11,6 +11,11 @@ from yul.Imports import Imports
 UINT256_MODULE = "starkware.cairo.common.uint256"
 
 
+class PrettyType(type):
+    def __repr__(self):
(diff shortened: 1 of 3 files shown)


warp/yul/BuiltinHandler.py
```diff
@@ -453,6 +467,22 @@ def __init__(self):
         )
 
 
+class SelfBalance(StaticHandler):
+    def __init__(self):
+        super().__init__(
+            function_name="address",
+            module="",
```
murcake, line 474: `module=None` and you won't need to override `required_imports`. Also, is there an address function in cairo? Is it related to balance somehow?

murcake (approved):

swapnilraj:
can you add the built-in pop, I encountered it when compiling UniSwapV3Router

PR #136, opened 2021-11-22
support for delegatecall

murcake:
The implementation abstracts from the concrete (usual or delegate)
     call function. It takes the call function as a parameter. This is
     unnecessary and will have to be reverted if/when functions for
     delegate and usual calls will differ in signature. The
     generalization was done as a fun exercise in low-level Cairo.

Adding a more complex test is complicated by some bugs in the way
     constructors are generated.

### DIFF
--- a/tests/behaviour/delegatecall-attack.sol
+++ b/tests/behaviour/delegatecall-attack.sol
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.6;
+
+/*
+MIT License
+
+Copyright (c) 2018 Tasuku Nakamura
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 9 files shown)


swapnilraj:
From slack:
S: Hey artem I was just going through your delegatecall PR and you mention in the comments that low-level cairo was a fun 
    exercise, do you plan to leave it like that? It took me a second to understand what was happening (I didn’t read the  PR 
    description just jumped to the code)
A: It would still be tough to read, because there is no high-level interface for delegate_call. However right now we manually 
   construct both calldata and call arguments. If we separate implementations, we wouldn't have to manually construct call 
   arguments
   I wrote it with the intention that if reviewers prefer to remove that part with callback functions, I'll do it
   If you wish to, could you please express so in the PR comment, just for history?
S: No I am fine with the callback.
    Sorry what do you mean by calldata and call arguments?
    I’ll copy my comments from here to the PR

murcake:
>  Sorry what do you mean by calldata and call arguments?

calldata is data that is passed through StarkNet to another contract when we use call_contract or delegate_call.
call arguments are arguments of the delegate_call function.

Manually constructing calldata means that we have to copy all arguments into an array. Manually constructing call argument means that we put them into  `[ap]` one by one.

So, suppose we have a function `foo(a, b) -> (c, d) `.That's how we usually call it: `let (c, d) = foo(1, 2)`

That's how we call it "low-level style" (when the function is dynamic, e.g.)

```
let (foo_label) = get_function_label(foo)
[ap] = 1; ap++
[ap] = 2; ap++
call abs foo_label
let c = [ap - 2]
let d = [ap - 1]
```

swapnilraj:
The code looks good to me, we should hold off merging this PR until the constructor bug is fixed and then we can add an e2e test for this as well.

swapnilraj (approved):

swapnilraj:
> Looks good to me

Oops pressed the wrong button there

PR #138, opened 2021-11-22
Implement a printer for the ast that returns a yul-like code

cicr99:
- The new printer for ast generates yul code, except for the `if` statements. Yul language does not support `else` branches, however our visitors generate them, therefore the printer outputs them. Also, type Uint256 is omitted from type annotations, since it is the default one.
- The `AstPrinter` was removed and test cases results were fixed according to the new format

### DIFF
--- a/tests/ast/cairo-keywords-for-names.ast.result
+++ b/tests/ast/cairo-keywords-for-names.ast.result
@@ -1,56 +1,10 @@
-Block:
-	FunctionDefinition:
-		Name: func__warp_mangled
-		Parameters:
-			TypedName:
-				felt__warp_mangled:Uint256
-			TypedName:
-				end__warp_mangled:Uint256
-		Return Variables:
-			TypedName:
-				jmp__warp_mangled:Uint256
-		Body:
-			Block:
-				VariableDeclaration:
-					Variables:
-						TypedName:
-							_1_7:Uint256
-					Value:
-						FunctionCall:
-							Identifier:iszero
-							Identifier:felt__warp_mangled
-				VariableDeclaration:
-					Variables:
-						TypedName:
-							_2_8:Uint256
-					Value:
-						FunctionCall:
-							Identifier:iszero
-							Identifier:_1_7
-				ExpressionStatement:
-					FunctionCall:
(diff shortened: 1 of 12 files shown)


tests/ast/cairo-keywords-for-names.ast.result
```diff
@@ -1,56 +1,10 @@
...
-					Variables:
-						Identifier:jmp__warp_mangled
-					Value:
-						FunctionCall:
-							Identifier:gt
-							Identifier:ret__warp_mangled
-							Identifier:_3_9
\ No newline at end of file
+{
+	function func__warp_mangled(felt__warp_mangled : Uint256, end__warp_mangled : Uint256) jmp__warp_mangled : Uint256 {
```
murcake, line 2: Isn't there supposed to be a `->` before the return variable?

warp/yul/AstTools.py
```diff
@@ -413,3 +295,116 @@ def get_word(self, start: int) -> str:
...
         name = "_".join(re.findall("[A-Z][^A-Z]*", name))
         return name.lower()
+
+
+class YulPrinter(AstVisitor):
+    def format(self, node: ast.Node, tabs: int = 0) -> str:
+        return self.visit(node, tabs)
+
+    def visit_typed_name(self, node: ast.TypedName, tabs: int = 0) -> str:
+        return f"{node.name} : {node.type}"
```
murcake, line 305: Just a suggestion, since Uint256 is the default type in the EVM dialect of Yul that we use, we can simply omit it. It leads to a bit less visual noise.

murcake (approved):
It's so much easier to decipher now!

tests/ast/cairo-keywords-for-names.ast.result
cicr99: Oh yes, that's true! I'll add it.

warp/yul/AstTools.py
cicr99: Also true, I'll change it as well

murcake (approved):

murcake:
Let's merge it

PR #141, opened 2021-11-29
fix returns not ending the execution

murcake:


JorikSchellekens (approved):
Do all the termination_token ifs cause a noticeable overhead?

murcake:
> Do all the termination_token ifs cause a noticeable overhead?

I'm waiting for #146 to get finished, so I haven't benchmarked yet. However, this PR only adds one `termination_token` check to all of our test contracts combined. It's in the contract that uses `return` in the inline assembly. So, the only burden is the burden of passing another `felt` implicit, which I assume is not a lot.

JorikSchellekens:
Yeah of course, that would be fine

PR #142, opened 2021-11-29
Murcake/simplify ctors handling

murcake:
Most of the size increases go to constructors. Constructors are called only once, so this increase is not as significant.


BitBaseBit (approved):
LGTM

murcake:
As you can see comparing to the [previous benchmark stats](https://github.com/NethermindEth/warp-benchmark/blob/master/5d492d7b208af4f3541ab2420e307240924adc25.md), stats [for this PR](https://github.com/NethermindEth/warp-benchmark/blob/master/d6e9516ee791827cece2fccd0ec3681df9296607.md) are much worse. However, this change is needed to enable inlining optimizations, which result in [even better performance, then before](https://github.com/NethermindEth/warp-benchmark/blob/master/209e44583c41159c35589885f4767db1e13eadcc.md)

PR #145, opened 2021-12-01
fix missing returns in the entry point function

murcake:
Problem: we replace returns with empty strings in the main
     function. The reason is that we change the signature of the main
     function, add return variables, so the returns are not
     valid. However, it might change semantics and it leads to
     non-compiling code for the new contract 'ctor-check.sol'.

     Another issue is that sometimes we don't finalize certain
    implicits on early returns

Solution: transpile the main and constructor function as is, without
     any changes. Add wrappers that initialize and finalize implicits
     and perform the outermost returning. The semantics are unchanged
     and finalization is guaranteed.

Also now we add only necessary implicits to the constructor and main
     functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,10 +25,10 @@ test_bats: warp $(BATS_FILES) tests/cli/*.bats
 test_yul: warp
 	mkdir -p benchmark/stats
 	mkdir -p benchmark/tmp
+	python -m pytest tests/ast/ -v --tb=short --workers=auto $(ARGS)
 	python -m pytest scripts/yul/transpile_test.py -v --tb=short --workers=auto $(ARGS)
 	python -m pytest scripts/yul/compilation_test.py -v --tb=short --workers=auto $(ARGS)
 	python -m pytest tests/yul/ -v --tb=short --workers=auto $(ARGS)
-	python -m pytest tests/ast/ -v --tb=short --workers=auto $(ARGS)
 .PHONY: test_yul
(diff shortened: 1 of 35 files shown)


BitBaseBit (approved):
LGTM

warp/yul/main.py
```diff
@@ -89,6 +91,49 @@ def transpile_from_yul(yul_ast: ast.Node) -> str:
...
+    )
+
+    deployment_names = {
+        x.name for x in deployment_functions if isinstance(x, ast.FunctionDefinition)
+    }
+    runtime_names = {
+        x.name for x in runtime_functions if isinstance(x, ast.FunctionDefinition)
+    }
+    duplicated = deployment_names & runtime_names
+    renamer = Renamer(lambda x: x + "_deployment" if x in duplicated else x)
```
JorikSchellekens, line 119: Could you write a test for this case?

murcake: Hmm, I believe it is tested by some of the current tests, because before I wrote that "rewrite" piece they didn't pass because of function duplication. I can write a specific test-case (but it would require a little changing)

JorikSchellekens: If it fixes the other I think it aught to be fine

murcake: This change is not actually related to this branch, it's just that I haven't rebased it yet. I've written a test in #142

PR #151, opened 2021-12-10
Test Scaffolding

BitBaseBit:


.gitignore
```diff
@@ -43,3 +44,132 @@ venv.bak/
...
+# Stores VSCode versions used for testing VSCode extensions
+.vscode-test
+.vscode/
+
+# yarn v2
+.yarn/cache
+.yarn/unplugged
+.yarn/build-state.yml
+.yarn/install-state.gz
+.pnp.*
```
murcake, line 175: Wow, javascript sure is noisy

tests/hardhat/test/Token.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// `describe` is a Mocha function that allows you to organize your tests. It's
+// not actually needed, but having your tests organized makes debugging them
+// easier. All Mocha functions are available in the global scope.
+
+// `describe` receives the name of a section of your test suite, and a callback.
+// The callback must define the tests of that section. This callback can't be
+// an async function.
+describe('Token contract', function () {
+  this.timeout(600_000); // 10 min
```
murcake, line 17: Is that normal?

tests/hardhat/test/Token.ts
```diff
@@ -0,0 +1,84 @@
...
+  let Token: ContractFactory;
+  let hardhatToken: Contract;
+  let owner: SignerWithAddress;
+  // eslint-disable-next-line no-unused-vars
+  let addrs: SignerWithAddress[];
+
+  // Prepend all Starknet/Cairo variables with c_
+  let c_TokenAddress: string;
+
+  beforeEach(function () {});
```
murcake, line 36: Is that necessary?

tests/hardhat/test/Token.ts
```diff
@@ -0,0 +1,84 @@
...
+      // ========================================================
+      [hardhatToken, c_TokenAddress] = await Promise.all([
+        pHardhatToken,
+        c_pTokenAddress,
+      ]);
+    });
+
+    // Get the ContractFactory and Signers here.
+    // `it` is another Mocha function. This is the one you use to define your
+    // tests. It receives the test name, and a callback function.
```
murcake, line 65: The description of `it` is a bit too late. Perhaps, we should move it before the first `it` function or remove altogether and just give a link to the docs

tests/hardhat/test/Token.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      // This test expects the owner variable stored in the contract to be equal
+      // to our Signer's owner.
+      const res: string = await warp_call(
+        'contracts/warp_artifacts/Token.json',
+        'owner',
+        '[]',
+        c_TokenAddress,
+      );
+      expect(res.trim()).equal('32 2 0 0');
```
murcake, line 80: So, the owner is no one — because it was called from the ~CLI. Is that expected?

warp/cli/commands.py
```diff
@@ -74,26 +146,45 @@ def starknet_compile(cairo_path, contract_base):
...
-    print(
-        os.popen(
-            f"starknet deploy "
-            f"--contract {contract_base}_compiled.json "
-            f"--inputs {inputs} "
-            f"--network {network} "
-        ).read()
-    )
+    if gateway_url:
+        click.echo(
```
murcake, line 171: Any reason to use `click.echo` here as opposed to `print` in all other places?

warp/cli/commands.py
```diff
@@ -27,29 +29,99 @@ async def send_req(method, url, tx: Optional[Union[str, Dict[str, Any]]] = None)
...
+        print(
+            os.popen(
+                f"starknet invoke "
+                f"--address {address} "
+                f"--abi {abi} "
+                f"--function __main "
+                f"--inputs {inputs} "
+                f"--network {network} "
+            ).read()
+        )
```
murcake, line 124: Minor, but maybe we can just extract `network_param = "--network {network}" \/ "--feeder={gateway}"` instead of duplicating the whole call?

Also, when `os.popen` fails, it doesn't report errors, so it would be nice to use something like [subprocess.check_output](https://docs.python.org/3/library/subprocess.html?highlight=subprocess%20run#subprocess.check_output)

warp/cli/warp_cli.py
```diff
@@ -77,7 +80,7 @@ def transpile(file_path, contract_name, cairo_output):
     ),
 )
 @click.pass_obj
-def invoke(obj, program, address, function, inputs):
+def call(obj, program, address, function, inputs, gateway_url):
```
murcake, line 83: Isn't `gateway_url` already passed from `obj`? This will be a duplicated parameter

warp/cli/commands.py
```diff
@@ -1,10 +1,12 @@
 import json
 import os
 import subprocess
-from typing import Any, Dict, Optional, Sequence, Union
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 import aiohttp
+import click
 from cli.encoding import get_cairo_calldata, get_ctor_evm_calldata, get_evm_calldata
+from starkware.starknet.services.api.contract_definition import ContractDefinition
```
murcake, line 9: Not used anywhere, AFAICS

murcake (commented):
Can you please add hardhat tests to CI? Preferably as a separate job.

Also, it would be nice to add a CLI test for `warp call` to `tests/cli/tests.bats`

tests/hardhat/test/Token.ts
BitBaseBit: Yep

swapnilraj:
I can't comment on the binary, but what has changed in kudu?

tests/starknet-testnet/adapt.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Returns the `calldata` converted to the type specified by `input_type` (tuple if struct or tuple, number). Also returns the incremented `calldata_i`.
+    """
+
+    if input_type == "felt":
+        return calldata[calldata_i], calldata_i + 1
+
+    arr = []
+    if input_type[0] == "(" and input_type[-1] == ")":
+        members = input_type[1:-1].split(", ")
```
murcake, line 118: Can there not be nested tuples?

tests/hardhat/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "ethereum-waffle": "^3.4.0",
+    "ethers": "^5.5.2",
+    "hardhat": "^2.7.1",
+    "lodash": "^4.17.21",
+    "mocha": "^9.1.3",
+    "prettier": "^2.5.1",
+    "typescript": "^4.5.3"
+  },
+  "dependencies": {
+    "@openzeppelin/contracts": "^4.4.0",
```
swapnilraj, line 45: Most of the dependencies should be dev dependencies

tests/hardhat/util.ts
```diff
@@ -0,0 +1,37 @@
+import { AxiosResponse } from 'axios';
```
swapnilraj, line 1: Merge the two imports

tests/hardhat/README.md
```diff
@@ -0,0 +1,9 @@
+# Run Tests
+
+Each contract/directory in the `contracts` directory, has a corresponding test script in the `test` directory.
+First start the StarkNet testnet server: `python starknet-testnet/server.py`.
+To run the tests, simply use:
+```shell
+yarn test
```
murcake, line 7: Can we add hardhat tests to CI?

tests/hardhat/util.ts
BitBaseBit: Can't

tests/hardhat/README.md
BitBaseBit: I'll do it in another PR

BitBaseBit:
> I can't comment on the binary, so commenting here. What has changed in kudu?

I changed it to a build that doesn't require the newest Glibc.

tests/hardhat/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "warp",
+  "version": "1.0.0",
+  "main": "index.js",
+  "repository": "git@github.com:NethermindEth/warp.git",
+  "author": "BitBaseBit <vardygreg23@gmail.com>",
+  "license": "Apache 2.0",
+  "scripts": {
+    "test": "npx mocha --extension ts --require hardhat/register --recursive --parallel --exit"
```
swapnilraj, line 9: Is there a specific reason for using npx? Can we use `yarn run mocha`

swapnilraj (approved):

PR #152, opened 2021-12-13
Implement pipelining order for optimizing visitors

Glitch18:
Implements a second parameter in `transpile_from_yul`. This is a string which allows one to change the order of the Optimizing Visitors according to the following map:

```
function_map = {
        "v": VariableInliner,
        "c": ConstantFolder,
        "o": FoldIf,
        "l": LeaveNormalizer,
        "r": RevertNormalizer,
        "f": FunctionPruner,
        "d": DeadcodeEliminator,
}
```

I tested these changes against the tests we have defined in our make file. All of these tests pass apart from a few unit tests in 
`python -m pytest scripts/yul/compilation_test.py -v --tb=short --workers=auto`

I'm working on fixing these failues. After that, I'll implement a command line parameter which will take a string of letters which will act as keys according to the above dictionary. Marking this PR as a draft until then.

### DIFF
--- a/warp/yul/main.py
+++ b/warp/yul/main.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -31,7 +32,9 @@ from yul.WarpException import warp_assert
 AST_GENERATOR = "kudu"
 
 
-def transpile_from_solidity(sol_src_path, main_contract) -> dict:
+def transpile_from_solidity(
+    sol_src_path, main_contract, optimizers_order="VFoLRFD"
+) -> dict:
     sol_src_path_modified = str(sol_src_path)[:-4] + "_marked.sol"
     if not shutil.which(AST_GENERATOR):
         sys.exit(f"Please install {AST_GENERATOR} first")
@@ -52,7 +55,7 @@ def transpile_from_solidity(sol_src_path, main_contract) -> dict:
 
     codes = json.loads(result.stdout)
     yul_ast = parse_to_normalized_ast(codes)
-    cairo_code = transpile_from_yul(yul_ast)
+    cairo_code = transpile_from_yul(yul_ast, optimizers_order)
     try:
         os.remove(sol_src_path_modified)
     except FileNotFoundError:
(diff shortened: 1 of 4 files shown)


warp/yul/main.py
```diff
@@ -57,24 +57,35 @@ def transpile_from_solidity(sol_src_path, main_contract) -> dict:
...
+
+    function_map = {
+        "v": VariableInliner,
+        "c": ConstantFolder,
+        "o": FoldIf,
+        "l": LeaveNormalizer,
+        "r": RevertNormalizer,
+        "f": FunctionPruner,
+        "d": DeadcodeEliminator,
+    }
```
murcake, line 81: You can make this map return lambdas to avoid additional special handling like below. For example, `"r": lambda: RevertNormalizer(builtins)`

warp/yul/main.py
```diff
@@ -57,24 +57,35 @@ def transpile_from_solidity(sol_src_path, main_contract) -> dict:
...
     yul_ast = ScopeFlattener(name_gen).map(yul_ast)
-    yul_ast = LeaveNormalizer().map(yul_ast)
-    yul_ast = RevertNormalizer(builtins).map(yul_ast)
-    yul_ast = FunctionPruner().map(yul_ast)
-    yul_ast = DeadcodeEliminator().map(yul_ast)
+
+    function_map = {
+        "v": VariableInliner,
+        "c": ConstantFolder,
+        "o": FoldIf,
```
swapnilraj, line 76: If we use just one letter for each pass it would limit us to just 26 passes, maybe you can look at the other warp repo to see how Dom implemented multi-character abbreviations.
And it would also help with choosing better abbreviations in the case of `FoldIf`

warp/yul/main.py
```diff
@@ -57,24 +57,35 @@ def transpile_from_solidity(sol_src_path, main_contract) -> dict:
...
     builtins = get_default_builtins(cairo_functions)
+
     yul_ast = ForLoopSimplifier().map(yul_ast)
     yul_ast = ForLoopEliminator(name_gen).map(yul_ast)
     yul_ast = MangleNamesVisitor().map(yul_ast)
     yul_ast = SwitchToIfVisitor().map(yul_ast)
-    yul_ast = VariableInliner().map(yul_ast)
-    yul_ast = ConstantFolder().map(yul_ast)
-    yul_ast = FoldIf().map(yul_ast)
     yul_ast = ExpressionSplitter(name_gen).map(yul_ast)
```
swapnilraj, line 69: We want to run variable inliner atleast once before expresssion splitter, I think that will also fix your tests

warp/yul/main.py
```diff
@@ -52,32 +55,41 @@ def transpile_from_solidity(sol_src_path, main_contract) -> dict:
...
-    yul_ast = FoldIf().map(yul_ast)
-    yul_ast = ExpressionSplitter(name_gen).map(yul_ast)
-    yul_ast = RevertNormalizer(builtins).map(yul_ast)
-    yul_ast = ScopeFlattener(name_gen).map(yul_ast)
-    yul_ast = LeaveNormalizer().map(yul_ast)
-    yul_ast = RevertNormalizer(builtins).map(yul_ast)
-    yul_ast = FunctionPruner().map(yul_ast)
-    yul_ast = DeadcodeEliminator().map(yul_ast)
+
+    pass_order = "FlsFleMSVERSf"
```
swapnilraj, line 71: Why not just do it one line?

warp/yul/main.py
```diff
@@ -31,7 +32,9 @@
 AST_GENERATOR = "kudu"
 
 
-def transpile_from_solidity(sol_src_path, main_contract) -> dict:
+def transpile_from_solidity(
+    sol_src_path, main_contract, optimizers_order="VFoLRFD"
```
swapnilraj, line 36: Do we need to have the default value for the `optimizers_order` in two places.
If it was in one place it would be easier to keep them in sync

swapnilraj (approved):

PR #157, opened 2021-12-15
optimize extract_unaligned_uint128

murcake:
Profiling `BaseJumpRateModelV2.utilizationRate` has shown that ~37% of
     steps are spent inside of `extract_unaligned_uint128`, in
     particular uint256 multiplication. We remove uint256 operations
     and instead use bitwise operations and masks to perform the
     extraction. It results in almost universal ~25-30% reduction in
     the number of steps.

### DIFF
--- a/warp/cairo-src/evm/bit_packing.cairo
+++ b/warp/cairo-src/evm/bit_packing.cairo
@@ -1,13 +1,19 @@
+from starkware.cairo.common.bitwise import bitwise_and, bitwise_not
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
 from starkware.cairo.common.math import assert_lt, unsigned_div_rem
 from starkware.cairo.common.pow import pow
-from starkware.cairo.common.uint256 import Uint256, uint256_shl
 
-func extract_unaligned_uint128{range_check_ptr}(shift, low, high) -> (value):
+const UINT128_BOUND = 2 ** 128
+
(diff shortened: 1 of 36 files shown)


warp/cairo-src/evm/bit_packing.cairo
```diff
@@ -1,13 +1,19 @@
...
+
+func extract_unaligned_uint128{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(
+        shift, low, high) -> (value):
     # Given two aligned uint128's, extract an unaligned uint128,
     # shifted by 'shift' bytes from the 'high' high end.
-    let x = Uint256(low=low, high=high)
-    let (y) = uint256_shl(x, Uint256(low=8 * shift, high=0))
-    return (y.high)
+    let (p) = pow(256, 16 - shift)
+    let (low_mask) = bitwise_not(p - 1)
```
swapnilraj, line 13: will extracting `p - 1` here going to have any performance implication, I don't imagine that the cairo-compiler can do that kind of optimisation.

murcake: Not much effect, probably, just one more memory cell, one less operation. Even one less felt division in this function gave me about just ~15 steps less.

swapnilraj: Okay

swapnilraj (approved):
This is really good!

PR #158, opened 2021-12-16
optimize uint256 shr

murcake:
The standard library version of this function it uses
     `uint256_unsigned_div_rem`, which is quite slow. The new version
     primarily uses bitwise operations and a much faster felt
     `unsigned_div_rem`.

That shaves ~300 steps from every call.


BitBaseBit (approved):
LGTM

BitBaseBit:
Merge when you're ready to :slightly_smiling_face:

PR #162, opened 2021-12-20
init

BitBaseBit:


warp/yul/BuiltinHandler.py
```diff
@@ -566,15 +579,19 @@ def get_function_call(self, function_args: Sequence[str]) -> str:
         return super().get_function_call([])
 
 
-class BaseFee(NotImplementedStarkNet):
-    pass
+class ChainID(DynamicHandler):
+    def __init__(self, cairo_functions: CairoFunctions):
+        super().__init__(lambda: cairo_functions.constant_function(1))
```
axic, line 584: Unsolicited comment: probably 1 is not a good idea, as that is Ethereum Mainnet (see [EIP-155](https://eips.ethereum.org/EIPS/eip-155)).

BitBaseBit: Thanks, I know, it's just a placeholder at the moment.

PR #167, opened 2021-12-27
restructure packages

murcake:
The first change is that we nest all warp-related packages into
     warp. It allows to have less namespace collisions, which we've
     had (logging, ast), and also a common practice.

The second change is that we put the new `warp` package to a src
     directory. Thus, it's impossible to use `warp` code without
     installing it first. That allows us to avoid situations when we
     check the code before uploading and it works, but the installed,
     about-to-upload, package is outdated and buggy.

Another change is using `pkg_resources`` and `importlib.resources`
     for managing resources. It is a recommended practice, because
     sometimes python modules are distributed in zip-files and
     `__file__` manipulations don't work.


murcake:
This is needed for the type fixing PR, because pytype stubmles upon our unusual code layout

BitBaseBit (approved):

PR #168, opened 2021-12-29
Run constant folder only on integers

swapnilraj:


BitBaseBit (approved):

warp/yul/ConstantFolder.py
```diff
@@ -46,7 +46,10 @@ def visit_function_call(
     ) -> Union[ast.Literal, ast.FunctionCall]:
         reduced_args = self.visit_list(node.arguments)
         reducer = REDUCERS.get(node.function_name.name)
-        if reducer and all(isinstance(arg, ast.Literal) for arg in reduced_args):
+        if reducer and all(
+            isinstance(arg, ast.Literal) and isinstance(arg.value, int)
```
murcake, line 50: While this does fix errors, we miss an opportunity for optimization. All values in Yul can be treated like ints. Bools are just 0/1. "Strings" are of limited length and just another presentation of 256-bit integers

PR #170, opened 2021-12-30
implement uint256_mod, uint256_mulmod, uint256_addmod, uint256_exp

BitBaseBit:


swapnilraj:
can you run cairo-format and add a behaviour test

warp/cairo-src/evm/uint256.cairo
```diff
@@ -108,13 +108,27 @@ func sgt{range_check_ptr}(op1 : Uint256, op2 : Uint256) -> (result : Uint256):
...
+        return (res=rem)
+    end
 end
 
 func uint256_addmod{range_check_ptr}(a : Uint256, b : Uint256, m : Uint256) -> (res : Uint256):
-    assert 0 = 1  # not implemented yet =(
-    jmp rel 0
+    let (pre_mod : Uint256) = uint256_add(a, b)
+    let (res : Uint256) = uint256_mod(pre_mod, m)
+    return (res=res)
```
murcake, line 125: The implementation of `addmod` and `mulmod` is not correct. Intermediate calculations, per yellow paper, should not be subject to modulo

The counterexample: `mulmod(2**128, 2**128, 3)` should be `3`, but in this implementation it is `0`, in fact, for every `m`.

murcake (changes requested):

murcake:
The PR description says that it implements `uint256_exp`, but it doesn't (am I missing something?)

murcake:
Yes, sorry, it does implement exponent, I overlooked it

PR #173, opened 2022-01-01
fix signextend

BitBaseBit:


murcake (commented):
Can you please, add a test to `warp/tests/cairo`?

PR #174, opened 2022-01-01
fix Exp in BuiltinHandler

BitBaseBit:


murcake:
This PR seems to repeat #170

PR #178, opened 2022-01-02
add dataoffset/size/copy builtins

BitBaseBit:


murcake (approved):

warp/yul/BuiltinHandler.py
```diff
@@ -598,6 +598,18 @@ class Create2(NotImplementedStarkNet):
     pass
 
 
+class DataOffset(NotImplementedStarkNet):
```
swapnilraj, line 601: How is this working, these builtins don't seem to be in the mapping

PR #180, opened 2022-01-03
fix sgt

BitBaseBit:


warp/cairo-src/evm/uint256.cairo
```diff
@@ -87,6 +87,10 @@ func is_eq{range_check_ptr}(op1 : Uint256, op2 : Uint256) -> (result : Uint256):
 end
 
 func is_gt{range_check_ptr}(op1 : Uint256, op2 : Uint256) -> (result : Uint256):
+    let (eq) = uint256_eq(op1, op2)
+    if eq == 1:
+        return (result=Uint256(0, 0))
+    end
```
cicr99, line 93: I think in this case you do not need to check for equality. You want to know if `op1 > op2` and you are returning `op2 < op1` when you do `let (res) = uint256_lt(op2, op1)`, which is supposed to be the same thing. Am I correct?

murcake: Yeah, I think this whole function should be `return is_lt(op2, op1)`

warp/cairo-src/evm/uint256.cairo
```diff
@@ -103,6 +107,10 @@ func slt{range_check_ptr}(op1 : Uint256, op2 : Uint256) -> (result : Uint256):
 end
 
 func sgt{range_check_ptr}(op1 : Uint256, op2 : Uint256) -> (result : Uint256):
+    let (eq) = uint256_eq(op1, op2)
+    if eq == 1:
+        return (result=Uint256(0, 0))
+    end
     let (res) = uint256_signed_lt(op1, op2)
     return (result=Uint256(1 - res, 0))
```
murcake, line 115: This one should be `return slt(op2, op1)`

PR #185, opened 2022-01-05
Encode strings manually

swapnilraj:


murcake (approved):
I'm a bit doubtful, because it makes generated code less readable. Still, generated code is not supposed to be read very often, so the tradeoff with simplified handling is okay, I think.
Also that automatically enables more optimizations, so that's good

swapnilraj:
> I'm a bit doubtful, because it makes generated code less readable. Still, generated code is not supposed to be read very often, so the tradeoff with simplified handling is okay, I think. Also that automatically enables more optimizations, so that's good

Yea strings are nice and easy to spot in the huge wall of numbers and warp_blocks, maybe later we can add the ability to generate comments in the cairo file.

PR #187, opened 2022-01-09
Benchmarking CI: Push commit history

Glitch18:
Modify the CI to push a text file containing the commit history. This is used by the benchmarking repository to build a dashboard with graphs showing Warp's progression.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ test_yul: warp
 benchmark: warp
 	mkdir -p benchmark/stats
 	mkdir -p benchmark/tmp
+	git log -20 --format=format:"%H" > ./benchmark/stats/commits.txt
 	python -m pytest tests/benchmark -v --tb=short --workers=auto $(ARGS)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -34,6 +34,7 @@ test_yul: warp
 benchmark: warp
 	mkdir -p benchmark/stats
 	mkdir -p benchmark/tmp
+	git log -20 --format=format:"%H" > ./benchmark/stats/commits.txt
```
swapnilraj, line 37: Why only the last 20 commits? We should have the ordering for all the commits and then the python script can decide how many to use for the graph.

Glitch18: Hey Swapnil. The 20 commits setting is just for testing as of now. I pushed it into the PR so that I could verify the CI. As of now there's some issue that is causing the CI to only retrieve the last commit hash. I'm working on a fix. I should have probably marked this PR as a draft, my bad.

PR #189, opened 2022-01-10
better error reporting for kudu failures

BitBaseBit:


warp/yul/main.py
```diff
@@ -45,8 +59,11 @@ def transpile_from_solidity(
             check=True,
             capture_output=True,
         )
+        if result.stderr:
+            print(bcolors.FAIL + bcolors.BOLD + result.stderr.decode("utf-8"))
+            sys.exit(1)
     except subprocess.CalledProcessError as e:
-        print(e.stderr.decode("utf-8"), file=sys.stderr)
+        print(bcolors.FAIL + bcolors.BOLD + e.stderr.decode("utf-8"), file=sys.stderr)
```
murcake, line 66: I am not an expert in ansi escape sequences, but don't they need a closing one? Otherwise, everything afterwards is also colored

BitBaseBit: Added closing (`bcolors.ENDC`)

warp/yul/main.py
```diff
@@ -45,8 +59,11 @@ def transpile_from_solidity(
             check=True,
             capture_output=True,
         )
+        if result.stderr:
+            print(bcolors.FAIL + bcolors.BOLD + result.stderr.decode("utf-8") + bcolors.ENDC)
+            sys.exit(1)
```
murcake, line 64: We probably don't want to exit, sometimes people print just warnings and stuff to stderr. If the app doesn't choose to fail (no throw), we shouldn't either

warp/yul/main.py
```diff
@@ -45,8 +59,18 @@ def transpile_from_solidity(
             check=True,
             capture_output=True,
         )
+        if result.stderr:
+            print(
+                bcolors.FAIL
+                + bcolors.BOLD
+                + result.stderr.decode("utf-8")
+                + bcolors.ENDC
+            )
```
murcake, line 68: Also needs `, file=sys.stderr`

murcake (approved):

PR #201, opened 2022-01-18
Add semanticTests with pedersen

Glitch18:
Add semanticTests substituting keccak256 with the pedersen built-in using inline assembly.

### DIFF
--- a/tests/behaviour/slot_access.sol
+++ b/tests/behaviour/slot_access.sol
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.6;
+
+contract WARP {
+    struct S {
+        uint256 a;
+        uint256 b;
+    }
+
+    mapping(uint256 => S) public mappingAccess;
+
+    function data() internal view returns (S storage _data) {
+        // We need to assign it from somewhere, otherwise we would
+        // get an "uninitialized access" error.
+        _data = mappingAccess[20];
+
+        assembly {
+            mstore(0, 1)
+            mstore(32, 0)
+            let slot := pedersen(0, 64)
+            _data.slot := slot
+        }
+    }
+
+    function set(uint256 x) public {
+        data().a = x;
+    }
+
+    function get() public view returns (uint256) {
+        return data().a;
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// get() -> 0
+// mappingAccess(uint256): 1 -> 0, 0
+// set(uint256): 4
+// get() -> 4
+// mappingAccess(uint256): 1 -> 4, 0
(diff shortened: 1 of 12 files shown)


murcake:
Can you please rebase onto the newest develop to fix CI?

tests/behaviour/array_storage_index_zeroed_test_test.py
```diff
@@ -0,0 +1,47 @@
...
+from starkware.starknet.testing.state import StarknetState
+from yul.main import transpile_from_solidity
+from yul.starknet_utils import deploy_contract, invoke_method
+
+warp_root = os.path.abspath(os.path.join(__file__, "../../.."))
+test_dir = __file__
+
+
+@pytest.mark.asyncio
+async def test_address():
```
murcake, line 14: Behaviour tests don't actually need cairo-files. We can get this cairo files by transpiling from solidity. Furthermore, it is useful as it allows to keep .sol and .cairo files out of sync. You can take a look at `for-loop-no-condition_test.py` to see how it can be done

tests/behaviour/array_storage_index_zeroed_test_test.py
```diff
@@ -0,0 +1,47 @@
...
+        starknet, program_info, contract_address, "test_zeroed_indicies", 10
+    )
+
+    res = await invoke_method(
+        starknet, program_info, contract_address, "test_zeroed_indicies", 15
+    )
+
+    # Following test fails as of now
+    # res = await invoke_method(
+    #     starknet, program_info, contract_address, "test_zeroed_indicies", 0xFF
```
murcake, line 46: Hmm, why do they fail? We need to investigate that

Glitch18:
> Can you please rebase onto the newest develop to fix CI?
Done

tests/behaviour/array_storage_index_zeroed_test_test.py
```diff
@@ -0,0 +1,47 @@
+import os
+
+import pytest
+from starkware.starknet.compiler.compile import compile_starknet_files
+from starkware.starknet.testing.state import StarknetState
+from yul.main import transpile_from_solidity
+from yul.starknet_utils import deploy_contract, invoke_method
```
murcake, line 7: We know have a different package structure. All of our modules are placed inside of warp package, so you should prefix imports with `warp.`

tests/behaviour/array_storage_index_zeroed_test_test.py
Glitch18: Have removed that test as of now. Will add it once we've figured out why it's failing

tests/behaviour/slot_access.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    function get() public view returns (uint256) {
+        return data().a;
+    }
+}
+// ====
+// compileViaYul: also
+// ----
+// get() -> 0
+// mappingAccess(uint256): 1 -> 0, 0
```
swapnilraj, line 37: There is no corresponding call in tests for this?

Glitch18: Ah my bad. Must have missed it. Have updated the PR with this. 👍

murcake (approved):

PR #202, opened 2022-01-18
use poetry

murcake:
Problem: we neither completely lock our dependencies nor specify them
     abstractly. That causes two issues. When something updates on
     PyPI builds get broken, because dependencies don't match. The
     other issue is when someone uses warp with library versions
     different from the ones we specify. If they use it as a library,
     and they want to use it together with another library that has
     another dependency range, it's impossible.

Solution: use lock files. We specify abstract dependencies without
     versions in pypyroject.toml and let poetry create a lock
     file. This way we always have concrete dependency versions that
     are guaranteed to work, but also anyone can try to use other
     versions.

Why not pipenv? The dependency resolution UX seems better with poetry,
     also it seems to be faster.


BitBaseBit (approved):

JorikSchellekens:
Can you update the docs to reflect the need to have poetry installed?

murcake:
> Can you update the docs to reflect the need to have poetry installed?

It's only necessary for development. At the moment there's nothing about development setup in the readme. It would be useful to write about it for contributors (and new team members), but I think that such work belongs to another PR

PR #203, opened 2022-01-18
udpate docs

BitBaseBit:


README.md
```diff
@@ -127,12 +127,55 @@ declaration order (i.e `age` first, `person` second). If the first argument to t
 --inputs = "[[42,1722,7], (26, 200), 7432533831]"
 ```
 
+If you're passing in the `bytes` Solidity type as an argument, use the python syntax, for example:
+```bash
+--inputs = "[[10,20], b"\x01\x02"]"
```
murcake, line 132: I don't think it's goint to work. Bash doesn't support nested quotes. The inner quotes must be single, while outer — double, or vice versa

README.md
```diff
@@ -127,12 +127,55 @@ declaration order (i.e `age` first, `person` second). If the first argument to t
...
 
+## Solidity Constructs Currently Not Supported
+
+
+|  Support Status                 | Symbol            | 
+|:-------------------------------:|:-----------------:|
+| Will likely never be supported  | :x:               |
+| Support will land soon          | :hammer_and_pick: |
+| Will be supported in the future | :question:        |
+| Currently Unkown                | :exclamation:     |
```
murcake, line 149: I'd swap :question: and :exclamation:, seems more intuitive

PR #204, opened 2022-01-18
Block timestamp  & number

BitBaseBit:


tests/behaviour/block_data_test.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    block_number = await invoke_method(
+        starknet, block_data_info, block_data_address, "getBlockNumber"
+    )
+    assert block_number.retdata == [
+        32,
+        2,
+        0,
+        3618502788666131213697322783095070105623107215331596699973092056135872020480,
```
murcake, line 40: What is that number? How do we know it's going to be it?

murcake (approved):

tests/behaviour/block_data_test.py
BitBaseBit: Not sure, if you run this on an empty starknet instance it gives you the same number:
```
%lang starknet
from starkware.starknet.common.syscalls import get_block_number, get_block_timestamp

@external
func testMe{syscall_ptr : felt*}() -> (res: felt):
    let (res_1) = get_block_number()
    return (res=res_1)
end
```

PR #206, opened 2022-01-19
Remove builtins directive

BitBaseBit:


tests/golden/c2c_marked.cairo
```diff
@@ -0,0 +1,476 @@
+%lang starknet
```
murcake, line 1: This file shouldn't be there

tests/behaviour/malicious-calldata_test.py
```diff
@@ -1,109 +0,0 @@
-import tempfile
```
murcake, line 1: Why is this file deleted?..

pyproject.toml
```diff
@@ -24,7 +24,7 @@ skip_gitignore = true
 
 [tool.poetry]
 name = "sol-warp"
-version = "0.2.12"
```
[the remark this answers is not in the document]
BitBaseBit: Because we've been forgetting to bump it for months lol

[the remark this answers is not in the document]
murcake: We didn't forget, it's just that there haven't been a release. It's normal to do a lot of changes without bumping the version until the new release

tests/behaviour/malicious-calldata_test.py
BitBaseBit: Because it was written in a way that was completely incongruent with the style of all the other tests.

murcake: But it needs to be there, it tests important things?.. The whole Uint256 library is based on the premise that it's given valid Uint256's, so we need to check for that. It's written that way, because unlike other tests it tests failure.

murcake: It also uses fixtures, in order to avoid computing the same things again and again

BitBaseBit: Fair enough, I'll add it back :).

murcake (approved):

murcake:
Let's rebase onto develop first, though

PR #209, opened 2022-01-20
Fix CLI status query

cicr99:
- Added network as a required argument
- Used `starknet tx_status` command for the query
- Added missing sections to the Table of Contents in the `readme` file


src/warp/cli/commands.py
```diff
@@ -4,7 +4,6 @@
 from typing import Any, Dict, Optional, Sequence, Union
 
 import aiohttp
```
murcake, line 6: You can now remove the import of `aiohttp` from here, and from the dependencies!

murcake (approved):

PR #212, opened 2022-01-20
Fail fast on unsuported solidity semantics

JorikSchellekens:
Some of these constructs were being compiled to nonsense cairo. For example we implement `callvalue` for a specific case which resulted in all contracts containing `msg.value` to be compiled.

The solution is to filter for these regexes to make sure we enforce that these won't be compiled.

### DIFF
--- a/src/warp/yul/source_filter.py
+++ b/src/warp/yul/source_filter.py
@@ -1,0 +1,58 @@
+import dataclasses
+import re
+import sys
+from os.path import exists
+
+
+@dataclasses.dataclass()
+class bcolors:
+    HEADER = "\033[95m"
+    OKBLUE = "\033[94m"
+    OKCYAN = "\033[96m"
+    OKGREEN = "\033[92m"
+    WARNING = "\033[93m"
+    FAIL = "\033[91m"
+    ENDC = "\033[0m"
+    BOLD = "\033[1m"
+    UNDERLINE = "\033[4m"
+
+
+readme_link = (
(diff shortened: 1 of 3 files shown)


BitBaseBit (approved):

src/warp/yul/source_filter.py
```diff
@@ -0,0 +1,58 @@
...
+    (r"\bblock.gaslimit\b", "Warp doesn't support block.gaslimit."),
+    (r"\bgasleft\b", "Warp doesn't support gasleft."),
+    (r"\bselfdestruct\b", "Warp doesn't support selfdestruct."),
+    (r"\bblockhash\b", "Warp doesn't support blockhash."),
+    (
+        r"\bkeccak256\b",
+        "Warp doesn't support keccak since it's too slow on starknet. Please consider using the pedersen primitive.",
+    ),
+    (
+        r"\btry\b",
```
murcake, line 41: This feels like it'll catch a loot of comments

src/warp/yul/source_filter.py
```diff
@@ -0,0 +1,58 @@
...
+        "Warp doesn't support try/catch since it's too slow on starknet. Please consider using the pedersen primitive.",
```
murcake, line 42: The message is weird)

murcake:
Now that we have `https://github.com/NethermindEth/nethersolc/pull/32` which can't catch false positives, should this PR be closed?

PR #215, opened 2022-01-21
Release 1.0

JorikSchellekens:
Adds a 'call' command and a cleaner error output


BitBaseBit (approved):

JorikSchellekens:
`weth10` is meant to fail because it references `msg.value`.

murcake:
I'd suggest first uploading to https://test.pypi.org/ before uploading to pypy.org

murcake:
Does it work on all macos versions?

PR #220, opened 2022-01-25
Remove redundent macos copies

JorikSchellekens:
Turns out that all the different `kudu` and `isoltest` bins for macos work on all the macos platforms so I've eliminated the different version and have grouped them collectively under `darwin`. This simplifies our release process because we just need to build one.


murcake (approved):
Great!

PR #222, opened 2022-01-27
Add arm linux binaries for isoltest and nethersolc

swapnilraj:
Our CI runners for semantic tests will be arm linux machines, so we need binaries for that platform

### DIFF
--- a/scripts/isoltest
+++ b/scripts/isoltest
@@ -14,10 +14,13 @@ from importlib_resources import as_file, files
 exec_name = "isoltest"
 
 
-def get_system_suffix(system: str) -> Optional[str]:
+def get_system_suffix(system: str, machine: str) -> Optional[str]:
     system = system.lower()
     if system == "linux":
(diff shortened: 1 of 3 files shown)


murcake:
So, does that mean that on every kudu update in develop we'll have to update binaries for both x86 linux and arm linux?

swapnilraj:
> So, does that mean that on every nethersolc update in develop we'll have to update binaries for both x86 linux and arm linux?

Yes, it shouldn't be that bad because the aws instances that was recommended to use were arm linux machines

murcake:
> > So, does that mean that on every nethersolc update in develop we'll have to update binaries for both x86 linux and arm linux?
> 
> Yes, it shouldn't be that bad because the aws instances that was recommended to use were arm linux machines

In that case, the PR that builds both solc and isoltest on ARM should quickly follow this one. 

Is this PR a preparation for running eth test-suite in CI?

PR #223, opened 2022-01-27
adding codedeploy config, codeowners file, scripts to test semantics …

matilote:
…and temporary copy binaries


scripts/copy_binaries.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# TEMP - Moves isoltest, kudu arm64 binaries to destination directory
+sudo cp /home/ubuntu/warp/src/warp/bin/linux/isoltest /home/ubuntu/warp-temporary/src/warp/bin/linux/isoltest
+sudo cp /home/ubuntu/warp/src/warp/bin/linux/kudu /home/ubuntu/warp-temporary/src/warp/bin/linux/kudu
```
murcake, line 4: We no longer use `kudu`, just `nethersolc`

PR #228, opened 2022-02-03
Make block_data explicit in behaviour test

swapnilraj:


tests/behaviour/block_data_test.py
```diff
@@ -26,6 +27,8 @@ async def test_starknet():
     )
 
     starknet = await StarknetState.empty()
+    starknet.state.block_info = BlockInfo(0xDEADBEEF, 0xCAFEBABE)
```
murcake, line 30: I'd prefer to give explicit names to parameters (`block_number=0xdeadbeef, block_timestamp=…` )

murcake (approved):

tests/behaviour/block_data_test.py
murcake: (it just wasn't very clear at a first glance which was which)

swapnilraj: fixed

PR #263, opened 2022-03-16
Implement benchmarking

Glitch18:
Implement benchmarking for a dashboard

### DIFF
--- a/starknet-testnet/generateMarkdown.py
+++ b/starknet-testnet/generateMarkdown.py
@@ -1,0 +1,107 @@
+import json
+import os
+import sys
+
+from pathlib import Path
+from starkware.starknet.business_logic.internal_transaction_interface import (
+    TransactionExecutionInfo,
+)
+
+WARP_ROOT = Path(__file__).parents[1]
+TMP = WARP_ROOT / "benchmark" / "json"
+FILE_NAME = "data"
+
+
+def steps_in_function_deploy(contract_name: str, result: TransactionExecutionInfo):
+    json_path = os.path.abspath(TMP / (FILE_NAME + ".json"))
+    if os.path.exists(json_path):
+        with open(json_path, "r") as json_file:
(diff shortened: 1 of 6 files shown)


starknet-testnet/generateMarkdown.py
```diff
@@ -0,0 +1,107 @@
...
+
+def steps_in_function_invoke(contract_name: str, result: TransactionExecutionInfo):
+    json_path = os.path.abspath(TMP / (FILE_NAME + ".json"))
+    if os.path.exists(json_path):
+        with open(json_path, "r") as json_file:
+            benchmark_data = json.load(json_file)
+    else:
+        benchmark_data = {}
+
+    benchmark_data.setdefault("_".join(contract_name.split("_")[:-1]), {})[
```
DomHenderson, line 39: What's the reasoning behind the name processing here?

starknet-testnet/server.py
```diff
@@ -63,6 +70,8 @@ async def invoke():
         print("-----Invoke info-----")
         print(execution_info)
         print("----------\n")
+        steps_in_function_invoke(data["function"], execution_info)
```
DomHenderson, line 73: Would it be possible (and not too time consuming) to add a flag to the testnet such that the benchmarking is only done when the flag is passed, and have the CI pass the flag? That way when we're running lots of behaviour tests during development it won't get slowed down

starknet-testnet/generateMarkdown.py
Glitch18: The names that are passed into `generateMarkdown.py` from the starknet testnet invoke command are of this format `unreachableCode_0d583609`
The above snippet was to keep everything as is but for the suffix. So that the names passed into the JSON/markdown are of the format `unreachableCode`

DomHenderson (approved):

PR #267, opened 2022-03-17
Remove Named Arguments from function Call

RohitRanjangit:
This pr makes us get rid of named arguments in a function call. Consider the following solidity contract:
```solidity
contract C {
    mapping(uint => uint) data;

    function f() public {
        set({value: 2, key: 3});  //named arg function call
    }

    function set(uint key, uint value) public {
        data[key] = value;
    }

}
```
The above contract will get transformed into to :
```solidity
contract C {
    mapping(uint => uint) internal data;

    function f() public {
        set(3, 2); // removed argument name, args in the correct order
    }

    function set(uint key, uint value) public {
        data[key] = value;
    }
}
```

This pr also fixes the Reference Declaration in case of the getter function call for a public state variable. It has been changed to `FunctionDefinition` corresponding to the getter function from `VariableDeclaration`.


src/passes/fixFunctionCallReferences/index.ts
```diff
@@ -0,0 +1,18 @@
+import { AST } from '../../ast/ast';
+import { ASTMapper } from '../../ast/mapper';
+
+import { FunctionDefinition } from 'solc-typed-ast';
+
+import { GetGetters } from './getGetters';
+import { FixFnCallRef } from './fixFnCallRef';
+
+export class FixFunctionCallReferences extends ASTMapper {
```
DomHenderson, line 9: Instead of having a separate top-level pass for this, I think it would be better to combine GettersPublicStateVars and FixFnCallRef into a compound pass like this one, with GettersPublicStateVars building up the getterFunctions map as it goes. This would also require the compound pass to run each pass using their map function rather than applying each one to each root

src/testing.ts
```diff
@@ -69,6 +69,9 @@ const expectedResults = new Map<string, ResultType>([
   ['example_contracts/mutableReferences/memory', 'Success'],
   ['example_contracts/mutableReferences/mutableReferences', 'NotSupportedYet'],
   ['example_contracts/mutableReferences/scalarStorage', 'Success'],
+  ['example_contracts/namedArgs/constructor', 'NotSupportedYet'],
+  ['example_contracts/namedArgs/events_and_errors', 'Success'],
+  ['example_contracts/namedArgs/function', 'Success'],
```
DomHenderson, line 74: These are good example contracts, but could we also have some behaviour tests for the cases which are currently supported, to verify that functions get called with the right arguments in the right order after the transform

src/passes/namedArgsRemover.ts
```diff
@@ -0,0 +1,97 @@
...
+        `Number of field names and number of arguments are not equal.`,
+      );
+    }
+
+    orderedFieldNames.forEach((fieldName, index) => {
+      //find the index of the field name in the fieldNames array
+      const fieldNameIndex = fieldNames.findIndex((name) => name === fieldName);
+      //if the index is not found, throw an error
+      if (fieldNameIndex === -1) {
+        throw new TranspilationAbandonedError(
```
DomHenderson, line 78: TranspilationAbandonedError should generally not be instantiated directly, the subclasses provide more detail as to why the transpilation was abandoned. The ones introduced in this PR should be TranspileFailedError instead

src/passes/namedArgsRemover.ts
```diff
@@ -0,0 +1,97 @@
...
+  TranspilationAbandonedError,
+  WillNotSupportError,
+} from '../utils/errors';
+
+export class NamedArgsRemover extends ASTMapper {
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    /*
+            Visit every function call and remove the named arguments
+        */
+    //console.log(node.vExpression);
```
DomHenderson, line 23: This should be deleted

src/passes/fixFunctionCallReferences/fixFnCallRef.ts
```diff
@@ -0,0 +1,39 @@
...
+        /*
+                    Getter function of a public state variable can be 
+                    only invoked through using this.`functionName` only
+                */
+        if (!(node.vExpression instanceof MemberAccess)) {
+          throw new TranspilationAbandonedError('FixFnCallRef: vExpression is not a MemberAccess');
+          ``;
+        }
+
+        node.vExpression.vReferencedDeclaration = this.getterFunctions.get(
```
DomHenderson, line 32: I think there should either be an assertion before this that getterFunctions contains the given id, at present this will silently set vReferencedDeclaration to undefined if the map is not comprehensive

DomHenderson (changes requested):

DomHenderson (approved):

PR #268, opened 2022-03-17
Fixed returnInserter

Purva-Chaudhari:
Added ReturnVariableInitializer pass, fixed return Inserter for conditions


tests/behaviour/expectations/behaviour.ts
```diff
@@ -342,6 +342,19 @@ export const expectations = flatten(
             Expect.Simple('doWhile_break', ['0', '2'], ['2']),
           ]),
         ]),
+        new Dir('returns', [
+          File.Simple('returnInserter', [
+            Expect.Simple('test1', ['6'], ['0']),
```
DomHenderson, line 347: These tests could do with more descriptive names so that people can tell which case each one is testing for without having to code and analyse the contract code. It would also be helpful to add a tag to the Expects that use the same function

tests/behaviour/expectations/behaviour.ts
```diff
@@ -342,6 +342,19 @@ export const expectations = flatten(
...
+            Expect.Simple('test2', ['1'], ['2']),
+            Expect.Simple('test2', ['2'], ['0']),
+            Expect.Simple('test3', ['3'], null),
+            Expect.Simple('test5', ['3', '5'], ['8']),
```
DomHenderson, line 351: Test4 and test6 should either be added here or removed from the contract

DomHenderson (changes requested):

DomHenderson (approved):

PR #269, opened 2022-03-18
README: add 'installing from sources' section

sblOWPCKCR:
published pip package is too old to be useful


BitBaseBit:
The latest version is already on pip

PR #270, opened 2022-03-18
Sanitize External Inputs (Enums, Boolean and Int)

GuyMcComb:
This PR adds bound checks for external/public functions that accept scalar data types.

Each type of input check has a separate pass. The int bound checks have been moved from the CairoWriter to their own pass now as well.


src/cairoUtilFuncGen/enumBoundCheck.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  gen(enumVarDec: VariableDeclaration, functionInput: Expression): FunctionCall {
+    const enumType = enumVarDec.vType;
+    assert(enumType instanceof UserDefinedTypeName);
+    const enumDef = enumType.vReferencedDeclaration;
+
+    assert(enumDef instanceof EnumDefinition);
+    const name = this.getOrCreate(enumDef);
+    const intType = new IntType(8, false);
```
GuyMcComb, line 32: Is this the correct type for an `enum`, since it is an `int` behind the scenes?

DomHenderson: Usually, but not always, it can be a larger type if the enum has many members. I'd recommend using `enumToTypeInt` and `typeNameFromTypeNode` like in enumConverter to generate the appropriate typeName

src/passes/externalInputChecker/enumBoundChecker.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+    this.commonVisit(node, ast);
+  }
+
+  private generateFunctionCall(
+    node: FunctionDefinition,
+    parameter: VariableDeclaration,
+    ast: AST,
+  ): FunctionCall {
+    const enumStubArgument = new Identifier(
```
DomHenderson, line 32: (also applicable to all 3 files) There's a util function in nodeTemplates for this, `createIdentifier`

src/passes/externalInputChecker/enumBoundChecker.ts
```diff
@@ -0,0 +1,67 @@
...
+  Statement,
+  ExpressionStatement,
+  VariableDeclaration,
+  FunctionCall,
+} from 'solc-typed-ast';
+import { ASTMapper } from '../../ast/mapper';
+
+export class EnumBoundChecker extends ASTMapper {
+  visitFunctionDefinition(node: FunctionDefinition, ast: AST): void {
+    if (FunctionVisibility.External === node.visibility) {
```
DomHenderson, line 16: In all 3 files, this should include a check that the function has a body, doesn't make sense to try to these checks into undefined functions

src/passes/externalInputChecker/intBoundChecker.ts
```diff
@@ -0,0 +1,86 @@
...
+import { createCairoFunctionStub, createCallToFunction } from '../../utils/functionStubbing';
+import { typeNameFromTypeNode } from '../../utils/utils';
+import assert = require('assert');
+
+export class IntBoundChecker extends ASTMapper {
+  visitFunctionDefinition(node: FunctionDefinition, ast: AST): void {
+    if (FunctionVisibility.External === node.visibility) {
+      node.vParameters.vParameters.forEach((parameter) => {
+        if (
+          (parameter.typeString.slice(0, 4) === 'uint' ||
```
DomHenderson, line 23: This is minor and not required, but it might be nicer to do getNodeType on the parameter, check if it's an IntType, then use the width stored in the IntType rather than manually parsing typeStrings

tests/behaviour/contracts/external_input_checks/bool_.sol
```diff
@@ -0,0 +1,22 @@
+pragma solidity ^0.8.6;
+
```
DomHenderson, line 2: No need for underscores at the end of contract names because test filter can include directories

warplib/maths/external_input_check_bool.cairo
```diff
@@ -0,0 +1,10 @@
+from starkware.cairo.common.math_cmp import is_le_felt
+
+func warp_external_input_check_bool{range_check_ptr}(x : felt):
+    let (inRange : felt) = is_le_felt(x, 1)
```
DomHenderson, line 4: Worth checking the step count to be sure, but I think `assert x * (x-1) = 0` would be more efficient

DomHenderson (changes requested):

src/cairoUtilFuncGen/enumBoundCheck.ts
GuyMcComb: Yeah I agree. I also saw that the upper bound for enum members was much higher than int8.

warplib/maths/external_input_check_bool.cairo
GuyMcComb: Yeah, you are right. Changing out that assert statement brings the step count down from 46->11.

src/passes/externalInputChecker/enumBoundChecker.ts
GuyMcComb: Done

src/passes/externalInputChecker/enumBoundChecker.ts
GuyMcComb: Added.

src/passes/externalInputChecker/intBoundChecker.ts
GuyMcComb: Replaced it.

DomHenderson (approved):

PR #271, opened 2022-03-18
If functionalisation

DomHenderson:
Handles the revoked reference issues that can come with ifs by packaging everything happening after an if into a function and calling it at the end of each branch


GuyMcComb (approved):
LGTM. Most points covered in Slack and accompanying tests are in place from previous attempts.

PR #272, opened 2022-03-18
Handle function modifiers

cicr99:


tests/behaviour/expectations/behaviour.ts
```diff
@@ -986,6 +986,20 @@ export const expectations = flatten(
           Expect.Simple('test', [], []),
           Expect.Simple('returnTest', [], ['12', '0']),
         ]),
+        new Dir('modifiers', [
```
DomHenderson, line 989: Test expectations should be laid out in alphabetical order at each level (at least for all the Files and Dirs, doesn't really matter for the Expects). With this many tests, they get lost otherwise

tests/behaviour/expectations/behaviour.ts
```diff
@@ -986,6 +986,20 @@ export const expectations = flatten(
...
+            Expect.Simple('f', ['90000', '0'], ['10000', '0']),
+            Expect.Simple('f', ['110000', '0'], ['0', '0']),
+          ]),
+          File.Simple('multipleModifiers', [
+            new Expect('modifier', [
+              ['openEvent', [], [], '0'],
+              ['donate', ['238', '0'], ['238', '0'], '0'],
+              ['donate', ['100', '0'], ['338', '0'], '0'],
+              ['closeEvent', [], [], '0'],
+            ]),
```
DomHenderson, line 1000: I'd like to see at least one test showing that each modifier is applied here. For example, one where donate is called after closeEvent to show that `isOpen` is applied correctly, and one with value < 100 to show that `cost` is applied correctly

src/utils/nodeTemplates.ts
```diff
@@ -81,3 +82,10 @@ export function createUint256TypeName(ast: AST): ElementaryTypeName {
   ast.setContextRecursive(typeName);
   return typeName;
 }
+
+export function createParameterList(
+  params: Iterable<VariableDeclaration>,
+  ast: AST,
+): ParameterList {
+  return new ParameterList(ast.reserveId(), '', 'ParameterList', params);
```
DomHenderson, line 90: Very much like to see more functions getting added here, I'm hoping over time a large number of the raw constructors used will get replaced with such functions. I do think it would be worth calling `ast.setContextRecursive` on the generated parameterList though

src/utils/cloning.ts
```diff
@@ -410,13 +411,16 @@ function cloneASTNodeImpl<T extends ASTNode>(
...
+  } else if (node instanceof PlaceholderStatement) {
+    newNode = clonePlaceholder(node, ast, remappedIds);
   }
 
   if (notNull(newNode) && sameType(newNode, node)) {
     ast.setContextRecursive(newNode);
     ast.copyRegisteredImports(node, newNode);
     return newNode;
   } else {
+    console.log(node);
```
DomHenderson, line 423: Is this print still needed, given that the error below includes which node was not cloneable?

src/utils/cloning.ts
```diff
@@ -410,13 +411,16 @@ function cloneASTNodeImpl<T extends ASTNode>(
       [...node.vOverrides].map((o) => cloneASTNodeImpl(o, ast, remappedIds)),
       node.raw,
     );
+  } else if (node instanceof PlaceholderStatement) {
```
DomHenderson, line 414: For ease of navigation the if-else chain is broken up into typed groups, this should go under Statements, in alphabetical order with the other statement types

src/passes/functionModifierHandler/modifierRemover.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ContractDefinition, ModifierDefinition } from 'solc-typed-ast';
+import { AST } from '../../ast/ast';
+import { ASTMapper } from '../../ast/mapper';
+import { printNode } from '../../utils/astPrinter';
+
+export class ModifierRemover extends ASTMapper {
+  visitModifierDefinition(node: ModifierDefinition, _ast: AST) {
+    const parent = node.getClosestParentByType(ContractDefinition);
+    assert(parent !== undefined, `Unable to find parent of ${printNode(node)}`);
+    parent?.removeChild(node);
```
DomHenderson, line 11: The `?` is unnecssary after the assert has asserted that parent is defined

warplib/maths/external_input_checks.cairo
```diff
@@ -0,0 +1,265 @@
+# AUTO-GENERATED
```
DomHenderson, line 1: warplib files starting with AUTO-GENERATED should never be added to git. Generally the warplib gitignore handles this, but this is an old file that doesn't get generated anymore, and can be safely deleted

DomHenderson (changes requested):
This is good work on a complex pass, not too much more left until we can get it merged

PR #273, opened 2022-03-18
modify external contract method call

RohitRanjangit:
modify cairoWriter such that calls to methods on contracts get changed to the form the cairo expects. E.g. `a.f(arg1)` in solidity becomes `ContractType.f(a, arg1)`

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -31,6 +31,7 @@ import {
   FunctionStateMutability,
   FunctionTypeName,
   FunctionVisibility,
+  getNodeType,
   Identifier,
   IdentifierPath,
   IfStatement,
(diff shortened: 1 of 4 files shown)


src/cairoWriter.ts
```diff
@@ -459,18 +460,37 @@ class IdentifierWriter extends CairoASTNodeWriter {
...
-        const args = node.vArguments.map((v) => writer.write(v)).join(', ');
-        const func = writer.write(node.vExpression);
         const arg = node.vArguments[0];
         if (node.vFunctionName === 'address' && arg instanceof Literal) {
           const val: BigInt = BigInt(arg.value);
           // Make sure literal < 2**251
           assert(val < BigInt('0x800000000000000000000000000000000000000000000000000000000000000'));
           return [`${args[0]}`];
         }
+        if (node.vFunctionName === 'contract') {
```
DomHenderson, line 491: I think this should be a `getNodeType` check on the function expression

DomHenderson (changes requested):
Just one request for this one

DomHenderson (approved):

PR #280, opened 2022-03-29
Public Function Split

GuyMcComb:
Feature to split Public functions into internal and external functions. 
PR uses 2 passes:
1. Modifies the public function to be internal and clones it to create the external function with a function call to the internal function.
2. Visits all the function calls to make sure they are pointed to the correct internal/external function.


tests/behaviour/expectations/behaviour.ts
```diff
@@ -848,6 +848,12 @@ export const expectations = flatten(
             Expect.Simple('uint256write', ['5', '6'], ['0', '0', '5', '6']),
           ]),
         ]),
+        new Dir('public_func_splitter', [
```
GuyMcComb, line 851: Will move to the correct alphabetical position.

GuyMcComb:
Might get rid of the cloning functions since they are not used in this pass. If it is decided they will stay I will move them to the correct alphabetical positions.

tests/behaviour/contracts/public_func_splitter/contract_to_contract.sol
```diff
@@ -0,0 +1,22 @@
+pragma solidity ^0.8.0;
```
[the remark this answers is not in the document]
DomHenderson: It should go in example contracts until we can do external calls

tests/behaviour/expectations/behaviour.ts
GuyMcComb: Done

src/passes/publicFunctionSplitter/externalFunctionCreator.ts
```diff
@@ -0,0 +1,85 @@
...
+      this.modifyPublicFunction(node);
+      this.InternalToExternalFunctionMap.set(node, newExternalFunction);
+    }
+    this.commonVisit(node, ast);
+  }
+
+  private modifyPublicFunction(node: FunctionDefinition): FunctionDefinition {
+    node.visibility = FunctionVisibility.Internal;
+    node.name = node.name + this.suffix;
+    return node;
```
DomHenderson, line 45: As far as I can tell this can be a void function

src/passes/publicFunctionSplitter/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import { AST } from '../../ast/ast';
+import { ASTMapper } from '../../ast/mapper';
+import { FunctionDefinition } from 'solc-typed-ast';
+
+export class PublicFunctionSplitter extends ASTMapper {
+  static map(ast: AST): AST {
+    const internalToExternalFunctionMap = new Map<FunctionDefinition, FunctionDefinition>();
+    ast.roots.forEach((root) => {
+      new ExternalFunctionCreator(internalToExternalFunctionMap).dispatchVisit(root, ast);
+      new PublicFunctionCallModifier(internalToExternalFunctionMap).dispatchVisit(root, ast);
```
DomHenderson, line 12: This should dispatch ExternalFunctionCreator on all roots, and then use PublicFunctionCallModifier. Otherwise PublicFunctionCallModifier can visit function calls before ExternalFunctionCreator has visited their definition

src/utils/cloning.ts
```diff
@@ -157,6 +159,19 @@ function cloneASTNodeImpl<T extends ASTNode>(
...
+    const tupleComponents = node.vOriginalComponents.map((component) => {
+      return component !== null ? cloneASTNodeImpl(component, ast, remappedIds) : null;
+    });
+    newNode = new TupleExpression(
+      replaceId(node.id, ast, remappedIds),
+      node.src,
+      'TupleExpression',
+      node.typeString,
+      node.isInlineArray,
+      tupleComponents,
```
DomHenderson, line 172: The non-null elements need to be cloned with `cloneASTNodeImpl`

DomHenderson (changes requested):

DomHenderson:
> Might get rid of the cloning functions since they are not used in this pass. If it is decided they will stay I will move them to the correct alphabetical positions.

I'd prefer you keep them if you don't mind, they'll be needed later down the line

src/passes/publicFunctionSplitter/externalFunctionCreator.ts
GuyMcComb: Yeah good spot.

src/passes/publicFunctionSplitter/index.ts
GuyMcComb: Yeah, that is a good shout!

src/utils/cloning.ts
GuyMcComb: Looks like the none-null elements are being cloned with that function, are you talking about the null ones?

DomHenderson: Right, sorry, my default check is to make sure that there's no ASTNodes that get passed to the constructor without being wrapper in cloneASTNodeImpl. I managed to completely miss that they're being cloned further up, this is all good

DomHenderson (approved):

PR #285, opened 2022-03-31
ReferencedLibraries

Purva-Chaudhari:
Pass adds referneced libraries to `linearizedBaseList`

### DIFF
--- a/tests/behaviour/expectations/behaviour.ts
+++ b/tests/behaviour/expectations/behaviour.ts
@@ -15,58 +15,6 @@ export const expectations = flatten(
             Expect.Simple('test256', ['3', '4'], ['3', '4']),
           ]),
         ]),
-        new Dir('conversions', [
-          File.Simple('signedIdentity', [
-            Expect.Simple('implicit', ['210', '11', '12'], ['210', '11', '12']),
-            Expect.Simple('explicit', ['200', '300', '400'], ['200', '300', '400']),
-          ]),
-          File.Simple('signedNarrowing', [
-            Expect.Simple(
-              'explicit',
-              ['63005', '1000', '2000'],
-              ['29', '70778732319555200400381918345807787983848'],
-            ),
-          ]),
-          File.Simple('signedWidening', [
-            Expect.Simple('implicit', ['10'], ['10', '10', '0']),
-            Expect.Simple(
-              'implicit',
-              ['250'],
-              [
-                '65530',
-                '340282366920938463463374607431768211450',
-                '340282366920938463463374607431768211455',
-              ],
-            ),
-            Expect.Simple('explicit', ['20'], ['20', '20', '0']),
-            Expect.Simple(
-              'explicit',
-              ['240'],
-              [
-                '65520',
-                '340282366920938463463374607431768211440',
-                '340282366920938463463374607431768211455',
-              ],
-            ),
-          ]),
-          File.Simple('unsignedIdentity', [
-            Expect.Simple('implicit', ['210', '11', '12'], ['210', '11', '12']),
-            Expect.Simple('explicit', ['200', '300', '400'], ['200', '300', '400']),
-          ]),
-          File.Simple('unsignedNarrowing', [
-            Expect.Simple(
-              'explicit',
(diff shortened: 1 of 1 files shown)


example_contracts/libraries/using_for_simple.sol
```diff
@@ -21,3 +13,12 @@ contract WARP {
     return t.amIReal(d);
   }
 }
+
+library Nums {
```
DomHenderson, line 17: What's the reasoning behind moving this?

src/passes/referencedLibraries.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ContractDefinition, ContractKind, FunctionCall, MemberAccess } from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+
+export class ReferencedLibraries extends ASTMapper {
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    if (node.vExpression instanceof MemberAccess) {
+      const contractDef = new Map();
+
+      node.vExpression.context?.map.forEach((astNode, key) => {
```
DomHenderson, line 11: It's best to use `ast.context` in situations where a context is needed to remove the possibility of it being undefined

src/passes/referencedLibraries.ts
```diff
@@ -0,0 +1,34 @@
+import assert = require('assert');
+import { ContractDefinition, ContractKind, FunctionCall, MemberAccess } from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+
+export class ReferencedLibraries extends ASTMapper {
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    if (node.vExpression instanceof MemberAccess) {
+      const contractDef = new Map();
```
DomHenderson, line 9: ContractDef should have a type annotation to specify the key and value types of the map. This would improve readability and remove the need for asserting the type of the node later on

src/passes/referencedLibraries.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      node.vExpression.context?.map.forEach((astNode, key) => {
+        if (astNode instanceof ContractDefinition && astNode.kind === ContractKind.Library) {
+          contractDef.set(key, astNode);
+        }
+      });
+
+      const func_id = node.vExpression.referencedDeclaration;
+
+      contractDef.forEach((astNode, key) => {
```
DomHenderson, line 19: This is clever, and as far as I can tell a completely valid way of doing this, but a comment block explaining how the function works would aid maintainability I think

src/passes/referencedLibraries.ts
```diff
@@ -0,0 +1,34 @@
...
+        assert(astNode instanceof ContractDefinition);
+        if (astNode.vFunctions.some((func) => func.id === func_id)) {
+          const parent = node.getClosestParentByType(ContractDefinition);
+
+          if (parent === undefined) return;
```
DomHenderson, line 24: This brings up an interesting point, I was mistaken in believeing libraries could not be called from free functions. Handling that will be a separate PR I think, for now this should throw a NotSupportedYet exception if the library function call is not in a contract

DomHenderson (changes requested):
This is looking good, just a few small changes needed. Also could you add some behaviour tests for libraries? I believe this pass was the last thing required for them to be used

example_contracts/libraries/using_for_simple.sol
Purva-Chaudhari: Was using `visitContractDefinition` prior for a part to accomplish it. It was a check for library declared after contract. Can we keep it this was for behaviour testing?

DomHenderson: I'd suggest having one behaviour test with it declared before, one with it declared after, and if it works or is easily fixable, one with it being imported from an adjacent file. Might as well cover all bases, especially as the sourceUnitSplitter is liable to change in the future

DomHenderson (approved):

PR #286, opened 2022-03-31
DictAccess memory model

DomHenderson:
Primarily, this PR moves our memory model from an inefficient list of write objects that were recursively checked for each read, to a DictAccess model.

Additionally:
- StringIndexedFuncGen introduced to reduce code duplication between the different CairoUtilFuncGens
- Memory moved from its own pass to being included in a multipass with storage to allow reuse of information and easier processing of copy semantics in the future
- Where possible, memory functions are included in warplib rather than being generated to reduce bloat
- CairoFunctionDefinitionWriter has been tidied up and now only external functions instantiate warp_memory, allowing for memory references to be passed to and between internal calls
- Fixed a bug where annotateImplicits was effectively removing all imports registered to function definitions
- VariableDeclarationInitialiser now uses getDefaultValue rather than its own more lacking set of defaults
- Fixed a bug where only elementary dynamic array types could be default instantiated
- Fixed the typing of the argument to default dyamic array new expressions


example_contracts/mutableReferences/deepDelete.sol
```diff
@@ -0,0 +1,13 @@
...
+
+contract WARP {
+  uint8[3][] arr;
+  function test() public returns (uint8) {
+    uint8[3] storage subArr = arr.push();
+    subArr[1] = 5;
+    delete arr;
+    return subArr[1];
+  }
+}
```
RohitRanjangit, line 13: Missing new Line; 😁

RohitRanjangit (changes requested):

GuyMcComb (approved):
LGTM LFG!

PR #289, opened 2022-04-02
Updated to solc-typed-ast 0.9

DomHenderson:


GuyMcComb:
Will ASTPrinter still work since it uses node.type attribute? `/src/utils/astPrinter.ts` line 45?

GuyMcComb (approved):
LGTM!

PR #291, opened 2022-04-04
external contract method call support

RohitRanjangit:


src/passes/addressHandler/addressArgumentPusher.ts
```diff
@@ -1,37 +0,0 @@
-import {
```
BitBaseBit, line 1: Why was this removed?

DomHenderson: Yeah, doing the transform for cross contract calling in the AST is a huge pain to do in a way that works nicely with other passes and doesn't have to be constantly designed around anytime we process functions. The addressHandler and its constituent passes have been deprecated for a while because of this, they were only kept in to throw a NotSupportedYet when they would have pushed an argument. Now that such calls are supported they're not needed.

PR #296, opened 2022-04-06
cross contract method call support

RohitRanjangit:
This PR will facilitate the support of cross-contact methods calls. For eg. the contracts:
```solidity

pragma solidity ^0.8.10;

//SPDX-License-Identifier: MIT

contract A{
    uint a;
    function f() public returns (uint) {
        a = 34;
        return a;
    }
}

contract WARP{
    function f(address addr) public returns (uint){
        return A(addr).f();
    }
}
```


tests/behaviour/expectations/behaviour.ts
```diff
@@ -132,6 +132,58 @@ export const expectations = flatten(
...
+              Expect.Simple(
+                'f',
+                ['address@tests/behaviour/contracts/cross_contract_calls/simple.A'],
+                ['69', '0'],
+              ),
+            ],
+            'WARP',
+          ),
+          File.Simple('public_vars', [Expect.Simple('f', [], ['696', '0'])], 'A'),
+          // File.Simple(
```
DomHenderson, line 149: Are these tests meant to be commented out? Also is there a reason that multiple tests on the same file are split into separate uses of `File.Simple`?

tests/behaviour/contracts/cross_contract_calls/simple.sol
```diff
@@ -0,0 +1,18 @@
...
+        a = 69;
+        return a;
+    }
+}
+
+contract WARP{
+    function f(address addr) public returns (uint){
+        return A(addr).f();
+    }
+}
```
DomHenderson, line 18: Missing newline

src/passes/externalContractHandler/index.ts
```diff
@@ -0,0 +1,16 @@
+import { AST } from '../../ast/ast';
+import { ASTMapper } from '../../ast/mapper';
+
+import { ContractDefinition } from 'solc-typed-ast';
+
+import { ExternalContractInterfaceInserter } from './externalContractInterfaceInserter';
+
```
DomHenderson, line 7: I'd appreciate a comment here outlining the overall transform and more importantly why it's needed.

DomHenderson (changes requested):
Just some initial comments. Overall this looks really good. I'll take a deep dive into the nitty gritty of it later today once I have some more time, and ideally after these comments have been addressed

tests/behaviour/expectations/behaviour.ts
RohitRanjangit: For the commented tests, a weird error is coming, similar to: ` \' message\': \' Contract address 0x0 is out of range\'})'})` when it executes the line `B.setA_f95a0ad2(__warp_usrid0_a, __warp_usrid2_addrA)` in the Cairo function:
```Cairo
func setB_57be73c5_internal{
      syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}(
      __warp_usrid2_addrA : felt, __warp_usrid3_addrB : felt) -> ():
  alloc_locals
  WS_WRITE0(__warp_usrid0_a, __warp_usrid3_addrB)
  B.setA_f95a0ad2(__warp_usrid0_a, __warp_usrid2_addrA)
  return ()
end
```
I think the value of `__warp_usrid0_a` is still 0 after the write operation and has to be read using `WS0_READ_felt(__warp_usrid0_a)` operation. Not sure though. 

The reason for multiple uses of `File.Simple` is to deploy each contract from the solc file. Now, `File.Simple` will accept a new optional contract name argument.

DomHenderson: Ahh I see now. I think this is the same issue Carmen was running into with using state variables in tests. It's fixed in my branch, but in the meantime a workaround should be to read the storage variable into a temporary before passing it to a function. So `a.setA(addrA)` becomes `B temp = a; temp.setA(addrA);` and similarly for other instances where storage variables end up being used like that

RohitRanjangit: Done, tests are passing

src/transpiler.ts
```diff
@@ -77,6 +77,7 @@ function applyPasses(ast: AST, options: TranspilationOptions): AST {
     ['Ufr', UsingForResolver],
     ['Na', NamedArgsRemover],
     ['Gp', PublicStateVarsGetterGenerator],
+    ['Ech', ExternalContractHandler],
```
DomHenderson, line 80: I think it would probably make most sense to put this after InheritanceInliner, so that any calls to base class functions don't cause interface generation

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -0,0 +1,154 @@
...
+      contractInterface.id,
+      func.kind,
+      func.name,
+      func.virtual,
+      func.visibility,
+      func.stateMutability,
+      func.isConstructor,
+      cloneASTNode(func.vParameters, ast),
+      cloneASTNode(func.vReturnParameters, ast),
+      func.vModifiers,
```
DomHenderson, line 58: Modifiers would also need to be cloned too. I would suggest doing as is done in `ExternalFunctionCreator.createExternalFunctionDefinition` (and ideally then refactoring the two into a `cloneFunctionWithoutBody` in `cloning.ts`) where the method is to disconnect the body, clone the FunctionDefinition, then reconnect the body to the original FunctionDefinition

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -0,0 +1,154 @@
...
+function genContractInterface(
+  contract: ContractDefinition,
+  sourceUnit: SourceUnit,
+  ast: AST,
+): ContractDefinition {
+  const contractId = ast.reserveId();
+  const contractInterface = new CairoContract(
+    contractId,
+    '',
+    `${contract.name}@interface`,
```
DomHenderson, line 31: I'm assuming this is to prevent redaclaration of existing contracts, is there any other reason for modifying the name like this?

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -0,0 +1,154 @@
...
+  assert(sourceUnit !== undefined, 'Trying to import a definition into an unknown source unit');
+
+  if (contractSourceUnit === undefined || sourceUnit === contractSourceUnit) return;
+
+  if (contract.kind === ContractKind.Library) return;
+  if (contractInterfaces.has(contract.id)) return;
+  contractInterfaces.set(contract.id, genContractInterface(contract, sourceUnit, ast));
+}
+
+export class ExternalContractInterfaceInserter extends ASTMapper {
```
DomHenderson, line 85: I think this pass has the potential to import more interfaces than are necessary. Would it be sufficient to specifically check FunctionCalls whose vExpression is a MemberAccess, and to check the type of the child node of that member access to see if it's a contract?

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
RohitRanjangit: Yes, it is just the avoid conflict with the existing contracts

src/passes/builtinHandler/explicitConversionToFunc.ts
```diff
@@ -10,32 +11,48 @@ import {
...
-    const typeNameType = getNodeType(node.vExpression, ast.compilerVersion);
     assert(
       typeNameType instanceof TypeNameType,
       `Got non-typename type ${typeNameType.pp()} when parsing conversion function ${
         node.vFunctionName
       }`,
     );
+
+    if (
+      typeNameType instanceof TypeNameType &&
```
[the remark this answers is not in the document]
RohitRanjangit: Removed!!

src/passes/builtinHandler/explicitConversionToFunc.ts
```diff
@@ -10,32 +11,48 @@ import {
...
+      typeNameType.type instanceof UserDefinedType &&
+      typeNameType.type.definition instanceof ContractDefinition
+    ) {
+      const operand = cloneASTNode(node.vArguments[0], ast);
```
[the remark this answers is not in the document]
RohitRanjangit: Removed!!

src/passes/externImporter.ts
```diff
@@ -30,9 +31,9 @@ export class ExternImporter extends ASTMapper {
     if (declarationSourceUnit === undefined || sourceUnit === declarationSourceUnit) return;
 
     if (
-      declaration instanceof ContractDefinition ||
       declaration instanceof FunctionDefinition ||
-      declaration instanceof EnumDefinition
+      declaration instanceof EnumDefinition ||
+      (declaration instanceof ContractDefinition && declaration.kind === ContractKind.Library)
```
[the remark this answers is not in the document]
RohitRanjangit: Libraries are being treated as a contract Definition in the ASTs hence it goes into separate SourceUnit, but they are only referenced using their name unlike for contracts where we have an address associated with it.

src/passes/externalContractHandler/index.ts
```diff
@@ -0,0 +1,22 @@
...
+/*
+  This is a compound pass which internally calls ExternalContractInterfaceInserter pass 
+  to insert interfaces for the external contracts that has been refrenced into the AST
+  for every SourceUnit.
+*/
```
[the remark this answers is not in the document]
RohitRanjangit: Added some more lines!

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -0,0 +1,154 @@
...
+  UserDefinedTypeName,
+  VariableDeclaration,
+} from 'solc-typed-ast';
+
+import { AST } from '../../ast/ast';
+import { CairoContract } from '../../ast/cairoNodes';
+import { ASTMapper } from '../../ast/mapper';
+import { cloneASTNode } from '../../utils/cloning';
+
+function genContractInterface(
```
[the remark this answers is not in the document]
RohitRanjangit: Done!!

src/transpiler.ts
RohitRanjangit: Position Changed!!

src/passes/index.ts
```diff
@@ -33,3 +32,4 @@ export * from './unreachableStatementPruner';
 export * from './usingForResolver';
 export * from './variableDeclarationExpressionSplitter';
 export * from './variableDeclarationInitialiser';
+export * from './externalContractHandler';
```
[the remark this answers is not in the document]
RohitRanjangit: Done!

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
RohitRanjangit: Done!

src/passes/externImporter.ts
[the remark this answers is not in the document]
DomHenderson: True, however we handle libraries by converting them into base contracts and doing internal calls (referencedLibraries and InheritanceInliner passes do this), so something has gone wrong if they're getting imported here

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
RohitRanjangit: It seems that it is sufficient and all tests are passing, I covered the extra cases to just be safe.

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
RohitRanjangit: It is clashing

src/passes/externImporter.ts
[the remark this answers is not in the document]
RohitRanjangit: Actually, they are not getting imported here. I think it is safe to remove this line from here. Is it?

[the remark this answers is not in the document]
DomHenderson: Yeah, I think it's a good idea to remove it

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
DomHenderson: As discussed, fixing the clashing without mangling the names like this would be more hassle than it's worth

DomHenderson (approved):

PR #298, opened 2022-04-07
External Memory Structs

GuyMcComb:
Adds a pass that writes memory struct arguments in external functions to the current warp memory implementation.


src/passes/externalArgumentModifier.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  generateStructIdentifier(varDecl: VariableDeclaration, ast: AST): Expression {
+    assert(
+      varDecl.vType instanceof UserDefinedTypeName &&
+        varDecl.vType.vReferencedDeclaration instanceof StructDefinition,
+    );
+    const funcIdentifier = new Identifier(
+      ast.reserveId(),
+      '',
+      `type(${varDecl.typeString} storage pointer)`,
```
GuyMcComb, line 142: Do we have a function that can do this?

src/passes/externalArgumentModifier.ts
```diff
@@ -0,0 +1,186 @@
...
+  ): VariableDeclarationStatement {
+    const memoryStruct = cloneASTNode(calldataStruct, ast);
+    memoryStruct.name = calldataStruct.name + '_mem';
+    memoryStruct.storageLocation = DataLocation.Memory;
+    this.oldToNewStructMapping.set(calldataStruct, memoryStruct);
+    const callDataVarDecl = createIdentifier(memoryStruct, ast);
+    const structDeclarationStatement = new VariableDeclarationStatement(
+      ast.reserveId(),
+      '',
+      [1],
```
DomHenderson, line 96: I think this should be `memoryStruct.id`

src/passes/externalArgumentModifier.ts
DomHenderson: CreateIdentifier is the usual goto for this, but it doesn't take into account the typestring differences for reference types or the need to wrap it in `type()`, so there's nothing currently that would produce this exact result

src/passes/index.ts
```diff
@@ -32,3 +32,8 @@ export * from './unreachableStatementPruner';
 export * from './usingForResolver';
 export * from './variableDeclarationExpressionSplitter';
 export * from './variableDeclarationInitialiser';
+export * from './publicFunctionSplitter';
+export * from './externalInputChecker';
+export * from './generateGetters';
+export * from './namedArgsRemover';
+export * from './externalArgumentModifier';
```
DomHenderson, line 39: All of these except externalArgumentModifier are already exported here

DomHenderson (changes requested):

src/passes/externalArgumentModifier.ts
```diff
@@ -0,0 +1,186 @@
...
+    return funcIdentifier;
+  }
+
+  private generateMemberAccess(varDecl: VariableDeclaration, ast: AST): Array<MemberAccess> {
+    assert(
+      varDecl.vType instanceof UserDefinedTypeName &&
+        varDecl.vType.vReferencedDeclaration instanceof StructDefinition,
+    );
+    const memberAccessArray = varDecl.vType.vReferencedDeclaration.vMembers.map((member) => {
+      const newIdentifier = new Identifier(
```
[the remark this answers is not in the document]
GuyMcComb: replaced.

src/passes/externalArgumentModifier.ts
GuyMcComb: Replaced.

src/passes/externalArgumentModifier.ts
```diff
@@ -0,0 +1,186 @@
...
+      );
+    });
+    return memberAccessArray;
+  }
+
+  visitIdentifier(node: Identifier, ast: AST): void {
+    if (
+      node.vReferencedDeclaration instanceof VariableDeclaration &&
+      this.oldToNewStructMapping.get(node.vReferencedDeclaration) !== undefined &&
+      !this.identifiersNotToBeChanged.includes(node)
```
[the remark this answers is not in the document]
GuyMcComb: Implemented.

src/passes/externalArgumentModifier.ts
```diff
@@ -0,0 +1,186 @@
...
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+import { cloneASTNode } from '../utils/cloning';
+import { createIdentifier } from '../utils/nodeTemplates';
+
+export class ExternalArgumentModifier extends ASTMapper {
+  /*
+
+  Looks for all memory struct VariableDeclarations in FunctionDefinitions. e.g FunctionExample(structDef memory structVar).
+  For this memory struct to be implemented in our own memory system it needs to be registered and allocated.
```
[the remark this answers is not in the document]
GuyMcComb: Done

DomHenderson (approved):

PR #300, opened 2022-04-08
Validate constructor input

tserg:
Validate constructor arguments by adding public functions to the external input checker passes.

This PR also improves the behaviour test for `simpleImmutable`, which spawned this fix and is now found to be associated with an unrelated discrepancy between Starknet CLI and the behaviour/Starknet test suite when providing negative integers as arguments to the deployment of a contract.

### DIFF
--- a/tests/behaviour/expectations/behaviour.ts
+++ b/tests/behaviour/expectations/behaviour.ts
@@ -82,10 +82,12 @@ export const expectations = flatten(
           new File(
             'simpleImmutable',
             'WARP',
-            ['0', '256'],
+            ['0', '256', '512'],
             [
               Expect.Simple('getUintValue', [], ['0', '256'], '0'),
+              Expect.Simple('getIntValue', [], ['512'], '0'),
               Expect.Simple('addUintValue', ['0', '256'], ['0', '512'], '0'),
+              Expect.Simple('addIntValue', ['256'], ['768'], '0'),
             ],
           ),
         ]),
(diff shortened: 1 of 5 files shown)


DomHenderson (changes requested):
Looks good. Just one thing before merging, if you could extend the input validation tests to also check that inputting invalid data into the constructor throws as expected that would be appreciated

tests/behaviour/expectations/behaviour.ts
```diff
@@ -88,6 +88,15 @@ export const expectations = flatten(
               Expect.Simple('getIntValue', [], ['512'], '0'),
               Expect.Simple('addUintValue', ['0', '256'], ['0', '512'], '0'),
               Expect.Simple('addIntValue', ['256'], ['768'], '0'),
+              new Expect('testing constructor argument out of bounds', [
```
DomHenderson, line 91: If we're using Expects specifically for failing constructor tests, I think it's worth putting something into the `Expect` constructor in `types.ts` that checks whether any of the function names in `steps` is `constructor` and if so, it should assert that the expected return value for that `Expect` is null

DomHenderson (changes requested):
This is a clever way of handling this, and probably ends up neater than what I was expecting we'd do (`File`s with `null` as their expectations), just one requested change and then it'll be good to go

DomHenderson (approved):

PR #301, opened 2022-04-08
Modifier inheritance

cicr99:
Added functionality to handle modifier inheritance within `inheritanceInliner` pass. It follows the same approach taken by function inheritance.


DomHenderson (approved):
Looks great, an elegant solution and well tested. Merge away!

PR #303, opened 2022-04-11
Implement documentation by preserving comments

Glitch18:
Preserves comments of the form `/** */` and `///` by utilizing the `StructuredDocumentation` node that gets added as a child to the node described in the documentation. 

To-Do;

- [x] Implement `StructuredDocumentationWriter`
- [x] Add support in `cairoWriter.ts` 
- [x] Add support in `cloning.ts`
- [x] Check [NatSpec](https://docs.soliditylang.org/en/v0.8.13/natspec-format.html)
- [x] Investigate the following example

Example:

```
/// Contract Declaration test
contract HelloWorld {
  /** Function 
  Declaration test
  */
  function getConstant() public pure returns (uint256) {
    return 5;
  }

  /** Variable Declaration test */
  uint256 x = 5;
}
```

In the above contract, the documentation for `uint256 x = 5` does not get preserved.

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -86,6 +86,17 @@ import { isCairoConstant } from './utils/utils';
 
 const INDENT = ' '.repeat(4);
 
+function getDocumentation(
+  documentation: string | StructuredDocumentation | undefined,
+  writer: ASTWriter,
+): string {
+  return documentation !== undefined
+    ? typeof documentation === 'string'
+      ? `# ${documentation.split('\n').join('\n#')}`
+      : writer.write(documentation)
+    : '';
+}
+
 export abstract class CairoASTNodeWriter extends ASTNodeWriter {
   ast: AST;
   throwOnUnimplemented: boolean;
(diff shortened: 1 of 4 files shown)


src/cairoWriter.ts
```diff
@@ -86,6 +86,17 @@ import { isCairoConstant } from './utils/utils';
 
 const INDENT = ' '.repeat(4);
 
+function getDocumentation(
+  documentation: string | StructuredDocumentation | undefined,
+  writer: ASTWriter,
+): string {
+  return typeof documentation !== 'undefined'
```
DomHenderson, line 93: This initial condition can be simplified to `documentation !== undefined`, no need for the `typeof`

DomHenderson (changes requested):
Looks good, just two small changes requested and then it'll be good to merge

src/cairoWriter.ts
Glitch18: Done

DomHenderson (approved):

PR #305, opened 2022-04-13
Support fixed-size byte arrays

tserg:
Convert fixed-size byte arrays to unsigned integers of the appropriate width(e.g. `bytes2` -> `uint16`)

**What I did**
- Update `typeString` for nodes of fixed-size byte arrays type.
- For literals, replace the value in hexadecimal representation with its integer value.

Notes:
- Based on the order of passes, `getFunctionTypeString()` will be used by this pass before enums are converted to their integer equivalents. Therefore, an additional check is inserted to exclude `FunctionDefinition` nodes with enum typestrings from throwing an error. This should be sound since enums are now supported (?).
- From Solidity docs: Prior to version 0.8.0, `byte` used to be an alias for `bytes1`.


src/passes/bytesConverter.ts
```diff
@@ -0,0 +1,152 @@
...
+  UnaryOperation,
+  typeNameToTypeNode,
+  VariableDeclaration,
+} from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+import { getFunctionTypeString } from '../utils/utils';
+
+/* Convert fixed-size byte arrays (e.g. bytes2, bytes8) to their equivalent unsigned integer.
+		This pass currently does not handle dynamically-sized bytes arrays (i.e. bytes).
```
DomHenderson, line 26: These tabs should be spaces

src/passes/bytesConverter.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  replacementIntTypeNode(typeNode: FixedBytesType): IntType {
+    return new IntType(typeNode.size * 8, false, undefined);
+  }
+
+  visitAssignment(node: Assignment, ast: AST): void {
+    const typeNode = getNodeType(node, ast.compilerVersion);
+    if (typeNode instanceof FixedBytesType) {
+      const replacementTypeNode = this.replacementIntTypeNode(typeNode);
```
DomHenderson, line 44: This is a good approach for scalar fixed-bytes-types, but what about arrays of fixed bytes types?

src/passes/bytesConverter.ts
```diff
@@ -0,0 +1,152 @@
...
+  replacementLiteralIntValue(node: Literal): string {
+    const literalInt = BigInt(node.value);
+    return literalInt.toString();
+  }
+
+  replacementIntTypeNode(typeNode: FixedBytesType): IntType {
+    return new IntType(typeNode.size * 8, false, undefined);
+  }
+
+  visitAssignment(node: Assignment, ast: AST): void {
```
DomHenderson, line 41: You may be able to simpliify the pass by combining many of these methods into a `visitExpression` function

src/passes/bytesConverter.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (typeNode instanceof FixedBytesType) {
+      const replacementTypeNode = this.replacementIntTypeNode(typeNode);
+      node.typeString = replacementTypeNode.pp();
+    }
+    this.commonVisit(node, ast);
+  }
+
+  visitLiteral(node: Literal, _: AST): void {
+    if (node.typeString.startsWith('int_const ') && node.value.startsWith('0x')) {
+      node.value = this.replacementLiteralIntValue(node);
```
DomHenderson, line 100: Why is this necessary? Literals of the form `0x1234` should work fine in cairo

src/passes/bytesConverter.ts
```diff
@@ -0,0 +1,152 @@
...
+  visitMemberAccess(node: MemberAccess, ast: AST): void {
+    const typeNode = getNodeType(node, ast.compilerVersion);
+    if (typeNode instanceof FixedBytesType) {
+      const replacementTypeNode = this.replacementIntTypeNode(typeNode);
+      node.typeString = replacementTypeNode.pp();
+    } else if (
+      node.vExpression instanceof Identifier &&
+      node.vExpression.vReferencedDeclaration instanceof VariableDeclaration &&
+      node.vExpression.vReferencedDeclaration.vType instanceof ArrayTypeName &&
+      node.memberName === 'push'
```
DomHenderson, line 113: Why the push function in particular? Fixed-size bytes types don't have a push function, and dynamically sized bytes types will have to handled completely differently

src/utils/utils.ts
```diff
@@ -316,7 +317,10 @@ export function getFunctionTypeString(node: FunctionDefinition, compilerVersion:
   const inputs = node.vParameters.vParameters
     .map((decl) => {
       const baseType = getNodeType(decl, compilerVersion);
-      if (baseType instanceof ArrayType || baseType instanceof UserDefinedType) {
+      if (
+        baseType instanceof ArrayType ||
+        (baseType instanceof UserDefinedType && !(baseType.definition instanceof EnumDefinition))
```
DomHenderson, line 322: This should be  a positive check for StructDefinition not a negative check for EnumDefinition, since Contract types can have default storage location

tests/behaviour/contracts/bytes/byteArrays.sol
```diff
@@ -0,0 +1,35 @@
...
+pragma solidity ^0.7.0;
+
+contract WARP {
+
+  bytes2[3] ba;
+	bytes2[] d;
+	uint constant bbb = 2;
+
+	constructor() {
+		ba[0] = 0x1234;
```
DomHenderson, line 11: There are many tabs in this file that should be spaces

tests/behaviour/expectations/behaviour.ts
```diff
@@ -15,6 +15,76 @@ export const expectations = flatten(
             Expect.Simple('test256', ['3', '4'], ['3', '4']),
           ]),
         ]),
+        new Dir('bytes', [
```
DomHenderson, line 18: Love to see good thorough testing like this

DomHenderson (changes requested):

src/passes/bytesConverter.ts
tserg: This was meant to handle pushes to an array of fixed-size bytes types. Although transpilation still succeeds if this typeString is not changed, I thought it would be cleaner to do so. What do you think?

```
      | +-FunctionCall #93
      |   |   typeString: tuple()
      |   +-MemberAccess #91
      |   | |   memberName: push
      |   | |   typeString: function (bytes2)
      |   | +-Identifier #89
      |   |       name: d
      |   |       referencedDeclaration: 8
      |   |       typeString: bytes2[] storage ref
      |   +-Literal #92
      |         typeString: int_const 4660
      |         value: 0x1234
```

tserg:
Updates:
- Fixed a bug with typestring in ConstantHandler
- Moved BytesConverter after Storage pass, and consequently ExternalInputChecker (needs to check converted bytes) and EnumConverter (needs to be after ExternalInputChecker) after BytesConverter. Consequential changes that follow:
  - Exclude interfaces from ExternalInputChecker.
  - Handle arrays, mappings and structs in VariableDeclarationExpressionSplitter.
  - add FixedBytesType to `cairoTypeSystem.ts`.

src/passes/variableDeclarationExpressionSplitter.ts
```diff
@@ -30,6 +43,90 @@ export class VariableDeclarationExpressionSplitter extends ASTMapper {
   generateNewConstantName(): string {
     return `__warp_td_${this.lastUsedConstantId++}`;
   }
+
+  // Helper function to generate the typeName for a given typeNode, to be used for
+  // the new VariableDeclarationStatement in this.splitDeclaration
+  generateTypeNameNode(
```
DomHenderson, line 49: We already have typeNameFromTypeNode in utils, would be worth checking that they produce the same result, adding any cases this covers that typeNameFromTypeNode doesn't, and then using that

src/testing.ts
```diff
@@ -103,7 +103,7 @@ const expectedResults = new Map<string, ResultType>([
   ['example_contracts/state_variables/arrays', 'CairoCompileFailed'],
   ['example_contracts/state_variables/mappings', 'Success'],
   ['example_contracts/state_variables/structs', 'CairoCompileFailed'],
-  ['example_contracts/state_variables/misc', 'NotSupportedYet'],
+  ['example_contracts/state_variables/misc', 'TranspilationFailed'],
```
DomHenderson, line 106: If it's possible to turn this into a NotSupportedYet that would be appreciated, I prefer not to have any TranspilationFailed merged in. It should also have a comment as to why it fails

DomHenderson (changes requested):

src/passes/externalInputChecker/booleanBoundChecker.ts
```diff
@@ -16,6 +18,13 @@ import { typeNameFromTypeNode } from '../../utils/utils';
 import { createIdentifier } from '../../utils/nodeTemplates';
 
 export class BooleanBoundChecker extends ASTMapper {
+  visitContractDefinition(node: ContractDefinition, ast: AST): void {
+    if (node.kind === ContractKind.Interface) {
```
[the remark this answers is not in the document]
tserg: I think it is an insufficient check due to the reordering of the ExternalInputChecker pass. The failing test is `tests/behaviour/contracts/cross_contract_calls/public_vars.sol` specifically.

src/passes/bytesConverter.ts
```diff
@@ -38,29 +37,51 @@ export class BytesConverter extends ASTMapper {
...
+      typeNode.to instanceof ArrayType &&
+      typeNode.to.elementT instanceof FixedBytesType
+    ) {
+      const replacementIntTypeNode = this.replacementIntTypeNode(typeNode.to.elementT);
+      typeNode.to.elementT = replacementIntTypeNode;
+      if (typeNode.location === DataLocation.Default)
+        node.typeString = `${typeNode.to.pp()} storage ref`;
+      else if (typeNode.location === DataLocation.Memory)
+        node.typeString = `${typeNode.to.pp()} memory`;
+      else node.typeString = `${typeNode.to.pp()} storage pointer`;
```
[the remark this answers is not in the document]
tserg: Yes. I will refactor this to generate a new TypeNode and rely on the `pp()` function instead as this feels quite verbose and brittle.

DomHenderson (approved):

PR #306, opened 2022-04-13
Adding tests

GuyMcComb:
Adding tests for static array tests.

Have a look at the test function names. What do you think of this convention. I found the names for tests were getting longer and longer as they got more complex. The description in `expectations/behaviour.ts` gives the details about the test rather than the name of the function?

### DIFF
--- a/tests/behaviour/contracts/external_function_inputs/static_array_return_index.sol
+++ b/tests/behaviour/contracts/external_function_inputs/static_array_return_index.sol
@@ -4,10 +4,29 @@ pragma solidity ^0.8.0;
 
 contract WARP {
 
+struct StrucDef {
+    uint8 x;
+    uint256 y;
+}
 
-    function testReturnIndex(int16[3] memory listA) pure external returns (int16) {
-        //int16[2] memory listB = [listA[1], listA[2]];
+    function testA(int16[3] memory listA) pure external returns (int16) {
+         return listA[2];
+    }
+
+    function testB(int16[3] memory listA) pure public returns (int16) {
         return listA[2];
     }
+    // Would return y but warp at this stage does not return whole structs.
+    function testC(StrucDef[3] memory listA) pure external returns (uint8) {
+        return listA[2].x;
+    }
+
+    function testD(StrucDef[3] memory listA) pure public returns (uint8) {
+         return listA[2].x;
+    }
+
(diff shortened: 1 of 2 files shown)


tests/behaviour/contracts/external_function_inputs/static_array_return_index.sol
```diff
@@ -4,10 +4,29 @@ pragma solidity ^0.8.0;
...
-    function testReturnIndex(int16[3] memory listA) pure external returns (int16) {
-        //int16[2] memory listB = [listA[1], listA[2]];
+    function testA(int16[3] memory listA) pure external returns (int16) {
+         return listA[2];
+    }
+
+    function testB(int16[3] memory listA) pure public returns (int16) {
         return listA[2];
     }
+    // Would return y but warp at this stage does not return whole structs.
```
DomHenderson, line 19: Would it make sense to change the function to `returns (uint8, uint256)` and the return statement to `return (listA[2].x, listA[2].y)`?

tests/behaviour/expectations/behaviour.ts
```diff
@@ -470,6 +470,35 @@ export const expectations = flatten(
               ],
             ),
           ]),
+          File.Simple('static_array_return_index', [
+            new Expect(
+              'testing a static array of ints can be passed into an external function and written to memory and index returned.',
```
DomHenderson, line 475: These descriptive test names are great, but I'm not a fan of the `testA, testB` naming convention, makes it harder to see what's being tested when looking at the contract and makes it easier to accidentally call the wrong function from a test

DomHenderson (changes requested):

tests/behaviour/expectations/behaviour.ts
GuyMcComb: OK sure thing, will change them back.

tests/behaviour/contracts/external_function_inputs/static_array_return_index.sol
GuyMcComb: uint256 is a struct. At this point we cannot return a whole struct, but I could change it to `return (listA[2].x, listA[2].y.low, listA[2].y.high)`

GuyMcComb: Resolved

DomHenderson (approved):

PR #307, opened 2022-04-14
support specific imports

Purva-Chaudhari:
Pass to handle specific imports eg: `import {A, B } from './file.sol';`
Import types tested for library, free Function, enum and struct. 
To be tested on solc version 0.8.13+


src/testing.ts
```diff
@@ -59,6 +59,10 @@ const expectedResults = new Map<string, ResultType>([
   ['example_contracts/idManglingTest8', 'Success'],
   ['example_contracts/idManglingTest9', 'Success'],
   ['example_contracts/if-flattening', 'Success'],
+  ['example_contracts/imports/importEnum', 'CairoCompileFailed'],
```
DomHenderson, line 62: Could you add comments explaining why the tests that don't succeed don't succeed

src/utils/utils.ts
```diff
@@ -320,6 +320,18 @@ export function typeNameFromTypeNode(node: TypeNode, ast: AST): TypeName {
   return result;
 }
 
+export function getContractTypeString(node: ContractDefinition): string {
```
DomHenderson, line 323: I think with the number of typestring generating functions we're creating, it would be best to have a separate file in the utils folder for all of them

example_contracts/imports/importEnum.sol
```diff
@@ -0,0 +1,13 @@
...
+
+// SPDX-License-Identifier: MIT
+
+//Specific Imports for enum
+ import { selectEnum } from './importfrom.sol';
+
+contract WARP{
+    function enumImport(uint8 num1) pure public {
+      selectEnum enum1;
+  }
```
DomHenderson, line 11: All of these new example contracts need formatting

DomHenderson (changes requested):
This looks good, and is a troublesome issue solved. Just a few tidying requests and it'll be done

DomHenderson (approved):

PR #308, opened 2022-04-14
UsingForDirective Support

RohitRanjangit:
This PR will resolve the function call for the given type that has been binded via `UsingForDirective`
  by visiting every MemberAccess Node inside the function call.

 For example, if we have:
```solidity
using {x} for uint; // x is a file-level function
then a.x() will resolve to x(a)
```

This PR also fixes the bug where in case of struct member access IndentifierMangler doesn't mangle struct Members.


src/testing.ts
```diff
@@ -96,6 +96,12 @@ const expectedResults = new Map<string, ResultType>([
   ['example_contracts/typeConversion/shifts', 'Success'],
   ['example_contracts/typeMinMax', 'Success'],
   ['example_contracts/units', 'Success'],
+  ['example_contracts/using_for/imports/global_directive', 'NotSupportedYet'],
+  ['example_contracts/using_for/imports/user_defined', 'NotSupportedYet'],
+  ['example_contracts/using_for/function', 'NotSupportedYet'],
+  ['example_contracts/using_for/private', 'TranspilationFailed'],
```
DomHenderson, line 102: We shouldn't be expecting tests to produce a TranspilationFailed, as that is the error for when something has gone wrong unexpectedly. I'd also appreciate it if you could add comments above entries marked NotSupportedYet or CairoCompileFailed to indicate what feature is blocking them

tests/behaviour/expectations/behaviour.ts
```diff
@@ -187,6 +187,17 @@ export const expectations = flatten(
             'C',
           ),
         ]),
+        new Dir('using_for', [
+          File.Simple('simple', [
+            Expect.Simple('bar', [], ['6', '0']),
+            Expect.Simple('fum', [], ['60', '0']),
```
DomHenderson, line 193: These tests could do with some more descriptive test names. For example `fum` could be named something lke `callOnConversion` and `bar` could be `callOnIdentifier`, something along those lines.

src/passes/identifierMangler.ts
```diff
@@ -75,7 +75,9 @@ export class IdentifierMangler extends ASTMapper {
       node.memberName = declaration.name;
     }
   }
-
+  visitStructDefinition(_node: StructDefinition, _ast: AST): void {
+    return;
```
DomHenderson, line 79: This could probably do with a comment explaining the reasoning behind the explicit lack of recursion here

tests/behaviour/contracts/using_for/library.sol
```diff
@@ -0,0 +1,46 @@
...
+    function foox() public view returns (uint256) {
+        return a.f() + a.g();
+    }
+    function bary() public view returns (uint256) {
+        return a.foo() * a.bar();
+    }
+    function baz() public view returns (uint256) {
+        return a.baz();
+    }
+}
```
DomHenderson, line 46: Missing newline

DomHenderson (changes requested):
I'll give the implementations details a thorough review once the testing.ts comment in particular has been addressed, but for now here's a few suggestions

src/passes/identifierMangler.ts
RohitRanjangit: Comment Added

src/testing.ts
RohitRanjangit: During transpilation of examples_contract/using_for/private.sol, InheritanceInliner throws an error with msg `AssertionError [ERR_ASSERTION]: Attempted to copy non public/external function`.

tests/behaviour/contracts/using_for/library.sol
RohitRanjangit: Done!

DomHenderson (approved):

PR #309, opened 2022-04-14
Eliminate user defined types

Glitch18:
Implments a new pass to eliminate User Defined types, including their declarations, arrays and member function calls.

To-Do:

 - [x] Verify contract test once array bug is fixed
 - [x] Handle multidimentional arrays
 - [x] Implement Mapping

Note: 

The transpilation runs just fine and the user defined types are transformed as expected when run with all the passes:-

`bin/warp transpile --print-trees example_contracts/userdefinedtypes.sol`

But runs into an error when run only upto the `UserDefinedTypesConverter` pass:-

`bin/warp transpile --print-trees --until Udt example_contracts/userdefinedtypes.sol`

The above is an observation exclusive to the `userdefinedtypes.sol` test.


src/passes/userDefinedTypesConverter.ts
```diff
@@ -0,0 +1,110 @@
...
+        userDefinition.underlyingType.typeString,
+        userDefinition.underlyingType.typeString,
+      ),
+    );
+  }
+
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    this.commonVisit(node, ast);
+    if (!(node.vExpression instanceof MemberAccess)) return;
+    if (!['unwrap', 'wrap'].includes(node.vExpression.memberName)) return;
```
DomHenderson, line 72: There should be an additional check here to ensure that these are the wrap and unwrap functions of a UserDefinedValueType, as it is this could catch calls to contract member functions called wrap and unwrap

src/passes/userDefinedTypesConverter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const argument = node.vArguments[0];
+    if (argument instanceof Identifier) {
+      const declaration: UserDefinedType = getNodeType(
+        argument,
+        ast.compilerVersion,
+      ) as unknown as UserDefinedType;
```
DomHenderson, line 79: `as` should be avoided unless absolutely necessary. Instead assert the type (making sure you're using the assert function from 'assert', not from 'console')

src/passes/userDefinedTypesConverter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      const userDefinition: UserDefinedValueTypeDefinition = <UserDefinedValueTypeDefinition>(
```
DomHenderson, line 81: Same as `as`, conversions like this are unsafe and should be avoided

src/passes/userDefinedTypesConverter.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  visitMapping(node: Mapping, ast: AST): void {
+    this.commonVisit(node, ast);
+    node.typeString = `mapping(${node.vKeyType.typeString} => ${node.vValueType.typeString})`;
+  }
+
+  visitUserDefinedTypeName(node: UserDefinedTypeName, ast: AST): void {
+    const otherTypes: string[] = ['contract', 'enum', 'struct'];
+    if (otherTypes.some((type: string) => node.typeString.startsWith(type))) return;
```
DomHenderson, line 48: It's generally better to check whether the result of getNodeType is a UserDefinedType whose definition is a UserDefinedValueTypeDefinition or not rather than analysing the type strings like this

example_contracts/userdefinedtypes.sol
```diff
@@ -0,0 +1,34 @@
...
+  /// arithmetic on uint256.
+  function mul(UFixed256x18 a, uint256 b) internal pure returns (UFixed256x18) {
+      return UFixed256x18.wrap(UFixed256x18.unwrap(a) * b);
+  }
+  // /// Take the floor of a UFixed256x18 number.
+  // /// @return the largest integer that does not exceed `a`.
+  function floor(UFixed256x18 a) internal pure returns (uint256) {
+      return UFixed256x18.unwrap(a) / multiplier;
+  }
+  // /// Turns a uint256 into a UFixed256x18 of the same value.
```
DomHenderson, line 29: Extraneous slashes

DomHenderson (changes requested):
Going to need some behaviour tests before this can be merged. Also I think it needs to modify the typestrings of identifiers (watch out for storage locations) that reference user defined value type variables, otherwise type analysis will not treat them as the underlying type

src/passes/userDefinedTypesConverter.ts
Glitch18: Done

src/passes/userDefinedTypesConverter.ts
Glitch18: Done

src/passes/userDefinedTypesConverter.ts
Glitch18: Done

src/passes/userDefinedTypesConverter.ts
Glitch18: Done

example_contracts/userdefinedtypes.sol
Glitch18: Done

DomHenderson (approved):

PR #310, opened 2022-04-15
Constructor inheritance

cicr99:
`inheritanceInliner` pass was restructured so each functionality was solved in different files for better readability. The code can be found inside `passes/inheritanceInliner` folder.

Added methods to solve constructor inheritance. When the constructor of a contract is invoked, it first executes the constructors of its base contracts in the lineraized order given by the ast. Therefore, the main idea here is to replace the constructor of the contract with a function that handles the call to each contract constructor in the correct order.


src/passes/inheritanceInliner/constructorInheritance.ts
```diff
@@ -0,0 +1,291 @@
...
+  VariableDeclaration,
+  VariableDeclarationStatement,
+} from 'solc-typed-ast';
+import { AST } from '../../ast/ast';
+import { cloneASTNode } from '../../utils/cloning';
+import { generateFunctionCall } from '../../utils/functionGeneration';
+import { createBlock, createIdentifier, createParameterList } from '../../utils/nodeTemplates';
+import { updateReferencedDeclarations } from './utils';
+
+export function solveConstructorInheritance(
```
DomHenderson, line 21: There are some quite non-trivial transforms going on in this file. There should be a top level comment summarising what's going on. It should explain how there are two different kinds of delegated constructors (modifier-like vs specifier) and at a high level how each is handled. Shorthand code examples of before and after could even be helpful

DomHenderson (changes requested):

DomHenderson (approved):

PR #311, opened 2022-04-16
Added the test case for width and sign conversion simultaneously for solidity 0.7.x version.

phyBrackets:
cc - @DomHenderson , Hey Dom 
Here is the draft PR, can you please spot the issue that I were talking about ?


DomHenderson (changes requested):
The issue is that the contract is named A whereas File.Simple expects the contract to be named WARP (the contract name gets included in the output file name, hence why it couldn't find the file). The tests pass once it's changed to WARP. If you could add a few more tests to cover all cases (there should be at least 4 for combinations of widening/narrowing and signed->unsigned or vice versa) and fix the formatting, then it'll be good to go

phyBrackets:
Going to close this PR , it's here https://github.com/NethermindEth/warp/pull/314

PR #312, opened 2022-04-18
Minor optimization to publicFunctionSplitter

rodrigo-pino:
Public functions not called from inside the contract are transpiled as external only


src/passes/publicFunctionSplitter/internalFunctionCallCollector.ts
```diff
@@ -0,0 +1,30 @@
...
+     This class collects all internal function calls in the contract
+  */
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    const functionDefintion = node.vReferencedDeclaration;
+    if (
+      node.kind === FunctionCallKind.FunctionCall &&
+      functionDefintion instanceof FunctionDefinition &&
+      (node.vExpression instanceof MemberAccess || node.vExpression instanceof Identifier)
+    ) {
+      this.internalFunctionCallSet.add(functionDefintion);
```
DomHenderson, line 26: Since this pass is run on all the roots before with a single `internalFunctionCallSet` will it not also collect external calls?

src/passes/publicFunctionSplitter/internalFunctionCallCollector.ts
```diff
@@ -0,0 +1,30 @@
...
+import { AST } from '../../ast/ast';
+import { ASTMapper } from '../../ast/mapper';
+
+export class InternalFunctionCallCollector extends ASTMapper {
+  constructor(public internalFunctionCallSet: Set<FunctionDefinition>) {
+    super();
+  }
+
+  /*
+     This class collects all internal function calls in the contract
```
DomHenderson, line 17: This should explain why the class is needed. For example in this case something along the lines of how if there are not internal calls to a public function then it doesn't need to be split, which reduces the step count and bloat of the resulting cairo code. It should also go above the class rather than in the middle of it

DomHenderson (changes requested):
This looks good, just a little nitpicking to deal with and then it'll be ready to go

src/passes/publicFunctionSplitter/internalFunctionCallCollector.ts
rodrigo-pino: Yes, it also collects external calls but since in the moment the pass is done I have no information of the Functions defined in the contract. It would be necessary to add

DomHenderson (approved):

PR #314, opened 2022-04-20
Added the test cases and expected behaviour for the width and sign co…

phyBrackets:
Added the test cases for the width-sign conversion.


tests/behaviour/contracts/conversions/widthsignWideningItoUchange.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.7.6;
+
+//SPDX-License-Identifier: MIT
+
+contract WARP {
+    function widthsignWitou(int16 x) pure public returns (uint32) {
+    return (uint32(x));
+  }
+}
```
DomHenderson, line 9: Could you add newlines at the end of all the files that have this symbol at the end

tests/behaviour/expectations/behaviour.ts
```diff
@@ -222,67 +222,21 @@ export const expectations = flatten(
             Expect.Simple('explicit', ['20'], ['20', '20', '0']),
             Expect.Simple('explicit', ['240'], ['240', '240', '0']),
           ]),
-        ]),
-        new Dir('cross_contract_calls', [
```
DomHenderson, line 226: A lot of important tests are getting deleted here. Could you please reverse these deletions

DomHenderson (changes requested):

DomHenderson (approved):

PR #317, opened 2022-04-22
Extern Dynamic Array Handling

GuyMcComb:
This PR splits a single dArray argument in an external function into the number of members and a pointer pointing to the array of members.


src/utils/functionGeneration.ts
```diff
@@ -48,6 +48,7 @@ export function createCairoFunctionStub(
   implicits: Implicits[],
   ast: AST,
   nodeInSourceUnit: ASTNode,
+  splitDarray?: boolean,
```
GuyMcComb, line 51: Not sure how I feel about this being an argument name, I didn't want to add anything like "externalInterface" since the functions should always be internal.

src/cairoWriter.ts
```diff
@@ -592,11 +625,34 @@ class IndexAccessWriter extends CairoASTNodeWriter {
     assert(node.vIndexExpression !== undefined);
     const baseWritten = writer.write(node.vBaseExpression);
     const indexWritten = writer.write(node.vIndexExpression);
+    if (
```
GuyMcComb, line 628: Pretty awful conditional here. Let me know what you think.

src/cairoWriter.ts
```diff
@@ -380,7 +391,23 @@ class ParameterListWriter extends CairoASTNodeWriter {
           : TypeConversionContext.Ref
         : TypeConversionContext.Declaration;
 
-    const params = node.vParameters.map((value, i) => {
+    const proccessed_params = node.vParameters.flatMap((decl) => {
+      // This conditional is placed here to split DynamicArrays into their corresponding length and pointer when they
+      // are an argument for an external function.
+      if (
+        decl.vType instanceof ArrayTypeName &&
```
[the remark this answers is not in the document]
DomHenderson: A function to check if a varDecl is a dynamic array, and then one that holds the full ParameterListWriter condition and one for the IndexAccessWriter condition seems entirely reasonable to me

DomHenderson (approved):

PR #320, opened 2022-04-25
Scope checker

DomHenderson:
Adds an extra pass in the AST checker that verifies that all function definition parameters and return parameters have their scope correctly set


cicr99 (approved):
Looks good to me

cicr99 (approved):

PR #321, opened 2022-04-27
Add no-stubs option

DomHenderson:
Adds the `--no-stubs` options to `transpile` and `transform` to ignore CairoFunctionDefinitions for stubbed functions when printing trees or the solidity output from `transform`


src/index.ts
```diff
@@ -101,12 +104,12 @@ program
 
 export type AnalyseOptions = {
   highlight?: string[];
+  stubs?: boolean;
 };
 
 program
   .command('analyse <file>')
   .option('--highlight <ids...>')
```
GuyMcComb, line 112: Not too familiar with the analyse command, but do we not need a --no-stubs option here? Just asking because there it is present in AnalyseOptions.

DomHenderson: I've replaced AnalyseOptions with PrintOptions, since that's what it was really describing. The analyse command now takes PrintOptions with stubs  always being undefined since there are no stubs to print or skip in the output from analyse

PR #324, opened 2022-04-27
Replace news with create functions

DomHenderson:
Using the functions in nodeTemplates rather than raw `new` expressions makes it easier to ensure nodes are well formed and have their contexts set correctly. This PR replaces all replaceable uses of `new` expressions with the associated node generation functions


src/utils/getTypeString.ts
```diff
@@ -112,3 +116,28 @@ export function generateLiteralTypeString(value: string): string {
   }
   return `int_const ${value}`;
 }
+
+export function generateExpressionTypeString(type: TypeNode): string {
+  if (type instanceof PointerType) return generateExpressionTypeString(type.to);
```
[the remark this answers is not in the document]
Glitch18: Implemented 👍

src/utils/getTypeString.ts
```diff
@@ -112,3 +119,52 @@ export function generateLiteralTypeString(value: string): string {
...
+  }
+  if (type instanceof TupleType) {
+    return `tuple(${type.elements
+      .map((element) => generateExpressionTypeString(element))
+      .join(',')})`;
+  }
+  if (type instanceof TypeNameType) {
+    return `type(${generateExpressionTypeString(type.type)})`;
+  }
+  return type.pp();
```
DomHenderson, line 169: Instead of returning `type.pp()` if execution gets to this point, this should be an if that checks if `type` is an instance of any of the TypeNode subclasses for which we can safely use the `pp` function. Then an error should be thrown if the node is not any of those types. That way, if solc-typed-ast is updated ands adds a new TypeNode it'll produce an error rather than just using the `pp` function

src/passes/userDefinedTypesConverter.ts
```diff
@@ -57,8 +59,21 @@ export class UserDefinedTypesConverter extends ASTMapper {
...
       );
+    } else if (typeNode instanceof FunctionType) {
+      const newTypeString: string = generateExpressionTypeString(typeNode);
+
+      ast.replaceNode(
+        node,
+        new Identifier(
+          node.id,
+          node.src,
+          newTypeString,
```
DomHenderson, line 70: Better to just assign to the typestring of the existing node here (and in the case above) rather than replacing it with a node of the same type with most properties the same

Glitch18: Implemented 👍

src/utils/getTypeString.ts
```diff
@@ -134,3 +154,66 @@ export function generateLiteralTypeString(
...
+    type instanceof IntLiteralType ||
+    type instanceof IntType ||
+    type instanceof ModuleType ||
+    type instanceof RationalLiteral ||
+    type instanceof StringLiteralType ||
+    type instanceof StringType
+  )
+    return true;
+  return false;
+}
```
rodrigo-pino, line 176: I think that code readability would improve if instead of:
```
if (condition)
    return true;
return false;
```
change it to:
```
return condition;
```

src/passes/externImporter.ts
```diff
@@ -60,7 +52,10 @@ export class ExternImporter extends ASTMapper {
     assert(sourceUnit !== undefined, 'Trying to import a definition into an unknown source unit');
     if (declarationSourceUnit === undefined || sourceUnit === declarationSourceUnit) return;
 
-    if (declaration instanceof FunctionDefinition || declaration instanceof StructDefinition) {
+    if (
+      declaration instanceof FunctionDefinition ||
+      (declaration instanceof StructDefinition && isFree(declaration))
```
rodrigo-pino, line 57: There is previoush check in this function that checks for non-free declarations making `isFree(declaration)` always evaluate to `true`. If it is intended to only apply this condition to structs then that previous check should be removed (line 47)

src/passes/variableDeclarationExpressionSplitter.ts
```diff
@@ -120,6 +119,12 @@ export class VariableDeclarationExpressionSplitter extends ASTMapper {
           node.vDeclarations.push(newDeclaration);
           ast.registerChild(newDeclaration, node);
 
+          console.log(
+            `Creating identifier for ${printNode(newDeclaration)}, new decl for ${
+              oldDeclaration.name
+            }`,
+          );
```
rodrigo-pino, line 126: I believe this logging is unintended

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -1,49 +1,339 @@
...
+            `let (copy${index}) = ${funcName}(${add('loc', storageOffset)})`,
+            `dict_write{dict_ptr=warp_memory}(${add('mem_start', index)}, copy${index})`,
+          ];
+        }
+      }),
+      `    return (mem_start)`,
+      `end`,
+    ].join('\n');
+
+    const funcName = `ws_to_memory${this.generatedFunctions.size}`;
```
GuyMcComb, line 100: Having funcHeader and implicit below the function body makes this hard to read.

GuyMcComb (changes requested):
Monster PR. Single change regarding that position of the funcHeader in storageToMemory and extra line needed.

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -1,49 +1,339 @@
...
+
+    this.requireImport('starkware.cairo.common.dict', 'dict_write');
+    this.requireImport('warplib.memory', 'wm_alloc');
+
+    return funcName;
+  }
+
+  private createStaticArrayCopyFunction(key: string, type: ArrayType): string {
+    assert(type.size !== undefined, 'Expected static array with known size');
+    return type.size <= 5
```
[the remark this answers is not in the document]
DomHenderson: I believe so, but it is the standard cutoff for things like inlining so it makes sense. We can always test and adjust down the line

GuyMcComb (approved):

PR #325, opened 2022-04-28
Functionalise `&&` and `||` when rhs has no side effects

rodrigo-pino:


DomHenderson (changes requested):
Overall this looks really good. Only two changes to request: the side effect checking needs to be more thorough, and we need more tests. Example_contracts already includes some contracts that happen to include ands, or, and conditionals, but it's worth having one or two that are specifically for the purpose of testing that expressions that require short circuiting throw a NotSupported. There should also be some behaviour tests for ands and ors that don't require short circuiting

DomHenderson (approved):

PR #327, opened 2022-05-02
Semantics

DomHenderson:
This PR is the result of reviewing all the solidity semantic tests and running all the relevant ones. It's not feasible to run the semantic tests in the CI, so we should continue to expand behaviour tests with each new PR, but they are an excellent tool to use manually for finding bugs


src/passes/expressionSplitter.ts
```diff
@@ -27,9 +34,35 @@ export class ExpressionSplitter extends ASTMapper {
...
+        node.vLeftHandSide instanceof Identifier &&
+        identifierReferenceStateVar(node.vLeftHandSide)
+      ) {
+        return;
+      }
+
+      const leftHandSide = cloneASTNode(node.vLeftHandSide, ast);
+      const rightHandSide = cloneASTNode(node.vRightHandSide, ast);
+
+      const tempVarStatement = createVariableDeclaration(
```
GuyMcComb, line 48: Feel that this name should be createVariableDeclarationStatement since it returns a statement and not just a VariableDeclaration.

GuyMcComb (approved):

PR #329, opened 2022-05-03
 Import StructDefinition from outside the Contract within the file.

phyBrackets:
It was the problem with importing the struct definition from outside the contract within the file but it seems like there are few edge cases where we are not able to Import struct definition not on every case , In this PR the file _struct_return_struct_ is the one edge case where we are not able to import struct definition. Although I'm working on it what actually creating the problem but I think , atleast where we are able to deal it with correctly we should remove the error _Importing struct definition not supported yet_ , might be there are more edge cases where we are not able to import struct , I'll be trying to find more edge cases where we fail to import.


src/passes/externImporter.ts
```diff
@@ -16,6 +18,49 @@ import { NotSupportedYetError } from '../utils/errors';
...
+      });
+      return;
+    }*/
+  visitVariableDeclaration(node: VariableDeclaration, ast: AST): void {
+    const declaration = node.vType;
+    let declarationSourceUnit;
+    if (declaration instanceof UserDefinedTypeName) {
+      declarationSourceUnit = declaration.vReferencedDeclaration.getClosestParentByType(SourceUnit);
+    } else {
+      declarationSourceUnit = declaration?.getClosestParentByType(SourceUnit);
```
DomHenderson, line 42: If `node.vType` is not a UserDefinedType, there's no need to do anything other than recurse with a `this.commonVisit` and return. `node.vType` will always be in the same sourceUnit as `node`

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -92,13 +45,15 @@ export class IdentifierMangler extends ASTMapper {
...
     }
+
     this.commonVisit(node, ast);
   }
   mangleVariableDeclaration(node: VariableDeclaration): void {
     node.name = this.createNewVariableName(node.name);
   }
   mangleStructDefinition(node: StructDefinition): void {
     node.vMembers.forEach((m) => this.mangleVariableDeclaration(m));
+    //node.name = this.createNewVariableName(node.name);
```
DomHenderson, line 56: Leftover commented code

src/testing.ts
```diff
@@ -68,15 +68,15 @@ const expectedResults = new Map<string, ResultType>([
   ['example_contracts/imports/importfrom', 'Success'],
   ['example_contracts/imports/importInterface', 'Success'],
   ['example_contracts/imports/importLibrary', 'Success'],
-  // Transpilation failure since importing StructDefinition is not implemented yet
-  ['example_contracts/imports/importStruct', 'NotSupportedYet'],
+  // Transpilation Success since importing StructDefinition is now implemented
```
DomHenderson, line 71: The comments in this file are so we can keep track of why things don't work or if things are passing by coincidence. Now that struct imports are supported, the comment can go

DomHenderson (approved):

PR #331, opened 2022-05-04
Source unit

Purva-Chaudhari:
- [x] Added a pass to astChecking.ts that asserts that all id based references refer to nodes that are children of the roots of the AST
- [x] Bug fix for ImportDirectives to refer to the current source unit
- [x] Bug fix for empty ExportedSymbols after souce unit splitting.


cicr99 (changes requested):

cicr99 (approved):
LGTM

PR #332, opened 2022-05-04
Add support for `constructors` on Semantic Tests

rodrigo-pino:


tests/behaviour/expectations/semantic.ts
```diff
@@ -155,27 +159,52 @@ function transcodeTest(
...
+          FunctionKind.Constructor,
+          '',
+          false,
+          FunctionVisibility.Public,
+          FunctionStateMutability.NonPayable,
+          true,
+          new ParameterList(ast.reserveId(), '', []),
+          new ParameterList(ast.reserveId(), '', []),
+          [],
+        ),
```
cicr99, line 201: There is a function `createDefaultConstructor` in `src/passes/inheritanceInliner/constructorInheritance` which does this exact same thing, you could move it into `utils/nodeTemplates` and use it instead.

DomHenderson (approved):

PR #334, opened 2022-05-07
Transpile "argless" fallback functions

rodrigo-pino:


src/utils/astChecking.ts
```diff
@@ -680,7 +681,7 @@ function checkIdNonNegative(node: ASTNode) {
 
 function checkOnlyConstructorsMarkedAsConstructors(nodes: FunctionDefinition[]) {
   nodes.forEach((func) => {
-    if ((func.name === '') != func.isConstructor)
+    if (func.kind === FunctionKind.Constructor && !func.isConstructor)
```
DomHenderson, line 684: This check originally picked up on a bug where IdentifierMangler was giving constructors a non-empty name, so I'd rather not remove the empty name check. How about a function which asserts that all functions with empty names are constructors or fallbacks, and all constructors and fallbacks have empty names?

DomHenderson (changes requested):
Looks great, just one comment to address and it'll be good to go

src/utils/astChecking.ts
rodrigo-pino: On it. Empty names has been the expected behavior of nameless especial functions like `fallback`, `constructor` and `receive`.

DomHenderson (approved):

PR #335, opened 2022-05-09
Inheritance removal

DomHenderson:
InheritanceSpecifiers can contain function calls, which causes issues with later passes which can try to extract those calls into statements because there's nowhere for the statements to go. As InheritanceInliner copies these calls into the constructor, they InheritanceSpecifiers can be safely removed once InheritanceInliner has run


cicr99 (approved):
LGTM

PR #338, opened 2022-05-10
add support for memory array len

RohitRanjangit:
This pr enables the support of length estimation in case of dynamic memory array. for exmaple 
```solidity
pragma solidity ^0.8.6;

contract WARP {

    function dynMemArrayLen() public pure returns (uint256){
        uint[] memory b = new uint[](45);
        b[0] = 1;
        return b.length;
    }
}

```

The `dynMemArrayLen` fn will return `uint256: 45`.


DomHenderson (changes requested):
Looks great, only issue is one leftover debug log. Lmk once it's deleted and I'll approve

DomHenderson (approved):

PR #340, opened 2022-05-10
Compile time type information

rodrigo-pino:


src/passes/typeInformationCalculator.ts
```diff
@@ -0,0 +1,146 @@
...
+              userDef,
+              userDef.vMembers[userDef.vMembers.length - 1],
+              ast,
+            );
+
+      if (userDef instanceof ContractDefinition) {
+        if (memberName === 'name')
+          return createLiteral(userDef.name, LiteralKind.String, ast, node.src, node.raw);
+        if (memberName === 'runtimeCode')
+          throw new NotSupportedYetError('`runtimeCode` member access not supported yet');
```
DomHenderson, line 124: Should probably be WillNotSupport rather than NotSupportedYet for both of these

src/passes/typeInformationCalculator.ts
```diff
@@ -0,0 +1,146 @@
...
+          : createEnumMemberAccess(
+              node,
+              userDef,
+              userDef.vMembers[userDef.vMembers.length - 1],
+              ast,
+            );
+
+      if (userDef instanceof ContractDefinition) {
+        if (memberName === 'name')
+          return createLiteral(userDef.name, LiteralKind.String, ast, node.src, node.raw);
```
DomHenderson, line 122: If you add a createStringLiteral to nodeTemplates I think that covers all the cases here

DomHenderson (changes requested):

DomHenderson (approved):

PR #347, opened 2022-05-16
Reorder structs

rodrigo-pino:


src/passes/orderNestedStructs.ts
```diff
@@ -0,0 +1,123 @@
...
+    // insert back in new order
+    newStructOrder.reverse().forEach((struct) => node.insertAtBeginning(struct));
+  }
+
+  private makeStructTree(
+    structs: readonly StructDefinition[],
+    ast: AST,
+  ): [Set<StructDefinition>, Map<StructDefinition, [StructDefinition]>] {
+    const roots = new Set<StructDefinition>(structs);
+    const tree = new Map<StructDefinition, [StructDefinition]>();
```
DomHenderson, line 64: Assuming you want this to be a one to many struct definition mapping it should be `Map<StructDefinition, StructDefinition[]>`, rather than the current map to arrays of length 1

DomHenderson (changes requested):

DomHenderson (approved):

PR #348, opened 2022-05-17
Add solc with widened addresses

swapnilraj:
Addresses are 256 bits as far as the typechecker is concerned
Address literals have to fit in a felt


JorikSchellekens:
Can we land this @DomHenderson @swapnilraj

swapnilraj:
> Can we land this @DomHenderson @swapnilraj 

Done

PR #352, opened 2022-05-18
Remove 'global' from using for

DomHenderson:
`using Nums for uint256 global` causes a compiler error in the new solc version (0.8.14), but did not in 0.8.13. This removes it from the test it was breaking


GuyMcComb (approved):
LFG

PR #353, opened 2022-05-18
Update EnumConverter pass

Glitch18:
This PR updates the EnumConverter pass to handle typestring better. The pass can now also handle Enum arrays, functions with Enum parameters, etc. 
This PR also adds an example contract which covers various cases like Enum arrays or function types.

### DIFF
--- a/src/passes/enumConverter.ts
+++ b/src/passes/enumConverter.ts
@@ -1,19 +1,29 @@
 import {
+  ArrayType,
+  ElementaryTypeName,
   EnumDefinition,
   enumToIntType,
   Expression,
+  FunctionType,
   getNodeType,
   Identifier,
+  MappingType,
   MemberAccess,
+  PointerType,
+  TupleType,
+  TypeName,
+  TypeNameType,
+  TypeNode,
   UserDefinedType,
   UserDefinedTypeName,
   VariableDeclaration,
 } from 'solc-typed-ast';
+import assert from 'assert';
 import { AST } from '../ast/ast';
 import { ASTMapper } from '../ast/mapper';
 import { TranspileFailedError } from '../utils/errors';
+import { generateExpressionTypeString } from '../utils/getTypeString';
 import { createNumberLiteral } from '../utils/nodeTemplates';
-import { typeNameFromTypeNode } from '../utils/utils';
 
 export class EnumConverter extends ASTMapper {
   getEnumValue(node: EnumDefinition, memberName: string): number {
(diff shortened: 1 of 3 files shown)


src/passes/enumConverter.ts
```diff
@@ -24,33 +35,43 @@ export class EnumConverter extends ASTMapper {
...
-      const enumDef = node.vType.vReferencedDeclaration;
-      if (enumDef instanceof EnumDefinition) {
-        const replacementIntType = enumToIntType(enumDef);
-        const replacementIntTypeName = typeNameFromTypeNode(replacementIntType, ast);
-        ast.replaceNode(node.vType, replacementIntTypeName);
-
-        const replacementTypeString = replacementIntType.pp();
-        node.typeString = replacementTypeString;
-      }
+  visitElementaryTypeName(node: ElementaryTypeName, ast: AST): void {
```
DomHenderson, line 47: I don't think visitElementaryTypeName is necessary for this pass because this pass should never affect ElementaryTypeNames. It was needed in BytesConverter because it does affect them and because the change could not be captured by that already in visitTypeName

src/passes/enumConverter.ts
```diff
@@ -24,33 +35,43 @@ export class EnumConverter extends ASTMapper {
...
-      const enumDef = nType.definition;
-      if (enumDef instanceof EnumDefinition) {
-        node.typeString = this.replacementTypeString(enumDef);
-      }
-    }
+    const typeNode = replaceEnumType(getNodeType(node, ast.compilerVersion));
+    node.typeString = generateExpressionTypeString(typeNode);
   }
 
   visitMemberAccess(node: MemberAccess, ast: AST): void {
```
DomHenderson, line 77: I can't seem to comment on the line in question seeing as it wasn't modified, but the createNumberLiteral here should specify a typestring based on the result of enumToInt. Should technically work either way, but doing so would make it less liable to breaking based on pass order, and should produce more efficient code

DomHenderson (changes requested):

src/passes/enumConverter.ts
Glitch18: Makes sense. But since our `visitTypeName` function uses `getNodeType` instead of `typeNameToTypeNode`, if we come across an `ElementaryTypeName` node with an undefined typestring, it would error out.

DomHenderson: We should never encounter an `ElementaryTypeName` with an undefined typestring. One showing up is an error that needs dealing with at the source, not here

src/passes/enumConverter.ts
Glitch18: Thanks. Implemented this.

src/passes/enumConverter.ts
Glitch18: Implemented a fix for above in #356

DomHenderson (approved):

PR #355, opened 2022-05-19
Memory to storage

DomHenderson:


src/passes/references/expectedLocationAnalyser.ts
```diff
@@ -108,29 +108,37 @@ export class ExpectedLocationAnalyser extends ASTMapper {
...
-            this.expectedLocations.set(node.vArguments[index], structType.location);
+            this.expectedLocations.set(argI, structType.location);
           } else {
             //Finally, default to the type in the pointer itself if we can't infer anything else
-            this.expectedLocations.set(node.vArguments[index], t.location);
+            this.expectedLocations.set(argI, t.location);
           }
+          // Need this case for when you call darray.push([1,2,3]) the tuple gets memory location
+        } else if (argI instanceof TupleExpression && argI.isInlineArray) {
+          this.expectedLocations.set(argI, DataLocation.Memory);
```
DomHenderson, line 136: I'm unconvinced that this is the correct solution in general as a tuple propagates its expected location, and the children should not have expected location memory, e.g. what if they were structs? Solving this should probably be a separate PR though

PR #358, opened 2022-05-19
Update UserDefinedTypesConverter pass

Glitch18:
This PR updates the UserDefinedTypesConverter pass to handle typestrings better. Introduces changes that ensure uniformity in how the three type replacement passes update the typestrings for Expressions and VariableDeclarations.

### DIFF
--- a/src/passes/userDefinedTypesConverter.ts
+++ b/src/passes/userDefinedTypesConverter.ts
@@ -1,13 +1,18 @@
 import {
-  ArrayTypeName,
+  ArrayType,
   ElementaryTypeName,
+  Expression,
   FunctionCall,
   FunctionType,
   getNodeType,
-  Identifier,
-  Mapping,
+  MappingType,
   MemberAccess,
+  PointerType,
+  TupleType,
+  TypeName,
+  typeNameToTypeNode,
   TypeNameType,
+  TypeNode,
   UserDefinedType,
   UserDefinedTypeName,
   UserDefinedValueTypeDefinition,
@@ -16,7 +21,6 @@ import {
 import assert from 'assert';
 import { AST } from '../ast/ast';
 import { ASTMapper } from '../ast/mapper';
-import { TranspileFailedError } from '../utils/errors';
 import { generateExpressionTypeString } from '../utils/getTypeString';
 
 class UserDefinedValueTypeDefinitionEliminator extends ASTMapper {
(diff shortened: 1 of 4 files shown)


src/passes/userDefinedTypesConverter.ts
```diff
@@ -76,15 +68,17 @@ export class UserDefinedTypesConverter extends ASTMapper {
...
+      if (['unwrap', 'wrap'].includes(node.vExpression.memberName)) {
+        const typeNode = getNodeType(node.vExpression.vExpression, ast.compilerVersion);
+        this.commonVisit(node, ast);
+        if (!(typeNode instanceof TypeNameType)) return;
+        if (!(typeNode.type instanceof UserDefinedType)) return;
+        if (!(typeNode.type.definition instanceof UserDefinedValueTypeDefinition)) return;
 
-    ast.replaceNode(node, node.vArguments[0]);
+        ast.replaceNode(node, node.vArguments[0]);
+      } else this.commonVisit(node, ast);
```
DomHenderson, line 80: I believe these should be visitExpression, not commonVisit so that if the typestring of the functionCall contains userdefinedvaluetypes then it can get corrected. Please correct me if I'm wrong

src/passes/userDefinedTypesConverter.ts
```diff
@@ -100,3 +94,43 @@ export class UserDefinedTypesConverter extends ASTMapper {
...
+  } else if (type instanceof PointerType) {
+    return new PointerType(replaceUserDefinedType(type.to), type.location, type.kind, type.src);
+  } else if (type instanceof TupleType) {
+    return new TupleType(type.elements.map(replaceUserDefinedType), type.src);
+  } else if (type instanceof TypeNameType) {
+    return new TypeNameType(replaceUserDefinedType(type.type), type.src);
+  } else if (type instanceof UserDefinedType) {
+    if (type.definition instanceof UserDefinedValueTypeDefinition) {
+      return typeNameToTypeNode(
+        new ElementaryTypeName(
```
DomHenderson, line 125: What's the motivation behind creating a new node here? I don't see what this covers that `typeNameToTypeNode(type.definition.underlyingType)` doesn't

Glitch18: Ah my bad. I didn't think of this. I had the misconception that I had to convert the underlying typestring into an `ElementaryTypeName` node and then call the function on it. Not realising that `type.definition.underlyingType` already has the required node.

src/passes/userDefinedTypesConverter.ts
Glitch18: Actually how I've designed this function is that, if the function call's expression is a member access of a user defined type, it replaces the `FunctionCall` node itself with it's argument expression. Since this function recurses bottom up, the argument expression would have already had it's typestring corrected when it replaces the `FunctionCall` node. 

But this is only to be done in the in the specific case when it's a `wrap` or `unwrap` function call from a user defined type. Otherwise it should recurse normally. Hence the two `else this.commonVisit(node, ast);` functions.

Also, when it is the case of a user defined type's member access, it's irrelevant weather we change the FunctionCall typestring or not since it's gonna get repalced anyway. But I capture the `typeNode` before calling the recursion otherwise the subsequent checks happen after the typestring is replaced and we never reach the `ast.replaceNode` call.

Please let me know if you think this approach is okay, or if I can make it better. Thanks!

DomHenderson (approved):

PR #361, opened 2022-05-23
bytesX index access

RohitRanjangit:
Now we can access the byte at specified index for bytesX types.
```solidity
pragma solidity ^0.8.6;

contract WARP {
    bytes17 a= 0x123456789abcdef0123456789abcdef012;
    // function to return a byte at index access of a
    function get(uint256 index) public view returns (bytes1) {
        return a[index];
    }
}

```
`a[2] = 0xde = 222`.


src/passes/bytesConverter.ts
```diff
@@ -47,6 +53,59 @@ export class BytesConverter extends ASTMapper {
...
+  visitIndexAccess(node: IndexAccess, ast: AST): void {
+    if (
+      !(
+        node.vIndexExpression &&
+        generalizeType(getNodeType(node.vBaseExpression, ast.compilerVersion))[0] instanceof
+          FixedBytesType &&
+        getNodeType(node, ast.compilerVersion) instanceof FixedBytesType
+      )
+    ) {
+      const typeNode = replaceFixedBytesType(getNodeType(node, ast.compilerVersion));
```
DomHenderson, line 65: I think this should just be a call to visitExpression rather than copying the implementation to keep things centralised in case they need changing in the future

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,83 @@
...
+from starkware.cairo.common.bitwise import bitwise_and
+from warplib.maths.utils import narrow_safe
+
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+from starkware.cairo.common.uint256 import Uint256
+from starkware.cairo.common.math_cmp import is_le
+
+from starkware.cairo.common.serialize import serialize_word
+
+func byte_accessor{range_check_ptr}(pos : felt) -> (offset : felt):
```
DomHenderson, line 11: This seems like a very inefficient way to generate a mask. Worth checking the step counts, but I would have thought it'd be more efficient to either do `255 * pow2(index*8)` or `pow2((index+1)*8))-pow2(index*8)`, or even set up a function like pow2 for masks (in which case make sure to write the mask values in hexadecimal to keep things simple)

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,83 @@
...
+    let (le) = is_le(pos, 256)
+    assert le = 1
+
+    let (offset) = pow2(pos)
+    let (offset1) = pow2(pos + 1)
+    let (offset2) = pow2(pos + 2)
+
+    # placing more than 3 bits will exceed the range of felt
+    # and will produce overflowing results
+    if pos == 31 * 8:
```
DomHenderson, line 21: This shouldn't be a concern as long as bounds are checked appropriately

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,83 @@
...
+
+    let (offset3) = pow2(pos + 3)
+    let (offset4) = pow2(pos + 4)
+    let (offset5) = pow2(pos + 5)
+    let (offset6) = pow2(pos + 6)
+    let (offset7) = pow2(pos + 7)
+    return (offset + offset1 + offset2 + offset3 + offset4 + offset5 + offset6 + offset7)
+end
+
+func byte_at_index_uint256{bitwise_ptr : BitwiseBuiltin*, range_check_ptr}(
```
DomHenderson, line 33: These functions need to validate their input based on the width of the `bytesX` type at play. You could either add a width argument to check against or extend generateWarplib.ts to generate functions for each width of FixedBytesType

DomHenderson (changes requested):

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,85 @@
...
+    base : Uint256, index : Uint256, width : felt
+) -> (res : felt):
+    alloc_locals
+
+    let (inRangeHigh : felt) = is_le_felt(base.high, 0xffffffffffffffffffffffffffffffff)
+    let (inRangeLow : felt) = is_le_felt(base.low, 0xffffffffffffffffffffffffffffffff)
+    assert inRangeHigh * inRangeLow = 1
+
+    let (index_felt) = narrow_safe(index)
+    let (base_felt) = narrow_safe(base)
```
DomHenderson, line 79: Narrowing the base like this will fail for values too large to fit into a felt. It's probably easiest to just check the low felt for indexes in the range [0,16), and the high felt (with adjusted index) for [16,32)

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,85 @@
...
+    return (res)
+end
+
+func byte256_at_index_uint256{bitwise_ptr : BitwiseBuiltin*, range_check_ptr}(
+    base : Uint256, index : Uint256, width : felt
+) -> (res : felt):
+    alloc_locals
+
+    let (inRangeHigh : felt) = is_le_felt(base.high, 0xffffffffffffffffffffffffffffffff)
+    let (inRangeLow : felt) = is_le_felt(base.low, 0xffffffffffffffffffffffffffffffff)
```
DomHenderson, line 75: When I said validate their input, I meant they need to check the index. Base can be assumed to be a valid uint256

warplib/maths/bytes_access.cairo
```diff
@@ -0,0 +1,85 @@
...
+from starkware.cairo.common.uint256 import Uint256
+from starkware.cairo.common.math import assert_le_felt
+from starkware.cairo.common.math_cmp import is_le_felt
+
+from starkware.cairo.common.serialize import serialize_word
+
+func byte_accessor{range_check_ptr}(index : felt) -> (offset : felt):
+    assert_le_felt(index, 31)
+    let (pow2index) = pow2(index * 8)
+    if index == 31:
```
DomHenderson, line 15: With the changes suggested below, the highest valid index into a felt is 30 (as that's the highest valid index into a uint248 and for uint256s you can index into each felt, and each up to a maximum of 16), so it doesn't make sense to have a different return here. There should be an assert or just no check if the asserts in the other functions are sufficient

tests/behaviour/contracts/bytes/bytesX.sol
```diff
@@ -0,0 +1,35 @@
...
+        return a[index];
+    }
+
+    function bytes32access(uint112 index) pure public returns (bytes1) {
+        bytes32 a = 0x0211111111111111111111111111111111111111111111111111111111111111;
+        return a[index];
+    }
+
+    function bytes32access256(uint256 index) pure public returns (bytes1) {
+        bytes32 a = 0x0711111111111111111111111111111111111111111111111111111111111111;
```
DomHenderson, line 32: Could you add a test that uses a value that won't fit in a felt, I think that would help cover a lot of the comments I've raised on this

DomHenderson (changes requested):

DomHenderson (approved):

PR #363, opened 2022-05-23
DynArray Memory to Calldata

GuyMcComb:


src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -47,10 +52,10 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
       return this.createStructCopyFunction(key, type);
     } else if (type instanceof ArrayType) {
       if (type.size === undefined) {
-        // return this.createDynamicArrayCopyFunction(key, type);
-        throw new NotSupportedYetError(
-          `Copying ${printTypeNode(type)} from memory to calldata not implemented yet`,
-        );
+        return this.createDynamicArrayCopyFunction(key, type);
+        // throw new NotSupportedYetError(
```
DomHenderson, line 56: commented code needs deleting

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -167,4 +172,104 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
+
+    this.generatedFunctions.set(key, {
+      name: funcName,
+      code: [
+        `func ${funcName}${implicits}(mem_loc: felt) -> (retData: ${outputType.toString()}):`,
+        `    alloc_locals`,
+        `    let (len_256) = wm_read_256(mem_loc)`,
+        `    let (ptr : ${outputType.vPtr.toString()}) = alloc()`,
+        `    let (len_felt) = narrow_safe(len_256)`,
+        `    let (ptr) = ${this.createDynArrayReader(elementT)}(len_256, ptr, mem_loc+ 2)`,
```
DomHenderson, line 211: Why does the dynArrayReader function take len_256 and not just len_felt? I don't see anything in it that requires the length be a Uint256

src/passes/externalArgumentModifier/dynamicArrayReturner.ts
```diff
@@ -0,0 +1,68 @@
...
+  the correct StructDefinitions are created to hold the pointers of dynamic arrays.
+  */
+
+  visitFunctionDefinition(node: FunctionDefinition, ast: AST): void {
+    const body = node.vBody;
+
+    if (
+      isExternallyVisible(node) &&
+      body !== undefined &&
+      body.lastChild instanceof Return &&
```
DomHenderson, line 31: Returns don't always have to be the last child of the body. e.g. if there's an `if` then it'll have two returns inside it

DomHenderson (changes requested):

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
GuyMcComb: This is going to sound crazy, but what if they have more than the felt limit entries lol. 
I actually saw you do this in storageToMemory so I thought that would be more correct.

DomHenderson: I used uint256 in storageToMemory because wm_alloc takes a uint256, because array lengths are uint256, but I don't think we need to worry about people returning more than 2^251 elements from a function at a time

tests/behaviour/expectations/index.ts
```diff
@@ -17,4 +17,4 @@ function filterTests(
   return tests.filter((test) => test.name.includes(filter));
 }
 
-export const expectations = filterTests(behaviour, semantic);
+export const expectations = filterTests(behaviour, semantic, 'copy_memory_to_calldata');
```
DomHenderson, line 20: Leftover filter

src/passes/externalArgumentModifier/dynamicArrayReturner.ts
```diff
@@ -25,20 +25,17 @@ export class DynArrayReturner extends ASTMapper {
...
-    ) {
-      const returnStatement = body.lastChild;
-      //
-      const retExpression = returnStatement.children[0];
-      assert(retExpression instanceof Expression);
-      this.createDynArrayStructDefs(retExpression, ast);
+    if (isExternallyVisible(node) && body !== undefined) {
+      const returnStatements = node
+        .getChildren(true)
+        .filter((n) => n instanceof Return && n.firstChild !== undefined);
```
DomHenderson, line 31: Probably better to use .vExpression instead of .firstChild here and below

DomHenderson (changes requested):

DomHenderson (approved):

PR #368, opened 2022-05-24
Support bytes and strings

JorikSchellekens:


src/transpiler.ts
```diff
@@ -111,9 +111,9 @@ function applyPasses(ast: AST, options: TranspilationOptions & PrintOptions): AS
     ['U', UnloadingAssignment],
     ['V', VariableDeclarationInitialiser],
     ['Vs', VariableDeclarationExpressionSplitter],
+    ['Rf', References],
     ['I', ImplicitConversionToExplicit],
     ['Dh', DeleteHandler],
-    ['Rf', References],
```
rodrigo-pino, line 116: DeleteHandler must go before References because the way we currently handle `memory` type deletion is by creating a new assignation to the default value, and then the Reference pass allocates it.

src/utils/nodeTypeProcessing.ts
```diff
@@ -71,6 +71,11 @@ export function getParameterTypes(functionCall: FunctionCall, ast: AST): TypeNod
 }
 
 export function typeNameToSpecializedTypeNode(typeName: TypeName, loc: DataLocation): TypeNode {
+  try {
+    typeNameToTypeNode(typeName);
+  } catch {
+    console.log(typeName);
+  }
```
rodrigo-pino, line 78: Is this logging intended?

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -5,39 +5,68 @@ import {
...
 */
 export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
-  gen(node: TupleExpression): FunctionCall {
+  stringGen(node: Literal): FunctionCall {
+    // Encode the literal to the uint-8 byte representation
+    assert(node.kind === LiteralKind.String);
+    const byteCode = [];
+    for (let i = 0; i < node.hexValue.length; i++) {
+      byteCode.push(node.hexValue.charCodeAt(i));
+    }
```
rodrigo-pino, line 35: If the intention is to decode the utf8 into bytes I think this is not correct because `charCodeAt` does not return the number which represents the byte representation of the character. It returns instead a number  which  represents the range where the char is in utf8 from which you can then get the bytes representation of the char. [Wikipedia](https://en.wikipedia.org/wiki/UTF-8) explains better than me.

A possible solution for this is using the `Buffer` class like:
```
const a = Buffer.from("some string")
const byteCode = a.toJSON().data
```

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -67,16 +96,26 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
...
+        `    let (start) = wm_alloc(${uint256(alloc_len)})`,
+        [
+          ...(dynamic ? [`wm_write_256{dict_ptr=warp_memory}(start, ${uint256(size)})`] : []),
+          ...mapRange(size, (n) => elementCairoType.serialiseMembers(`e${n}`))
+            .flat()
+            .map(
+              (name, index) =>
+                `dict_write{dict_ptr=warp_memory}(${add(
+                  'start',
+                  dynamic ? index + 2 : index,
```
rodrigo-pino, line 115: Just curious, why is index increased by two when it is a dynamic array

src/transpiler.ts
DomHenderson: We also need References after ImplicitToExplicit for array state variable initialisation to work properly. The rvalue array needs to have been converted to the right type before the References pass allocates the space for that type

src/cairoUtilFuncGen/memory/arrayLiteral.ts
JorikSchellekens: The length is stored as a Uint256 at the start of the memory area

JorikSchellekens:
This pr doesn't support string literal storage var initialisers. I've asked @rodrigo-pino to look into it

src/passes/bytesConverter.ts
```diff
@@ -35,20 +39,24 @@ import {
...
+    const typeNode = replaceBytesType(getNodeType(node, ast.compilerVersion));
     node.typeString = generateExpressionTypeString(typeNode);
     this.commonVisit(node, ast);
   }
 
   visitElementaryTypeName(node: ElementaryTypeName, ast: AST): void {
     const typeNode = typeNameToTypeNode(node);
-    const replacementTypeNode = replaceFixedBytesType(typeNode);
+    if (typeNode instanceof StringType || typeNode instanceof BytesType) {
+      replaceNode(node, createArrayTypeName(createUint8TypeName(ast), ast));
```
DomHenderson, line 56: This should be ast.replaceNode, I should have been clearer when you mentioned that

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -5,39 +5,68 @@ import {
...
-  Converts [a,b,c] into WM0_arr(a,b,c), which allocates new space in warp_memory
+  Converts [a,b,c] and "abc" into WM0_arr(a,b,c), which allocates new space in warp_memory
   and assigns the given values into that space, returning the location of the
   start of the array
 */
 export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
-  gen(node: TupleExpression): FunctionCall {
+  stringGen(node: Literal): FunctionCall {
+    // Encode the literal to the uint-8 byte representation
+    assert(node.kind === LiteralKind.String);
```
DomHenderson, line 32: What about LiteralKind.UnicodeString (or even hexString, is that relevant here?)

src/passes/implicitConversionToExplicit.ts
```diff
@@ -294,7 +294,7 @@ function insertConversionIfNecessary(expression: Expression, targetType: TypeNod
       `Unexpected unresolved rational literal ${printNode(expression)}`,
     );
   } else if (currentType instanceof StringLiteralType) {
-    insertConversion(expression, targetType, ast);
+    return;
```
DomHenderson, line 297: Is something not needed here so that we can handle both string literals that become arrays and string literals that become fixedBytesTypes?

src/utils/cairoTypeSystem.ts
```diff
@@ -77,7 +77,12 @@ export abstract class CairoType {
     } else if (tp instanceof BuiltinStructType) {
       throw new NotSupportedYetError('Serialising BuiltinStructType not supported yet');
     } else if (tp instanceof BytesType) {
-      throw new NotSupportedYetError('Serialising BytesType not supported yet');
+      switch (context) {
+        case TypeConversionContext.CallDataRef:
+          return new CairoPointer(new CairoFelt());
```
DomHenderson, line 82: This should be CairoDynArray, matching the case for dynamic array I think

warplib/memory.cairo
```diff
@@ -119,6 +126,143 @@ func wm_dyn_array_length{warp_memory : DictAccess*}(arrayLoc : felt) -> (len : U
     return (Uint256(low, high))
 end
 
+func wm_bytes_new{range_check_ptr, warp_memory : DictAccess*}(len : Uint256) -> (loc : felt):
```
DomHenderson, line 129: As glorious as this all is it should probably be commented/moved to another file to be clear it's for future use and not for now

DomHenderson (commented):
As requested, a preliminary look. Overall this is great though

src/passes/implicitConversionToExplicit.ts
JorikSchellekens: Would it be acceptable to replace it with an int literal of the correct size at this location?

DomHenderson (approved):

PR #372, opened 2022-05-24
Stored pointers

DomHenderson:
Replaces the pattern of setting expected location to default for complex types that need an extra read with a modification of dataAccessFunctionaliser's visitExpression to add the extra reads where necessary


tests/behaviour/contracts/copy_memory_to_calldata/dynArray.sol
```diff
@@ -35,71 +35,68 @@ function returnUint256() pure external returns (uint256[] memory) {
...
+     y[2] = 3;
+
+     return (y, x);
+}
+
+function returnDynArrayAsMemberAccess() pure external returns (uint[] memory){
+    uint256[] memory y = new uint256[](2);
+    y[0] = 10;
+    y[1] = 100;
+   // y[2] = 1000;
```
GuyMcComb, line 75: Could you delete this line in the next PR that is rebased off this one? I don't want the CI to have to run again.

GuyMcComb (commented):
Glad we got this working in a generalized way. One comment on a line that should be deleted, but get rid of it in the next PR to prevent the CI running again.

GuyMcComb (approved):

PR #375, opened 2022-05-24
Memory copy: calldata <-> storage

rodrigo-pino:
Handles copies of reference types from calldata to storage and storage to calldata


tests/behaviour/contracts/copy_calldata_to_storage/static_arrays.sol
```diff
@@ -0,0 +1,10 @@
+contract WARP {
```
GuyMcComb, line 1: Compiler version.

GuyMcComb:
Looks good, does this support nested reference types, obviously referring to the ones that are possible to pass in and out of functions using call data:

- Structs of structs
- Dyn/Static arrays of Structs?

PR #379, opened 2022-05-26
add support for % operator

RohitRanjangit:
- [x] Add behaviour test
- [x] Fix `arithmetics/signed_mod.sol` semantic test error


warplib/maths/mod_signed.cairo
```diff
@@ -0,0 +1,384 @@
+# AUTO-GENERATED
```
DomHenderson, line 1: The warplib gitignore should be used to exclude the autogenerated warplib files from being included in git

DomHenderson (changes requested):
Looks good, almost ready to merge

warplib/maths/mod_signed.cairo
RohitRanjangit: Warplib ingores are moved to root gitignore file . And, there is already an entry for this in .gitignore. I'll delete this file and will push.

DomHenderson (approved):

PR #386, opened 2022-05-27
Index static arrays

rodrigo-pino:
Add a pass for indexing in static arrays


src/passes/staticArrayIndexer.ts
```diff
@@ -0,0 +1,207 @@
...
+
+function getReferenceDeclaration(expr: Expression, ast: AST): Identifier {
+  if (expr instanceof Identifier) {
+    return expr;
+  } else if (expr instanceof IndexAccess) {
+    return getReferenceDeclaration(expr.vBaseExpression, ast);
+  } else if (expr instanceof MemberAccess) {
+    return getReferenceDeclaration(expr.vExpression, ast);
+  }
+  throw new Error(`Unexpected expression ${printNode(expr)} while searching for identifier`);
```
DomHenderson, line 206: Should be TranspileFailedError

src/passes/staticArrayIndexer.ts
```diff
@@ -0,0 +1,207 @@
...
+    if (expr instanceof Identifier) {
+      replacement = createIdentifier(refVarDecl, ast, DataLocation.Memory);
+    } else if (expr instanceof IndexAccess) {
+      assert(expr.vIndexExpression instanceof Expression);
+      replacement = new IndexAccess(
+        expr.id,
+        expr.src,
+        generateExpressionTypeString(exprMemoryType),
+        this.setExpressionToMemory(expr.vBaseExpression, refVarDecl, ast),
+        cloneASTNode(expr.vIndexExpression, ast),
```
DomHenderson, line 101: Is this clone necessary? I would have thought that if the old node is getting replaced then it should suffice to move its children without cloning them

src/passes/staticArrayIndexer.ts
```diff
@@ -0,0 +1,207 @@
...
+      '',
+      false,
+      false,
+      `memory_${identifier.name}`,
+      parentFunction?.id,
+      false,
+      DataLocation.Memory,
+      StateVariableVisibility.Internal,
+      Mutability.Mutable,
+      generateExpressionTypeString(memoryType),
```
DomHenderson, line 143: VariableDeclarations have different typestrings to expressions as they don't contain the memory location. I think what you'll want to do here is to both not specialise `memoryType` and just call `.pp()` on it rather than `generateExpressionTypeString`

DomHenderson (changes requested):

DomHenderson (approved):

PR #387, opened 2022-05-27
Array element conversion

DomHenderson:
Fixes case where elements inside inline arrays were not being converted to the element type of the array


example_contracts/typeConversion/unusedArrayConversion.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.8.10;
+
+// SPDX-License-Identifier: MIT
+
+contract WARP {
+  function test() public {
+    uint a;
+    uint8 b;
+    [a,b,3];
```
GuyMcComb, line 9: out of interest what is the default type chosen to be and does this work with assignments such as
`uint256[] = [a,b,3]`

GuyMcComb (changes requested):

example_contracts/typeConversion/unusedArrayConversion.sol
GuyMcComb: Just ran an example and saw that it works.

GuyMcComb (approved):

PR #392, opened 2022-05-28
Enumtoint

phyBrackets:
Something's was going wrong with enum<->int conversions. the test at _tests/behaviour/solidity/test/libsolidity/semanticTests/enums/constructing_enums_from_ints.sol_ . There was the problem in _enumConverter.ts_ , Added a visitFunctionCall method to EnumConverter to transform the Identifier child of enum TypeConversion FunctionCalls into the appropriate typename.


src/passes/enumConverter.ts
```diff
@@ -34,6 +37,37 @@ export class EnumConverter extends ASTMapper {
     return val;
   }
 
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    this.visitExpression(node, ast);
+    if (node.kind !== FunctionCallKind.TypeConversion) return;
+    const tNode = getNodeType(node.vExpression, ast.compilerVersion);
+    assert(
+      tNode instanceof TypeNameType,
+      `Got non-typename type ${tNode.pp()} when parsing conversion functiobin/warp transpile --print-trees tests/behaviour/solidity/test/libsolidity/semanticTests/enums/constructing_enums_from_ints.sol
```
DomHenderson, line 46: I think you must have accidentally pasted the command into this comment

DomHenderson (changes requested):
Could you please add a test into the behaviour tests that covers the case fixed here so that it runs as part of the CI. Other than that just one comment and it should be good to merge

src/passes/enumConverter.ts
phyBrackets: Yeah, Sorry for this.

JorikSchellekens (approved):

PR #393, opened 2022-05-28
Prepare warp for packaging 

swapnilraj:
Adds a new cli option `warp install` to setup the required python venv for cairo-lang
Prefixes the calls to cairo-lang packages with an updated path that includes the venv


JorikSchellekens:
Beautiful, if we can get the post install doing some heavy lifting to resolve `python3.7` then installing would be a seamless experience

JorikSchellekens (approved):

swapnilraj:
> Beautiful, if we can get the post install doing some heavy lifting to resolve `python3.7` then installing would be a seamless experience

On first look there is no good way of doing this. We will have to write a custom thing for this and I fear it'll be brittle and a decent overhead. I'll look into packaging a python3.7 interpreter or downloading one with a script (pyenv kinda thing).

PR #396, opened 2022-05-28
Add account option to all cli commands

JorikSchellekens:
Also defaults to the default account used by the starknet cli


swapnilraj (approved):

src/index.ts
```diff
@@ -134,6 +134,10 @@ program
     'Arguments to be passed to constructor of the program.',
     undefined,
   )
+  .option(
+    '--account <account>',
```
DomHenderson, line 138: Deploy has an account option but no wallet option, is this intentional? My understanding was that deploying does not use a wallet or account

JorikSchellekens: Ah, yes

DomHenderson (approved):

PR #397, opened 2022-05-28
Cleanup SourceUnit filepath mangling

JorikSchellekens:
Now that #395 replaces `_` with `__` we can replace our `__WARP_CONTRACT__` with `_` and guarantee uniqueness


JorikSchellekens:
@swapnilraj pointed out this might not be the best idea. So I'm closing it

PR #398, opened 2022-05-29
Skip test fix

DomHenderson:
The test suite changes to prevent unlimited multithreading also broke the skipping of the compile stage if the transpile stage fails. This fixes it


phyBrackets (commented):
_res.result_ make more sense to me , really don't know how previously it worked only with the _res_ . I think this is ready to merge .

GuyMcComb (approved):

PR #399, opened 2022-05-29
Increase max buffer size for solc json output

JorikSchellekens:
The buffer carrying the json representation of the solidity contracts was being maxed out for intermediate sized contracts (for example https://github.com/JorikSchellekens/warp-demo/blob/main/ERC20/presets/ERC20PresetFixedSupply.sol). 

This is a short term fix

### DIFF
--- a/src/solCompile.ts
+++ b/src/solCompile.ts
@@ -89,6 +89,7 @@ function cliCompile(
   const nethersolcVersion: SupportedSolcVersions = solcVersion.startsWith('0.7.') ? `7` : `8`;
   const solcCommand = nethersolcPath(nethersolcVersion);
 
+  let allowedPaths = '';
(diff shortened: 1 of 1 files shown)


src/solCompile.ts
```diff
@@ -96,14 +101,18 @@ function cliCompile(
       result: JSON.parse(
         execSync(`${solcCommand} --standard-json --allow-paths ${allowPath}`, {
```
swapnilraj, line 102: Change looks good to me.
Can we change the structure so that we don't have the repetition in the two branches

swapnilraj (dismissed):

src/solCompile.ts
swapnilraj: done

DomHenderson (approved):

PR #401, opened 2022-05-30
Ref input checks

GuyMcComb:
Changing the external function argument checks to handle reference types.

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -1,5 +1,6 @@
 import assert from 'assert';
 import {
+  ArrayType,
   ArrayTypeName,
   Assignment,
   ASTNode,
@@ -32,6 +33,7 @@ import {
   FunctionStateMutability,
   FunctionTypeName,
   FunctionVisibility,
+  generalizeType,
   getNodeType,
   Identifier,
   IdentifierPath,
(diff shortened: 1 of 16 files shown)


warplib/maths/external_input_check_address.cairo
```diff
@@ -0,0 +1,12 @@
+# AUTO-GENERATED
+from starkware.cairo.common.math_cmp import is_le_felt
+from starkware.cairo.common.uint256 import Uint256
+
+func warp_external_input_check_address{range_check_ptr}(x : felt):
```
swapnilraj, line 5: I don't think we need this, starknet's calldata checks should be enough now.

src/cairoWriter.ts
```diff
@@ -716,7 +717,7 @@ class FunctionCallWriter extends CairoASTNodeWriter {
           }
         } else if (
           node.vReferencedDeclaration instanceof CairoFunctionDefinition &&
-          node.vReferencedDeclaration.functionStubKind === FunctionStubKind.StructDefStub
+          node.vReferencedDeclaration.acceptsRawDarray
         ) {
           return [`${func}(${args}_len, ${args})`];
         }
```
swapnilraj, line 723: I can't comment on the specific line, but line 734 checks address literals.
The starknet field prime is slightly bigger than 2 ** 251.
FIELD_PRIME = 2 ** 251 + 17 * 2 ** 192 + 1
[starkware-libs/cairo-lang@cf8266f/src/starkware/crypto/starkware/crypto/signature/nothing_up_my_sleeve_gen.py#L35](https://github.com/starkware-libs/cairo-lang/blob/cf8266fd5d1ff66962579ff7967ac5cdcf699f77/src/starkware/crypto/starkware/crypto/signature/nothing_up_my_sleeve_gen.py#L35)

src/passes/externalInputChecker/inputBoundChecker.ts
```diff
@@ -0,0 +1,68 @@
...
+  private insertFunctionCall(node: FunctionDefinition, funcCall: FunctionCall, ast: AST): void {
+    const body = node.vBody;
+    assert(body !== undefined && funcCall.vArguments !== undefined);
+    const expressionStatement = createExpressionStatement(ast, funcCall);
+    body.insertAtBeginning(expressionStatement);
+    ast.setContextRecursive(expressionStatement);
+  }
+
+  checkableType(type: TypeNode): boolean {
+    if (
```
swapnilraj, line 55: minor nit: you can just return the if condition

src/cairoWriter.ts
DomHenderson: We asked at a recent meeting with Starkware and they said the addresses are 251 bits wide. Is there something that suggests they take up the fult set of felt values?

swapnilraj: Nope, I just assumed that they would take the full felt space.

swapnilraj (approved):

PR #402, opened 2022-05-30
Fix fixed length array encoding for semantic tests

DomHenderson:
Fixed length arrays are encoded as tuples, not as ptr,len pairs, so the prepended length is not needed


rodrigo-pino (approved):
All good

PR #403, opened 2022-05-30
Array element location

DomHenderson:
Fixes a bug where scalar types in inline arrays were given the expected location of the array, not default


rodrigo-pino (approved):
Looks good!

PR #407, opened 2022-05-30
Benchmark: Add JSON file size, and group functions by contract

Glitch18:
This PR adds a new field in the benchmark denoting the size of the JSON file given to the testnet. 
This PR also groups the function names by their contract. Earlier, the invoked functions were being written out individually which was not very readable. 

This PR also changes the CI to push benchmarks only when a commit is merged and not on pull requests. This will clear up clutter on the warp-ts-benchmark branch.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,7 +79,7 @@ yarn test
 In order to generate benchmarks locally during development:
 
 ```bash
-yarn testnet-b
+yarn testnet:benchmark
 yarn test
 ```
 
--- a/readme.md
+++ b/readme.md
@@ -76,6 +76,19 @@ then to run the tests:
 yarn test
 ```
 
+In order to generate benchmarks locally during development:
+
+```bash
+yarn testnet-b
+yarn test
+```
+
+```python
+python starknet-testnet/generateMarkdown.py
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -76,6 +76,19 @@ then to run the tests:
 yarn test
 ```
 
+In order to generate benchmarks locally during development:
+
+```bash
+yarn testnet-b
```
JorikSchellekens, line 82: Can we use `testnet:benchmark` instead? All the other scripts use this convention

JorikSchellekens (dismissed):

readme.md
Glitch18: Thanks, Implemented 👍

Glitch18:
> Just a nitpick. Otherwise it looks good.
> 
> For further work it would be nice to get an overview of how the step counts are changing. I know there were some graphs for warp 1. Can we get the same for warp 2?

Yes, I'm working on adding graphs similar to warp v1 benchmarks. I'll post it as a different PR by end of this week hopefully.

JorikSchellekens (approved):

PR #408, opened 2022-05-30
Fix memoryToCalldata not reading nested complex types properly

DomHenderson:
Nested complex types are stored as 'pointers' in memory and need to be read before being passed to nested copy functions


rodrigo-pino (approved):
All good

PR #410, opened 2022-05-30
Adding input checks for nested reference types that we do not support

GuyMcComb:
Adding reject for nested reference types inputs into external functions.


example_contracts/functionInputs/arrayTest/arrayDynArrayDynArray.sol
```diff
@@ -0,0 +1,8 @@
+pragma solidity ^0.8.6;
+
+// SPDX-License-Identifier: MIT
+
+contract WARP {
+    function inputFunc(uint8[][][3] memory y) pure external {
+    }
+}
```
phyBrackets, line 8: can you please add extra line after this? This not having much affect on the result but good practice to add extra line at the end . Update other files as well where you can see this red circle.

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -113,10 +119,49 @@ export class RejectUnsupportedFeatures extends ASTMapper {
   }
 
   visitFunctionDefinition(node: FunctionDefinition, ast: AST): void {
+    if (isExternallyVisible(node)) {
+      node.vParameters.vParameters.forEach((decl) => {
+        const type = getNodeType(decl, ast.compilerVersion);
```
phyBrackets, line 124: just for the sake of maintaining the code structure , it's better to change the name  _type_  to  _typeNode_

phyBrackets (changes requested):

src/testing.ts
```diff
@@ -76,6 +76,33 @@ const expectedResults = new Map<string, ResultType>([
...
+  ['example_contracts/functionInputs/dynArrayTest/dynArrayStructDynArray', 'NotSupportedYet'],
+  ['example_contracts/functionInputs/dynArrayTest/dynArrayStructStruct', 'Success'],
+  ['example_contracts/functionInputs/structTest/structArrayArray', 'Success'],
+  ['example_contracts/functionInputs/structTest/structArrayDynArray', 'NotSupportedYet'],
+  ['example_contracts/functionInputs/structTest/structArrayStruct', 'Success'],
+  ['example_contracts/functionInputs/structTest/structDynArrayArray', 'NotSupportedYet'],
+  ['example_contracts/functionInputs/structTest/structDynArrayDynArray', 'NotSupportedYet'],
+  ['example_contracts/functionInputs/structTest/structDynArrayStruct', 'NotSupportedYet'],
+  ['example_contracts/functionInputs/structTest/structStructArray', 'Success'],
+  ['example_contracts/functionInputs/structTest/structStructDynArray', 'NotSupportedYet'],
```
DomHenderson, line 104: As discussed there should be WillNotSupport, at least until we tidy up the wording to be clear that it means Will Not Support until dependencies change, vs NotSupportedYet being used as a "not implemented yet but needs doing"

DomHenderson (changes requested):

BitBaseBit (approved):

BitBaseBit (approved):

phyBrackets (approved):

DomHenderson (approved):

PR #412, opened 2022-05-30
Type conversion

rodrigo-pino:


src/cairoUtilFuncGen/memory/implicitCoversion.ts
```diff
@@ -0,0 +1,287 @@
...
+    const targetType = getNodeType(targetExpression, this.ast.compilerVersion);
+    const sourceType = getNodeType(sourceExpression, this.ast.compilerVersion);
+
+    const targetBaseType = getBaseType(targetType);
+    const sourceBaseType = getBaseType(sourceType);
+
+    const targetBaseCairoType = CairoType.fromSol(
+      targetBaseType,
+      this.ast,
+      TypeConversionContext.MemoryAllocation,
```
DomHenderson, line 46: I'm unconvinced about this, and think it might be the cause of some of the issues you mentioned. Types inside a memory array are allocated with TypeConversionContext Ref, only the top level uses TypeConversionContext Memory. Checking whether a conversion is necessary should probably use the solidity types rather than the cairo types

tests/behaviour/contracts/conversions/storageArrays.sol
```diff
@@ -0,0 +1,102 @@
+pragma solidity ^0.8.10;
+
+//SPDX-License-Identifier: MIT
+
+contract WARP {
+    uint[] x;
+    uint[5] y;
+    function tryX1(uint8[] memory m) public returns (uint[] memory) {
```
DomHenderson, line 8: Worth adding tests that use int here as well as uint, as a conversion function is needed to widen ints, even when they're both felts

DomHenderson (changes requested):

DomHenderson (approved):

PR #418, opened 2022-06-01
Ref input checks v2

GuyMcComb:


src/cairoUtilFuncGen/inputArgCheck/inputCheck.ts
```diff
@@ -0,0 +1,267 @@
...
+      type instanceof BoolType
+    );
+  }
+
+  checkForImport(type: TypeNode): void {
+    if (
+      type instanceof UserDefinedType &&
+      (type.definition instanceof StructDefinition || type.definition instanceof EnumDefinition)
+    ) {
+      assert(this.sourceUnit !== undefined);
```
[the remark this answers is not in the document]
DomHenderson: Should probably be added to CairoUtilFuncGenBase tbh

src/cairoUtilFuncGen/inputArgCheck/inputCheck.ts
```diff
@@ -0,0 +1,267 @@
...
+export class InputCheckGen extends StringIndexedFuncGen {
+  sourceUnit: SourceUnit | undefined;
+  gen(node: VariableDeclaration, nodeInSourceUnit: ASTNode): FunctionCall {
+    const type = generalizeType(getNodeType(node, this.ast.compilerVersion))[0];
+    const functionInput = createIdentifier(node, this.ast);
+    this.sourceUnit = this.ast.getContainingRoot(nodeInSourceUnit);
+    const name = this.getOrCreate(type);
+    const functionStub = createCairoFunctionStub(
+      name,
+      [['ref_var', typeNameFromTypeNode(type, this.ast), DataLocation.CallData]],
```
DomHenderson, line 44: This should be a locationIfComplexType rather than just always being CallData, even for scalars

src/cairoUtilFuncGen/inputArgCheck/inputCheck.ts
```diff
@@ -0,0 +1,267 @@
...
+    });
+    return funcName;
+  }
+
+  private createStaticArrayInputCheck(key: string, type: ArrayType): string {
+    const funcName = `extern_input_check${this.generatedFunctions.size}`;
+    const implicits = '{range_check_ptr : felt}';
+
+    // Set an empty entry so recursive function generation doesn't clash
+    this.generatedFunctions.set(key, { name: funcName, code: '' });
```
DomHenderson, line 137: All the instances of doing this can be moved into getOrCreate so there's only one and no chance of missing one

src/cairoUtilFuncGen/inputArgCheck/inputCheck.ts
```diff
@@ -0,0 +1,267 @@
...
+      assert(typeDefSourceUnit instanceof SourceUnit);
+      if (this.sourceUnit !== typeDefSourceUnit) {
+        this.requireImport(this.formatPath(typeDefSourceUnit.absolutePath), type.definition.name);
+      }
+    } else if (type instanceof IntType && type.nBits === 256) {
+      this.requireImport('starkware.cairo.common.uint256', 'Uint256');
+    }
+  }
+
+  formatPath(path: string): string {
```
DomHenderson, line 262: Can this be exported/imported rather than having the same function here and in externImporter

src/passes/externalInputChecker.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  private insertFunctionCall(node: FunctionDefinition, funcCall: FunctionCall, ast: AST): void {
+    const body = node.vBody;
+    assert(body !== undefined && funcCall.vArguments !== undefined);
+    const expressionStatement = createExpressionStatement(ast, funcCall);
+    body.insertAtBeginning(expressionStatement);
+    ast.setContextRecursive(expressionStatement);
+  }
+
+  checkableType(type: TypeNode): boolean {
```
DomHenderson, line 54: Can this also be imported/exported so that if it changes in time we don't need to remember to update both versions

src/warplib/implementations/external_input_checks/external_input_checks_address.ts
```diff
@@ -0,0 +1,22 @@
+import { generateFile, mask } from '../../utils';
+
+const INDENT = ' '.repeat(4);
+
+const import_strings: string[] = [
+  'from starkware.cairo.common.math_cmp import is_le_felt',
+  'from starkware.cairo.common.uint256 import Uint256',
+];
+
+const AddressBoundChecker: Array<string> = [
```
DomHenderson, line 10: How come this is in a warplib file gen file rather that being handwritten cairo when there's only one function generated?

DomHenderson (changes requested):

DomHenderson (approved):

PR #421, opened 2022-06-01
Storage static to dynamic copies

DomHenderson:


ex.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.4.0 <0.9.0;
+
+contract WARP {
+  uint[] x = y;
+  uint[3] y;
+}
+
```
rodrigo-pino, line 8: Is this contract used anywhere?

DomHenderson: Nope, you're right it should have been removed

src/cairoUtilFuncGen/storage/copyToStorage.ts
```diff
@@ -122,71 +123,182 @@ export class StorageToStorageGen extends StringIndexedFuncGen {
...
+    const [fromElementMapping, fromLengthMapping] = this.dynArrayGen.gen(fromElementCairoType);
+    const [toElementMapping, toLengthMapping] = this.dynArrayGen.gen(fromElementCairoType);
+
+    const copyCode = createElementCopy(toElementCairoType, fromElementCairoType, elementCopyFunc);
 
     return {
       name: funcName,
       code: [
-        `func ${funcName}_elem${implicits}(fromLoc: felt, toLoc: felt, length: Uint256) -> ():`,
+        `func ${funcName}_elem${implicits}(toLoc: felt, fromLoc: felt, length: Uint256) -> ():`,
```
GuyMcComb, line 215: Just making sure you are aware that this function is writing in the array from end to beginning since the length and index  are the same value

DomHenderson: Yep, no special reason other than that it was nicer to write the function like that

PR #422, opened 2022-06-01
Return index access bug

yaziciahmet:
While returning an index of an array `return arr[0];`, both length and value (with a wrong syntax) were returned previously, while it should have been only the value.


src/cairoWriter.ts
```diff
@@ -685,7 +685,8 @@ class IdentifierWriter extends CairoASTNodeWriter {
       isDynamicCallDataArray(getNodeType(node, this.ast.compilerVersion)) &&
       ((node.getClosestParentByType(Return) !== undefined &&
         node.getClosestParentByType(FunctionDefinition)?.visibility ===
-          FunctionVisibility.External) ||
+          FunctionVisibility.External &&
+        node.getClosestParentByType(IndexAccess) === undefined) ||
```
DomHenderson, line 689: This will likely need changing in the calldata array length PR, but since that's the only case I think, it's fine for now

DomHenderson (approved):

PR #423, opened 2022-06-01
Add solc with 2^251 addresses

swapnilraj:
The previous nethersolc used STARKNET_FIELD_PRIME, which is slightly bigger than 2^251


swapnilraj:
> Just pinging ohad with a confirmation on this.

Dom confirmed it in a different PR

JorikSchellekens (approved):

PR #426, opened 2022-06-02
`this` methods calls

RohitRanjangit:
Support `this.method_name(..args)`

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -737,10 +737,15 @@ class FunctionCallWriter extends CairoASTNodeWriter {
             nodeType instanceof UserDefinedType &&
             nodeType.definition instanceof ContractDefinition
           ) {
+            const currentContractName = node.getClosestParentByType(ContractDefinition)?.name;
             const contractType = nodeType.definition.name;
             const memberName = node.vExpression.memberName;
             const contract = writer.write(node.vExpression.vExpression);
-            return [`${contractType}.${memberName}(${contract}${args ? ', ' : ''}${args})`];
+            return [
+              `${contractType}${
+                currentContractName === contractType ? '_interface_for_this' : ''
+              }.${memberName}(${contract}${args ? ', ' : ''}${args})`,
+            ];
           }
         } else if (
           node.vReferencedDeclaration instanceof CairoFunctionDefinition &&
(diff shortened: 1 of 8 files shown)


src/passes/builtinHandler/thisKeyword.ts
```diff
@@ -29,4 +39,93 @@ export class ThisKeyword extends ASTMapper {
...
     }
   }
+
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    if (
+      node.vExpression instanceof MemberAccess &&
+      node.vExpression.vExpression instanceof Identifier &&
+      node.vExpression.vExpression.name === 'this'
+    ) {
+      const replacementCall = createCallToFunction(
```
DomHenderson, line 49: Is there a reason to have this here rather than relying on visitIdentifier to do the transformation when the recursion reaches it?

src/cairoWriter.ts
```diff
@@ -737,10 +737,15 @@ class FunctionCallWriter extends CairoASTNodeWriter {
...
             nodeType.definition instanceof ContractDefinition
           ) {
+            const currentContractName = node.getClosestParentByType(ContractDefinition)?.name;
             const contractType = nodeType.definition.name;
             const memberName = node.vExpression.memberName;
             const contract = writer.write(node.vExpression.vExpression);
-            return [`${contractType}.${memberName}(${contract}${args ? ', ' : ''}${args})`];
+            return [
+              `${contractType}${
+                currentContractName === contractType ? '_interface_for_this' : ''
```
DomHenderson, line 746: I'm not a fan of the 'interface_for_this' appearing across the codebase as string literals since it might change depending on mangling. Can it be encapsulated into a mangleOwnContractInterface function or similar, that gets shared between the files that use it?

DomHenderson (changes requested):

src/passes/builtinHandler/thisKeyword.ts
RohitRanjangit: To avoid unnecessary interface creation for the current contract, for eg.
```solidity
contract B{
  address addr = address(this);
}
```
In the above case interface is not needed because there is no any method call to `this`

src/cairoWriter.ts
RohitRanjangit: I'm not sure what I've done matches what'd you had expect, could you see it?

DomHenderson (approved):

PR #427, opened 2022-06-02
Push fixes

DomHenderson:
Uses https://github.com/NethermindEth/warp/pull/421 to fix push for complex types


src/passes/implicitConversionToExplicit.ts
```diff
@@ -257,34 +256,42 @@ export class ImplicitConversionToExplicit extends ASTMapper {
...
+    if (currentLoc === DataLocation.Memory) {
+      const parent = expression.parent;
+      const [replacement, shouldReplace] = ast
+        .getUtilFuncGen(expression)
+        .memory.convert.genIfNecesary(expression, targetType);
+      if (shouldReplace) {
+        ast.replaceNode(expression, replacement, parent);
+      }
     }
+    // if (expression instanceof TupleExpression && expression.isInlineArray) {
```
GuyMcComb, line 282: Commented code

src/passes/implicitConversionToExplicit.ts
```diff
@@ -257,34 +256,42 @@ export class ImplicitConversionToExplicit extends ASTMapper {
...
-        insertConversionIfNecessary(element, elementT, ast);
-      });
-      expression.typeString = generateExpressionTypeString(
-        specializeType(targetType, DataLocation.Memory),
-      );
+    if (currentLoc === DataLocation.Memory) {
+      const parent = expression.parent;
+      const [replacement, shouldReplace] = ast
+        .getUtilFuncGen(expression)
+        .memory.convert.genIfNecesary(expression, targetType);
```
GuyMcComb, line 277: Just noting that there are now two memory.convert in the codebase. One here and one in dataAccessFunctionalizer. I am guessing there are edge cases for both?

GuyMcComb (changes requested):

src/passes/implicitConversionToExplicit.ts
DomHenderson: This aws actually on purpose this time because this is a potential optimisation, I can remove it if you'd like though

src/passes/implicitConversionToExplicit.ts
DomHenderson: Yep, this one is the general case one, but I lfet the other one in for safety and for if anything gets introduced between the passes. As they only gen if necessary, it shouldn't ever lead to a double conversion

src/passes/implicitConversionToExplicit.ts
GuyMcComb: Nope no worries.

src/passes/implicitConversionToExplicit.ts
GuyMcComb: OK cool cool

GuyMcComb (approved):

PR #428, opened 2022-06-02
Exclude WARP FREE from compilation

Glitch18:
Excludes the WARP FREE functions from compilations.
Updates the testing.ts expected outcomes.

### DIFF
--- a/src/testing.ts
+++ b/src/testing.ts
@@ -289,6 +289,7 @@ function runCairoFileTest(
   onlyResults: boolean,
   throwError = false,
 ): void {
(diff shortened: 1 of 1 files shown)


JorikSchellekens:
While this is a tempting fix this is not the solution to the problem faced here

JorikSchellekens:
@DomHenderson pointed out that this works perfectly fine since the functions would be imported by the standard contracts.

JorikSchellekens (approved):

Glitch18:
Thanks for the review!

PR #433, opened 2022-06-02
Calldata dynamic array.length implemented

yaziciahmet:
Since `array.length` is by default uint256 in Solidity, it is enforced that any operation done with the `length` member must be also uint256. Hence anytime the `length` member is seen on a dynamic calldata array, it is wrapped with a `felt_to_uint256` function.


src/passes/arrayLenFunctionalizer.ts
```diff
@@ -0,0 +1,34 @@
+import { FunctionStateMutability, generalizeType, getNodeType, MemberAccess } from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { FunctionStubKind } from '../ast/cairoNodes';
+import { ASTMapper } from '../ast/mapper';
+import { createCairoFunctionStub, createCallToFunction } from '../utils/functionGeneration';
+import { isDynamicCallDataArray } from '../utils/nodeTypeProcessing';
+import { typeNameFromTypeNode } from '../utils/utils';
+
+export class ArrayLenFunctionalizer extends ASTMapper {
```
DomHenderson, line 9: I think this should be in the References subpass ArrayFunctions unless there's a good reason to keep it separate

DomHenderson (changes requested):

src/passes/references/arrayFunctions.ts
```diff
@@ -58,30 +63,49 @@ export class ArrayFunctions extends ReferenceSubPass {
...
+        const funcCall = createCallToFunction(funcStub, [node], ast);
+
+        ast.registerImport(funcCall, 'warplib.maths.utils', 'felt_to_uint256');
+        ast.replaceNode(node, funcCall, parent);
+        node.memberName = 'len';
+      } else if (
+        baseType.location === DataLocation.Storage ||
+        baseType.location === DataLocation.Memory
+      ) {
+        if (baseType.to.size !== undefined) {
```
DomHenderson, line 90: This branch should apply to all arrays, not just memory and storage arrays

src/passes/references/arrayFunctions.ts
```diff
@@ -58,30 +63,49 @@ export class ArrayFunctions extends ReferenceSubPass {
...
-            ? ast.getUtilFuncGen(node).storage.dynArrayLength.gen(node, baseType.to)
-            : ast.getUtilFuncGen(node).memory.dynArrayLength.gen(node, ast);
-        // The length function returns the actual length rather than a storage pointer to it,
-        // so the new actual location is Default
-        this.replace(node, replacement, undefined, DataLocation.Default, expectedLoc, ast);
+
+        const funcCall = createCallToFunction(funcStub, [node], ast);
+
+        ast.registerImport(funcCall, 'warplib.maths.utils', 'felt_to_uint256');
+        ast.replaceNode(node, funcCall, parent);
```
DomHenderson, line 84: The reference subpasses use `this.replace` rather than `ast.replace` as it allows them to set the actual and expected location of the inserted node. For this, the actual should be default, and the expected should be the expected of the MemberAccess. Similarly the expected location of the member access should be set to Default

tests/behaviour/contracts/array_len/calldataArray.sol
```diff
@@ -0,0 +1,30 @@
...
+        return x;
+    }
+
+    function fnCallWithArrLength(uint128[] calldata arr) public pure returns (uint256) {
+        return add1(arr.length);
+    }
+
+    function fnCallArrLengthNestedCalls(uint128[] calldata arr) public pure returns (uint256) {
+        return add1(arr.length) + mul2(arr.length) + add1(mul2(arr.length) + mul2(arr.length));
+    }
```
DomHenderson, line 29: Cases such as assigning the length of a calldata array to a storage variable would be good to include here, as well as static calldata arrays rather than just dynamic ones

DomHenderson (changes requested):

DomHenderson (approved):

PR #435, opened 2022-06-02
Add support for ecrecover precompile

swapnilraj:


.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         run: |
           yarn
           pipx install black[colors]
-          pipx install cairo-lang==0.8.2
+          pipx install cairo-lang==0.8.2.1
```
DomHenderson, line 27: The cairo lang version appears twice in this file, would be best to keep them in sync

src/utils/nodeTemplates.ts
```diff
@@ -207,6 +216,15 @@ export function createUint8TypeName(ast: AST): ElementaryTypeName {
   return typeName;
 }
 
+// prettier-ignore
+type UintN = 0 | 8 | 16 | 24 | 32 | 40 | 48 | 56 | 64 | 72 | 80 | 88 | 96 | 104 | 112 | 120 | 128 | 136 | 144 | 152 | 160 | 168 | 176 | 184 | 192 | 200 | 208 | 216 | 224 | 232 | 240 | 248 | 256;
```
DomHenderson, line 220: If you're going to specify all the widths like this, I don't think 0 should be included

DomHenderson (changes requested):
This looks great, just two nitpicks

DomHenderson (approved):

PR #436, opened 2022-06-02
reject try/catch & index param

RohitRanjangit:


src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -86,6 +89,13 @@ export class RejectUnsupportedFeatures extends ASTMapper {
     this.visitExpression(node, ast);
   }
 
+  visitParameterList(node: ParameterList, ast: AST): void {
+    // any of node.vParameters has indexed flag true then throw error
+    if (node.vParameters.some((param) => param.indexed)) {
+      throw new WillNotSupportError(`Indexed parameters are not supported`);
+    }
+  }
```
DomHenderson, line 97: Either this function needs to recurse, or the check should go in visitVariableDeclaration

DomHenderson (changes requested):

DomHenderson (changes requested):
Turns out we had indexed parameters that were simply being treated as normal parameters in various tests. These need to be replaced with normal parameters before this can be merged

DomHenderson (approved):

PR #440, opened 2022-06-03
Add byte and string concatenation

rodrigo-pino:


JorikSchellekens (approved):
This is fantastic. Just a style nit and a small change around error handling.

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -0,0 +1,147 @@
...
+      const exprType = getNodeType(expr, this.ast.compilerVersion);
+      if (!(exprType instanceof PointerType))
+        throw new NotSupportedYetError(
+          `Concatenation of expression ${printNode(expr)} of type ${printTypeNode(
+            exprType,
+          )} not supported yet`,
+        );
+      if (exprType.location !== DataLocation.Memory)
+        throw new TranspileFailedError(
+          `Concatenation of '${exprType.location}' different than memory`,
```
DomHenderson, line 24: Since DataAccessFunctionaliser runs after this I don't think the location should be specified here. Instead we should rely on the appropriate copy from storage or calldata getting inserted if necessary. Could you also then add these cases to the tests

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -0,0 +1,147 @@
...
+    return { name: funcName, code: code };
+  }
+
+  private generateFeltCopy() {
+    const func = this.generatedFunctions.get('felt_copy');
+    if (func !== undefined) return func.name;
+
+    const funcName = 'felt_copy';
+    const code = [
+      `func ${funcName}${IMPLICITS}(
```
DomHenderson, line 122: Since this is always the same, can it go in warplib rather than being generated and inserted into the contract file

DomHenderson (changes requested):
Looks really good, just a little bit that needs doing and then should be good to go

DomHenderson (approved):

PR #441, opened 2022-06-04
Support free functions in a library, stop exposing library functions as external, nuke free file compilation errors

JorikSchellekens:
Supports free functions in a library by removing a nasty assert
Prevents library functions from being leaked as external
Resolves all `__WARP_FREE__` compile issues by eliminating function definitions defined in the free source unit


GuyMcComb:
There is a situation where free functions are accessing/manipulating storage references. Because of the way that the storage cairoUtilGens work the free contract will not contain the storage_vars the cairoUtilGens need to functions. So once the free functions are inlined into the needed contracts, can we delete them?

JorikSchellekens:
The free functions are inlined and deleted in this pr

src/cairoUtilFuncGen/base.ts
```diff
@@ -61,20 +58,7 @@ export abstract class CairoUtilFuncGenBase {
...
-      const typeDefSourceUnit = type.definition.root;
-      assert(
-        typeDefSourceUnit instanceof SourceUnit,
-        `Unable to find SourceUnit holding type definition ${printTypeNode(type)}.`,
-      );
-      if (this.sourceUnit !== typeDefSourceUnit) {
-        this.requireImport(formatPath(typeDefSourceUnit.absolutePath), type.definition.name);
-      }
-    } else if (type instanceof IntType && type.nBits === 256) {
+    if (type instanceof IntType && type.nBits === 256) {
```
swapnilraj, line 61: let unify this with cairo util importer

JorikSchellekens:
I think we'll need further prs to sort out edge cases between interfaces and struct definitions

DomHenderson (approved):

PR #445, opened 2022-06-06
Concat fixed bytes

rodrigo-pino:
In solidity you can concatenate bytesX as well

### DIFF
--- a/src/cairoUtilFuncGen/memory/arrayConcat.ts
+++ b/src/cairoUtilFuncGen/memory/arrayConcat.ts
@@ -1,23 +1,35 @@
-import { DataLocation, FunctionCall, getNodeType, PointerType, TypeName } from 'solc-typed-ast';
+import assert from 'assert';
+import {
+  ArrayType,
+  DataLocation,
+  FunctionCall,
+  getNodeType,
+  IntType,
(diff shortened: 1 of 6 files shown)


src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -32,39 +44,58 @@ export class MemoryArrayConcat extends StringIndexedFuncGen {
...
   }
 
-  private genearteBytesConcat(argAmount: number): CairoFunction {
-    const funcName = `concat_${argAmount}`;
+  private genearteBytesConcat(argTypes: TypeNode[], implicits: string): CairoFunction {
+    const argAmount = argTypes.length;
+    const funcName = `concat${this.generatedFunctions.size}_${argAmount}`;
 
     if (argAmount === 0) {
       this.requireImport('warplib.memory', 'wm_new');
```
DomHenderson, line 94: This needs to import Uint256 too

DomHenderson (changes requested):
Looks great, just one import issue. The only other things not working atm with this is literals as arguments, but I think that's outside the scope of this PR

DomHenderson (approved):

PR #448, opened 2022-06-06
Reject nested bytes types as appropriate and support dynarrays of static arrays

DomHenderson:


src/testing.ts
```diff
@@ -80,31 +80,33 @@ const expectedResults = new Map<string, ResultType>(
...
-    ['example_contracts/functionInputs/arrayTest/arrayStructDynArray', 'NotSupportedYet'],
+    ['example_contracts/functionInputs/arrayTest/arrayStructDynArray', 'WillNotSupport'],
     ['example_contracts/functionInputs/arrayTest/arrayStructStruct', 'Success'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayArray', 'NotSupportedYet'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayDynArray', 'NotSupportedYet'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayStruct', 'NotSupportedYet'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayDynArrayArray', 'NotSupportedYet'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayDynArrayDynArray', 'NotSupportedYet'],
-    ['example_contracts/functionInputs/dynArrayTest/dynArrayDynArrayStruct', 'NotSupportedYet'],
+    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayArray', 'Success'],
```
GuyMcComb, line 92: Array Child of DynArray here. Should not be supported since no pointer to tuple type. Unless the contract is wrong? The input to the contract is `uint[3][3][] memory y`

src/testing.ts
```diff
@@ -80,31 +80,33 @@ const expectedResults = new Map<string, ResultType>(
...
+    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayDynArray', 'WillNotSupport'],
+    ['example_contracts/functionInputs/dynArrayTest/dynArrayArrayStruct', 'Success'],
```
GuyMcComb, line 94: Same as the above with this one `S[3][] memory y`.

GuyMcComb (changes requested):

src/testing.ts
GuyMcComb: Looks like pointers to tuples are supported.

GuyMcComb (approved):

GuyMcComb (changes requested):
Requesting changest to block merge until DynArray return test is checked.

GuyMcComb (approved):

GuyMcComb (changes requested):
Waiting for DynArray of Array returns to be added.

GuyMcComb (approved):

PR #452, opened 2022-06-06
Fix mod implicits

DomHenderson:
The warplib mod function requires bitwise pointer when dealing with signed values narrower than 256 bits, all other versions only require range_check. This progresses the viaYul/detect_mod_signed test, though it still requires the div/mod by 0 not throwing task to succeed fully

### DIFF
--- a/src/warplib/implementations/maths/mod.ts
+++ b/src/warplib/implementations/maths/mod.ts
@@ -1,5 +1,6 @@
 import { BinaryOperation } from 'solc-typed-ast';
 import { AST } from '../../../ast/ast';
+import { Implicits } from '../../../utils/implicits';
 import { mapRange } from '../../../utils/utils';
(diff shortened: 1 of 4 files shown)


RohitRanjangit (dismissed):

RohitRanjangit:
Would you please change the variable/argument type to signed from unsigned in`tests/behaviour/contracts/maths/remainder.sol` for signed functions?? They were meant to be signed type: https://github.com/NethermindEth/warp/blob/89001f82061368b6d313c51443f31d92ab4c16a7/tests/behaviour/contracts/maths/remainder.sol#L26

RohitRanjangit:
Would you please change the type of parameters (from usigned to signed) in `tests/behaviour/contracts/maths/remainder.sol` for signed functions. https://github.com/NethermindEth/warp/blob/89001f82061368b6d313c51443f31d92ab4c16a7/tests/behaviour/contracts/maths/remainder.sol#L26

PR #455, opened 2022-06-07
Add support for keccak

swapnilraj:
Depends on #435


example_contracts/precompiles/keccak.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.6;
+
+contract WARP {
+
+  function test(bytes memory m) public view {
+
```
DomHenderson, line 6: This seems to be missing the keccak

DomHenderson (changes requested):

example_contracts/precompiles/keccak.sol
swapnilraj: oops adding that and rebasing now

PR #457, opened 2022-06-07
Read id for memory

rodrigo-pino:
Adding `wm_read_id` function to use instead of `wm_read_felt` to read reference type pointers. In case there is no pointer initialized it creates a new one.


src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -98,15 +99,17 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
         `    alloc_locals`,
         ...structDef.vMembers.map((decl, index) => {
           const memberType = getNodeType(decl, this.ast.compilerVersion);
-          if (
-            (memberType instanceof UserDefinedType &&
-              memberType.definition instanceof StructDefinition) ||
-            memberType instanceof ArrayType
-          ) {
+          const allocSize =
+            memberType instanceof ArrayType && memberType === undefined
```
DomHenderson, line 103: Will need changing post bytesFix

DomHenderson (changes requested):
Just a few changes requested based on what's actually in the branch, and then I just left comments as reminders for all the things introduced that will need to be adjusted after the merge

DomHenderson (approved):

PR #462, opened 2022-06-08
Add different size static array copy from storage to storage

rodrigo-pino:


DomHenderson (changes requested):
I don't believe this covers all necessary cases. From testing in remix, `a=b` where `a` is a long array and `b` is a short array assigns the early elements of a the values stored in b, and clears the remaining elements. Worth creating some examples with both simple arrays and nested ones with local references to the elements

PR #466, opened 2022-06-08
External same name

DomHenderson:
Inserts and uses an interface for the current contract in the presence of external calls to an object of the same type


src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -80,12 +80,8 @@ function importExternalContract(
...
 ) {
-  const contractSourceUnit = contract.getClosestParentByType(SourceUnit);
-
   assert(sourceUnit !== undefined, 'Trying to import a definition into an unknown source unit');
 
-  if (contractSourceUnit === undefined || sourceUnit === contractSourceUnit) return;
-
   if (contract.kind === ContractKind.Library) return;
   if (contractInterfaces.has(contract.id)) return;
   contractInterfaces.set(contract.id, genContractInterface(contract, sourceUnit, ast));
```
swapnilraj, line 87: unrelated to your changes but on line 115 we only filter constructors from the interface, we should also filter not externally visible functions

tests/behaviour/contracts/cross_contract_calls/other_contract_same_type.sol
```diff
@@ -0,0 +1,27 @@
+pragma solidity ^0.8.10;
+
+// SPDX-License-Identifier: MIT
+
+contract WARP {
+  uint public counter;
+	function getAndIncrementCounter() external returns (uint) { return counter++; }
+	function getAndIncrementOtherCounter(address addr) external returns (uint) {
```
swapnilraj, line 8: nit: indentation

swapnilraj (approved):

PR #476, opened 2022-06-09
Memory To Storage Bug fix

rodrigo-pino:
When copying a smaller memory array into a larger storage array, the elements with index greater than the smaller array of the large array must be deleted


DomHenderson (changes requested):
This looks really elegantly done, but I'd like a test in the behaviour test suite that covers the full case (an array of objects with a local reference to one that gets deleted by the assignment of the shorter array)

DomHenderson (approved):
This is amazing

PR #485, opened 2022-06-10
fix flask async installation

BitBaseBit:


GuyMcComb (approved):
Looks like it solved it. Tested on local.

PR #486, opened 2022-06-10
Split external function calls returning dynarrays properly

DomHenderson:
External function calls with return types like `bytes memory` need to be treated as calldata returns. This modifies extractToConstant to check for this case and change the location of the constant

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -97,6 +97,7 @@ import {
   mergeImports,
   mangleOwnContractInterface,
   primitiveTypeToCairo,
+  isExternalCall,
 } from './utils/utils';
 
 const INDENT = ' '.repeat(4);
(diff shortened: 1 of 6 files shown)


src/cairoWriter.ts
```diff
@@ -767,9 +767,7 @@ class IdentifierWriter extends CairoASTNodeWriter {
           FunctionVisibility.External &&
         node.getClosestParentByType(IndexAccess) === undefined &&
         node.getClosestParentByType(MemberAccess) === undefined) ||
-        (node.parent instanceof FunctionCall &&
-          node.parent.vReferencedDeclaration instanceof FunctionDefinition &&
-          node.parent.vReferencedDeclaration.visibility === FunctionVisibility.External))
+        (node.parent instanceof FunctionCall && isExternalCall(node.parent)))
```
GuyMcComb, line 770: So satisfying to see this go!

GuyMcComb (approved):

PR #487, opened 2022-06-10
Handle unitialised storage pointers

DomHenderson:
Previously default values for reference types and mappings went into memory. This switches it so that if the variable is a type that can't go in memory, then it is given a placeholder initialValue that matches its type


tests/behaviour/expectations/semantic_whitelist.ts
```diff
@@ -907,7 +907,6 @@ const tests: string[] = [
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/mapping_returns_in_library.sol', // STRETCH conditional
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/internal_library_function_bound_to_address_named_send_transfer.sol', // WILL NOT SUPPORT raw address calls
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/internal_library_function_bound_to_interface.sol', // STRETCH new
-    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_library_structs.sol', // nested dynarray inputs
```
GuyMcComb, line 910: Was this test moved to another file? What does this mean?

GuyMcComb (changes requested):

tests/behaviour/expectations/semantic_whitelist.ts
DomHenderson: It got duplicated in a previous PR, ended up with one entry disallowed and one allowed. Turns out it's a valid test so I removed the duplicate

GuyMcComb (approved):

PR #489, opened 2022-06-10
Interface excess

DomHenderson:
Prunes excess members from interfaces, handles interface name mangling when calling an external function in the same contract type, fixes the inheritance/inherited_function_calldata_memory semantic test, leaving all inheritance semantic tests passing


rodrigo-pino (approved):
All good to me

PR #490, opened 2022-06-10
CallData -> Storage array implicit conversion.

GuyMcComb:


src/cairoUtilFuncGen/calldata/implicitArrayConversion.ts
```diff
@@ -0,0 +1,503 @@
...
+          code = [
+            `    let (storage_loc${index}) = ${this.dynArrayIndexAccessGen.getOrCreate(
+              targetElmType,
+            )}(ref, Uint256(${index}, 0))`,
+            `    ${this.storageWriteGen.getOrCreate(
+              targetElmType,
+            )}(storage_loc${index}, arg[${index}])`,
+          ].join('\n');
+        } else if (targetElmType.signed) {
+          // TODO find util to convert index to Uint256
```
rodrigo-pino, line 326: in `warplib/utils` there is function `uint256` that fits your need here

src/cairoUtilFuncGen/calldata/implicitArrayConversion.ts
```diff
@@ -0,0 +1,503 @@
...
+        `    ${dynArrayLengthName}.write(ref, Uint256(source.len, 0))`,
+        `    ${loaderName}(ref, source.len, source.ptr, 0)`,
+        '    return ()',
+        'end',
+      ].join('\n');
+      this.requireImport(
+        'warplib.maths.int_conversions',
+        `warp_int${sourceElmType.nBits}_to_int${targetElmType.nBits}`,
+      );
+      this.requireImport('warplib.maths.utils', 'felt_to_uint256');
```
rodrigo-pino, line 472: Perhaps importing only one of this depending if you're treating signed or unsigned integers

src/cairoUtilFuncGen/calldata/implicitArrayConversion.ts
GuyMcComb: Yeah I was asking Dom about that last night hahaha. I forgot where to find it.

src/cairoUtilFuncGen/calldata/implicitArrayConversion.ts
GuyMcComb: Yeah good idea. I am thinking of a way to compress the code, but there are such subtle differences in the actual copy code :(

rodrigo-pino: I had a similar issue in memory to storage conversion, in my case conversion code between the target and the source element type remained the same across all conversion functions so I wrapped it in a function. Perhaps the same thing can be applied here, also it will make it easier later if it's needed to add new conversion cases  like fixed bytes.

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -425,13 +425,21 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
 
-function getBaseType(type: TypeNode): TypeNode {
+export function getBaseType(type: TypeNode): TypeNode {
   const deferencedType = generalizeType(type)[0];
   return deferencedType instanceof ArrayType
     ? getBaseType(deferencedType.elementT)
     : deferencedType;
 }
 
+export function getNestedNumber(type: TypeNode): string {
```
DomHenderson, line 435: This isn't used in this file anymore. Could you move it to the file where it is. Also I don't think it should be called something else given that it's not getting a number

DomHenderson (approved):
Only comment is a tidyness nitpick that can be done in a further PR

PR #492, opened 2022-06-10
Inline array to storage bug

yaziciahmet:
Dynamic / static conversions were not handled properly. New conditions for conversion are added, old ones are most likely going to be deleted but I kept them for now. Few small bugs are solved in memory conversion functions.


yaziciahmet:
Already solved

PR #493, opened 2022-06-11
Init storage returns

DomHenderson:
Initialises non-state var storage variable declarations with placeholder literals instead of an object of equivalent type in memory, which causes an error because there's no storage location to assign the object to


tests/behaviour/expectations/semantic_whitelist.ts
```diff
@@ -1105,9 +1105,9 @@ const tests: string[] = [
   ...[
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/bytes_and_strings.sol',
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/multiline_comments.sol',
-    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/multiline.sol',
-    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/structs.sol', // nested dynarray
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/structs.sol',
+    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/multiline.sol', // attempts to call non-existant function
+    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/smoke/structs.sol', // nested dynarray
```
rodrigo-pino, line 1110: Really small thing, but you have `structs.sol` test twice, just to point it out

tests/behaviour/contracts/returns/initialiseStorageReturns.sol
```diff
@@ -0,0 +1,11 @@
+contract WARP {
+	uint[] arr1;
+	uint[][] arr2;
+	function getStorageArrays() internal returns (uint[] storage ptr1, uint[][] storage ptr2) {
+		ptr1 = arr1;
+		ptr2 = arr2;
+	}
```
rodrigo-pino, line 7: `getStorageArrays` is not being used in any test

GuyMcComb (approved):

PR #496, opened 2022-06-11
Fix function call assignment bug from short circuiting

DomHenderson:
Fixes the recurring bug with assignments of the form `let x = f()` arising from handling of short circuiting operators


GuyMcComb (approved):

GuyMcComb (changes requested):
Waiting for CI

PR #498, opened 2022-06-11
Var decl gaps

DomHenderson:
Handles variable declaration statements with gaps on the lhs

### DIFF
--- a/src/cairoWriter.ts
+++ b/src/cairoWriter.ts
@@ -208,7 +208,7 @@ class VariableDeclarationStatementWriter extends CairoASTNodeWriter {
     const documentation = getDocumentation(node.documentation, writer);
     const declarations = node.assignments.flatMap((id) => {
       if (id === null) {
-        return ['__warp_gv'];
+        return [`__warp_gv${this.gapVarCounter++}`];
       }
       const declaration = node.vDeclarations.find((decl) => decl.id === id);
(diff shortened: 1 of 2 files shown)


src/cairoWriter.ts
```diff
@@ -206,10 +207,9 @@ class VariableDeclarationStatementWriter extends CairoASTNodeWriter {
 
     const documentation = getDocumentation(node.documentation, writer);
     const declarations = node.assignments.flatMap((id) => {
-      if (id === null)
-        throw new NotSupportedYetError(
-          `VariableDeclarationStatements with gaps not implemented yet`,
-        );
+      if (id === null) {
+        return [`__warp_gv${this.gapVarCounter++}`];
+      }
```
GuyMcComb, line 212: Out of interest, where would there be an empty variable declaration?

DomHenderson: Usually something like `(uint a,) = f()` where the person doesn't care about the second return from f

GuyMcComb: I thought these would have been handled by the tuple filler? or was that only for assignments and not vardecl statements?

DomHenderson: Yep, they aren't actually TupleExpressions, they're just VariableDeclarationStatements with more slots than children

GuyMcComb: HA ok cool.

GuyMcComb (approved):

PR #500, opened 2022-06-11
Mul fix

DomHenderson:
Fixes incorrect handling of overflow for mul_signed256


src/warplib/implementations/maths/mul.ts
```diff
@@ -118,17 +118,28 @@ export function mul_signed(): void {
...
           `    let (res_abs, overflow) = uint256_mul(lhs_abs, rhs_abs)`,
           `    assert overflow.low = 0`,
           `    assert overflow.high = 0`,
-          `    let (msb) = bitwise_and(res_abs.high, ${msb(128)})`,
-          `    assert msb = 0`,
-          `    let (res) = uint256_cond_neg(res_abs, (lhs_nn + rhs_nn) * (2 - lhs_nn - rhs_nn))`,
-          `    return (res)`,
+          `    let res_should_be_neg = lhs_nn + rhs_nn`,
+          `    if res_should_be_neg == 1:`,
+          `        let (in_range) = uint256_le(res_abs, Uint256(0,${msb(128)}))`,
```
GuyMcComb, line 134: Uint256(0,${msb(128)})) Should this not be Uint256(${msb(128}, msb(128)}. int256 is always a bit tricky.

DomHenderson: Nope, it should be (0,msb(128)). The actual goal value is msb(256), which has all the lower 128 bits (and the next 127) set to 0

GuyMcComb: OH yes!

src/warplib/implementations/maths/mul.ts
```diff
@@ -118,17 +118,28 @@ export function mul_signed(): void {
           `func warp_mul_signed256{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(`,
           `    lhs : Uint256, rhs : Uint256) -> (result : Uint256):`,
           `    alloc_locals`,
+          `    # 1 => lhs >= 0, 0 => lhs < 0`,
           `    let (lhs_nn) = uint256_signed_nn(lhs)`,
+          `    # 1 => rhs >= 0, 0 => rhs < 0`,
           `    let (local rhs_nn) = uint256_signed_nn(rhs)`,
```
GuyMcComb, line 124: strange local here for rhs_nn but not lhs_nn

GuyMcComb (approved):

src/warplib/implementations/maths/mul.ts
DomHenderson: Yeah, probably not strictly necessary anymore since the cairo compiler got upgraded, but it used to prevent rhs_nn getting revoked

PR #507, opened 2022-06-11
Add filter for Member Access in ref_lib pass

RohitRanjangit:
Now `tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/internal_library_function_bound_to_fixed_bytes.sol` is passing

PR #508, opened 2022-06-12
Fallback function

cicr99:
Fallback functions can be overriden and modified, so a new contract was added to test all these cases. Running these tests, issues related to special functions inheritance arose:
As well as constructors, fallback and receive functions have an empty name in the ast, so the following issues needed to be fixed:
- When creating a cloned private function for a function from a base contract, the new name is `{previous_name}_s{depth}`, which means that if there are a fallback and receive functions inside of a contract, two different functions with the same name would be created, which would cause a failed compilation of the resulting cairo contract. Plus, the new function kind was not reset and so it resulted in a fallback function with a non-empty name, which causes an `InsaneASTError`
- When solving names for visible functions, it assumes that there are no repeated names for functions in the same contract, so if there are at least 2 from fallback, receive and constructor functions, it throws an assertion error.


BitBaseBit (changes requested):
Not sure if its this PR or #507 or #509 but one of them makes about 500 semantic tests fail that were passing before

DomHenderson (approved):

PR #509, opened 2022-06-12
write 0 if typename exp : ExpressionStatement

RohitRanjangit:
Now `tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/library_enum_as_an_expression.sol` is passing


RohitRanjangit:
#526

PR #510, opened 2022-06-12
throw unsupported error: Free fns in exp stmt

RohitRanjangit:
Throw `WillNotSupportError` in `tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/library_stray_values.sol`


DomHenderson (commented):
If you're making it throw an error for a semantic test, please include an update to the whitelist in this PR

DomHenderson (approved):

PR #518, opened 2022-06-13
Import DynArray Structs.

GuyMcComb:
Looks like Struct Definitions are being imported correctly from FREE contracts now. So removed a redundant function that also checked that they were imported.

Unfortunately, it does not fix dynArray structs being imported since the node representing them is actually a function definition stub that just writes the struct def to the file.

This PR fixes part of `tests/behaviour/solidity/test/libsolidity/semanticTests/freeFunctions/storage_calldata_refs.sol`.
The above test will be fully fixed once this PR: https://github.com/NethermindEth/warp/pull/441 is merged.

### DIFF
--- a/e.sol
+++ b/e.sol
@@ -1,0 +1,9 @@
+pragma solidity ^0.8.10;
+
+//SPDX-License-Identifier: MIT
+
+contract WARP {
+    function ints(uint8[] calldata x) public {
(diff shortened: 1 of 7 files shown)


GuyMcComb:
Not fixed, example contracts failing

PR #523, opened 2022-06-13
Map string index hash

yaziciahmet:
Mappings with string keys are now possible. Strings are uint8[] arrays, hash_chain is applied to the arrays with pedersen hash. If the string empty, returns 0.


src/cairoUtilFuncGen/storage/mappingIndexAccess.ts
```diff
@@ -2,40 +2,70 @@ import assert from 'assert';
...
+    const indexCairoType = CairoType.fromSol(baseType.to.keyType, this.ast);
+    const valueCairoType = CairoType.fromSol(
+      nodeType,
+      this.ast,
+      TypeConversionContext.StorageAllocation,
     );
 
+    if (baseType.to.keyType instanceof PointerType) {
+      assert(
+        baseType.to.keyType.to instanceof StringType,
```
swapnilraj, line 50: can this also be bytes?

yaziciahmet: yes, it should be, added it as well

src/cairoWriter.ts
```diff
@@ -851,7 +852,8 @@ class IdentifierWriter extends CairoASTNodeWriter {
           FunctionVisibility.External &&
         node.getClosestParentByType(IndexAccess) === undefined &&
         node.getClosestParentByType(MemberAccess) === undefined) ||
-        (node.parent instanceof FunctionCall && isExternalCall(node.parent)))
+        (node.parent instanceof FunctionCall && isExternalCall(node.parent)) ||
+        (node.parent instanceof FunctionCall && isHashFunction(node.parent)))
```
DomHenderson, line 856: CairoFunctionDefinitions have an acceptsRawDArray property, that can be set during creation with createCairoFunctionStub. I think that should be used rather than this

src/cairoUtilFuncGen/storage/mappingIndexAccess.ts
```diff
@@ -60,8 +91,12 @@ export class MappingIndexAccessGen extends StringIndexedFuncGen {
       return existing.name;
     }
 
-    const funcName = `WS${this.generatedFunctions.size}_INDEX_${indexType.typeName}_to_${valueType.typeName}`;
-    const mappingName = `WARP_MAPPING${this.generatedFunctions.size}`;
+    const funcName = `WS${this.generatedFunctions.size - this.generatedHashFunctionNumber}_INDEX_${
+      indexType.typeName
+    }_to_${valueType.typeName}`;
+    const mappingName = `WARP_MAPPING${
+      this.generatedFunctions.size - this.generatedHashFunctionNumber
```
DomHenderson, line 98: Can you explain what this is for?

DomHenderson (changes requested):

src/cairoUtilFuncGen/storage/mappingIndexAccess.ts
yaziciahmet: I wanted to distinguish the numbers given as postfix to the mapping functions and hashing functions. So since it already has `WARP_MAPPING0...`, the hash functions were automatically labeled as `ws_string_hash1`. So I separated the numbers of mapping functions and hash functions

src/cairoWriter.ts
yaziciahmet: The problem with that is, that it generates variables that don't exist. Assume `str` is the calldata string as the input to function, `str_len` and `str_ptr` becomes its parameters. Then it combines these into `str_dstruct` as a struct. Then, when the acceptsRawDArray property is set to true, its parameters become `str_dstruct_len, str_dstruct_ptr` which basically does not exist.

src/cairoUtilFuncGen/storage/mappingIndexAccess.ts
```diff
@@ -1,41 +1,72 @@
...
+      );
+      const stringLoc = generalizeType(getNodeType(index, this.ast.compilerVersion))[1];
+      assert(stringLoc !== undefined);
+      const call = this.createStringHashFunction(node, stringLoc, indexCairoType);
+      index = call;
+    }
+
+    const name = this.getOrCreate(indexCairoType, valueCairoType);
+
     const functionStub = createCairoFunctionStub(
```
GuyMcComb, line 62: If you add true for the argument in createCarioFunctionStrub it will split the DynArray args correctly

GuyMcComb: This relates to your convo above with Dom

src/cairoWriter.ts
yaziciahmet: cleared the cairowriter and moved conditions to utils

DomHenderson (approved):

PR #526, opened 2022-06-13
Removes TypeNameType Nodes From Contracts. 

GuyMcComb:
PR removes free standing TypeNameType expressions  from blocks, tuple assignments & variable declaration statements with empty slots.

Fixes  `tests/behaviour/solidity/test/libsolidity/semanticTests/structs/lone_struct_array_type`


tests/behaviour/contracts/ElementaryTypeNames/type_name_type_removal.sol
```diff
@@ -0,0 +1,29 @@
...
+    uint8 c;
+
+    function simple() public returns (int8) {
+        uint;
+        return 3; 
+    }
+
+    function indexAccess() public returns (int8) {
+        s[7][];
+        return 3;
```
swapnilraj, line 16: maybe the tests should return the unpacked values instead of just literals, to make sure they are getting unpacked correctly

GuyMcComb: What we are trying to do here is remove the expression in line 15. We just want to see that it does not throw a transpilation error.

src/passes/typeNameRemover.ts
```diff
@@ -0,0 +1,68 @@
...
+import assert from 'assert';
+import { Expression, TupleExpression } from 'solc-typed-ast';
+import { notNull } from '../utils/typeConstructs';
+import { expressionHasSideEffects } from '../utils/utils';
+
+export class TypeNameRemover extends ASTMapper {
+  visitExpressionStatement(node: ExpressionStatement, ast: AST): void {
+    if (
+      (node.vExpression instanceof IndexAccess ||
+        node.vExpression instanceof ElementaryTypeNameExpression) &&
```
RohitRanjangit, line 20: There can be more expression besides IndexAccess and ElementaryTypeNameExpression, which you'd want to cover. 
e.g
```solidity
library D {
    enum A{a}
    struct B{
        A a;
    }
}
contract C {
    function f(bytes calldata _x) public pure{
        D.A;
        C;
        D;
        D.B;
    }
}
```

RohitRanjangit (changes requested):

src/passes/typeNameRemover.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const emptySlots = node.assignments
+      .map((value, index) => (value === null ? index : null))
+      .filter(notNull);
+    const typeNameTypeSlots = rhs.vOriginalComponents
+      .map((value, index) => (this.isTypeNameType(value, index, ast) ? index : null))
+      .filter(notNull);
+
+    node.assignments = lhs.filter(
+      (_value, index) => !emptySlots.includes(index) && !typeNameTypeSlots.includes(index),
```
DomHenderson, line 51: These conditions should be || not &&. As is this transforms `(uint a,) = (f(), f())` into `uint a = f()`

DomHenderson (approved):

src/passes/typeNameRemover.ts
RohitRanjangit: Could you handle one more case, if these type of expressions are being used inside TupleExpression for e.g `(C, D.A, 2)` should also be disregarded.

RohitRanjangit (approved):

RohitRanjangit (approved):

PR #531, opened 2022-06-13
Fix exp implicit type conversion

DomHenderson:
The exp operator has special rules for type conversions. It evaluates to the type of the lhs (meaning an implicit conversion to the result type should be added to the lhs in case it's a literal) and only asserts that the rhs is unsigned, with the rhs width not effecting the operation. As such versions of the warplib function had to be implemented for all combinations of felt/Uint256 on the lhs and rhs. This was tested against all semantic exp tests and they all now pass. This was also the last failure in the cleanup tests that wasn't because of the test suite abiencoder wrapping invalid data, so they now also all pass

### DIFF
--- a/src/passes/implicitConversionToExplicit.ts
+++ b/src/passes/implicitConversionToExplicit.ts
@@ -93,10 +93,7 @@ export class ImplicitConversionToExplicit extends ASTMapper {
 
       insertConversionIfNecessary(node.vLeftExpression, resultType, ast);
       if (rightNodeType instanceof IntLiteralType) {
-        console.log(node.vRightExpression.typeString);
         const bound = getLiteralValueBound(node.vRightExpression.typeString);
-        console.log(bound);
-        console.log(node.vRightExpression.typeString);
         insertConversionIfNecessary(
           node.vRightExpression,
           intTypeForLiteral(`int_const ${bound}`),
--- a/src/passes/implicitConversionToExplicit.ts
+++ b/src/passes/implicitConversionToExplicit.ts
@@ -88,7 +88,22 @@ export class ImplicitConversionToExplicit extends ASTMapper {
         );
       }
       return;
-    } else if (['**', '*', '/', '%', '+', '-', '&', '^', '|', '&&', '||'].includes(node.operator)) {
+    } else if (node.operator === '**') {
+      const rightNodeType = getNodeType(node.vRightExpression, ast.compilerVersion);
+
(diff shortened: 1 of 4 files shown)


src/passes/implicitConversionToExplicit.ts
```diff
@@ -88,7 +88,22 @@ export class ImplicitConversionToExplicit extends ASTMapper {
         );
       }
       return;
-    } else if (['**', '*', '/', '%', '+', '-', '&', '^', '|', '&&', '||'].includes(node.operator)) {
+    } else if (node.operator === '**') {
+      const rightNodeType = getNodeType(node.vRightExpression, ast.compilerVersion);
+
+      insertConversionIfNecessary(node.vLeftExpression, resultType, ast);
+      if (rightNodeType instanceof IntLiteralType) {
+        console.log(node.vRightExpression.typeString);
```
RohitRanjangit, line 96: Nitpick: remove debug statements

src/warplib/implementations/maths/exp.ts
```diff
@@ -35,18 +45,38 @@ function createExp(signed: boolean, unsafe: boolean) {
     forAllWidths((width) => {
       if (width === 256) {
         return [
-          `func _repeated_multiplication${width}{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(op : Uint256, count : Uint256) -> (res : Uint256):`,
+          `func _repeated_multiplication${width}{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(op : Uint256, count : felt) -> (res : Uint256):`,
+          `    if count == 0:`,
+          `        return (Uint256(1, 0))`,
+          `    end`,
+          `    let (x) = _repeated_multiplication${width}(op, count - 1)`,
```
RohitRanjangit, line 52: No. of steps can be significantly reduced if we use logarithmic complexity repeated multiplication instead of linear one. 
```cairo
let (y) = _repeated_multiplication${width}(op, count/2)
let (x) = y*y
if count is even : let res = x
else :  let res = warp_mul${suffix}${width}(op, x)

```

RohitRanjangit (changes requested):

src/warplib/implementations/maths/exp.ts
DomHenderson: This is definitely worth looking into post audit to see how much it does

RohitRanjangit (approved):

PR #535, opened 2022-06-14
Inline all constants

swapnilraj:
Fixes:
`tests/behaviour/solidity/test/libsolidity/semanticTests/constants/constant_string_at_file_level`


src/passes/constantHandler.ts
```diff
@@ -1,91 +1,30 @@
...
 
-    const block = node.vBody;
-    const blockId = block.id;
-
-    [...unboundConstantsToReplace.entries()].map(([decl, ids]) => {
-      if (
-        decl.vType === undefined ||
-        decl.vValue === undefined ||
-        !(decl.vValue instanceof Literal)
+  visitIdentifier(node: Identifier, ast: AST): void {
```
DomHenderson, line 14: This needs a visitMemberAccess method as well

DomHenderson (changes requested):

PR #538, opened 2022-06-14
Fix documentation bug in ConstantHandler

Glitch18:
While trying to transpile BaseJumpRateModelV2, the transpilation was running into an error where a `StructuredDocumentation` node was being assigned a wrong parent. 
This PR implments a fix for this by ensuring that a copy of the documentation node is given to the new `VariableDeclaration` node.

### DIFF
--- a/src/passes/constantHandler.ts
+++ b/src/passes/constantHandler.ts
@@ -5,12 +5,12 @@ import {
   Literal,
   Mutability,
   StateVariableVisibility,
-  StructuredDocumentation,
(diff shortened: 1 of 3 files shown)


DomHenderson (changes requested):
From a search through the codbase for `.documentation` this also occurs in DeleteHandler. Also this same code to safely clone a documentation property already exists in cloning.ts and can just be exported

DomHenderson (approved):

PR #539, opened 2022-06-14
Solve replacement issue when adding extra read

rodrigo-pino:
Error when replacing a node over a disconnected one, it could not find a a user defined type (a struct in this case).
Semantic test `copy_struct_array_from_storage.sol` pass now

### DIFF
--- a/src/cairoUtilFuncGen/storage/dynArrayPushWithoutArg.ts
+++ b/src/cairoUtilFuncGen/storage/dynArrayPushWithoutArg.ts
@@ -19,7 +19,7 @@ export class DynArrayPushWithoutArgGen extends StringIndexedFuncGen {
     super(ast, sourceUnit);
   }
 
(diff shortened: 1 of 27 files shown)


src/cairoUtilFuncGen/storage/dynArrayPushWithoutArg.ts
```diff
@@ -34,7 +27,7 @@ export class DynArrayPushWithoutArgGen extends StringIndexedFuncGen {
       [['newElemLoc', typeNameFromTypeNode(elementType, this.ast), DataLocation.Storage]],
       ['syscall_ptr', 'pedersen_ptr', 'range_check_ptr'],
       this.ast,
-      nodeInSourceUnit ?? push,
+      this.sourceUnit,
```
DomHenderson, line 30: Is there a reason for doing this rather than passing in a valid nodeInSourceUnit parameter?

DomHenderson (approved):

PR #540, opened 2022-06-14
Change ext fn def with it's typestring

rjnrohit:
In case of external lib/contract functions , it's typestring generated by solc-typed-ast was mismatching in case of `calldata` arguments. This PR should handle that.

for e.g
```solidity
library D {
    function f(bytes calldata _x) public pure returns (bytes calldata) {
        return _x;
    }
    function g(bytes calldata _x) public pure returns (bytes memory) {
        return _x;
    }
}

contract C {
    using D for bytes;
    function f(bytes calldata _x) public pure returns (bytes1, bytes1) {
        return (_x.f()[0], _x.g()[0]);
    }
}
```

Now, semantic tests:
`tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/bound_returning_calldata_external.sol` and
`tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/bound_to_calldata_external.sol` are passing.

PS: If this PR gets merged then it'll reduce no. of semantic tests fails by 8(failing  + pending) ;)

### DIFF
--- a/tests/behaviour/expectations/semantic_whitelist.ts
+++ b/tests/behaviour/expectations/semantic_whitelist.ts
@@ -901,7 +901,7 @@ const tests: string[] = [
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_for_overload.sol',
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_for_storage_structs.sol',
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_library_mappings_return.sol',
-    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_library_structs.sol', // Dynamic arrays are not allowed as (indirect) children of structs
+    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_library_structs.sol',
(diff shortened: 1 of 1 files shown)


src/passes/functionTypeStringMatcher/identifierTypeStringMatcher.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+  visitIdentifier(node: Identifier, _ast: AST): void {
+    if (
+      node.vReferencedDeclaration instanceof VariableDeclaration &&
+      this.declarations.has(node.vReferencedDeclaration)
+    ) {
+      const typeString = node.vReferencedDeclaration.typeString;
+      if (node.vReferencedDeclaration.storageLocation !== DataLocation.Default) {
+        node.typeString = `${typeString} ${node.vReferencedDeclaration.storageLocation}`;
+        console.log(node.id, node.vReferencedDeclaration.id, node.typeString);
```
swapnilraj, line 18: extra console log

swapnilraj (approved):
one nit but apart from that looks good!

src/passes/functionTypeStringMatcher/identifierTypeStringMatcher.ts
RohitRanjangit: shit!! will remove it

PR #542, opened 2022-06-14
Add input check in ecrecover and add semanti tests

swapnilraj:


yaziciahmet (commented):
Seems ok, will approve after CI.

PR #544, opened 2022-06-14
Do not run abiExtractor on abstract contracts

DomHenderson:
Abstract contracts do not need abis generated because they cannot be deployed, and because their constructors are only called internally they can take mappings, which cannot be abi encoded.


yaziciahmet (commented):
Looks good. Will approve it once it passes the CI.

PR #546, opened 2022-06-14
Throw Asserts when enums overflow.

GuyMcComb:


src/cairoUtilFuncGen/enumInputCheck.ts
```diff
@@ -0,0 +1,84 @@
...
+      name,
+      [['arg', typeNameFromTypeNode(inputType, this.ast), DataLocation.Default]],
+      [['ret', typeNameFromTypeNode(nodeType, this.ast), DataLocation.Storage]],
+      ['range_check_ptr'],
+      this.ast,
+      nodeInSourceUnit ?? nodeInput,
+      FunctionStateMutability.Pure,
+      FunctionStubKind.FunctionDefStub,
+    );
+    const functionInput = cloneASTNode(nodeInput, this.ast);
```
DomHenderson, line 41: Shouldn't be any need to clone here I don't think

src/cairoUtilFuncGen/enumInputCheck.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    this.ast.setContextRecursive(functionInput);
+    return createCallToFunction(functionStub, [functionInput], this.ast);
+  }
+
+  private getOrCreate(type: TypeNode, enumDef: EnumDefinition): string {
+    const key = type.pp();
```
DomHenderson, line 48: key needs to be more specific than this. Needs to include both the enum and the input type

src/cairoUtilFuncGen/enumInputCheck.ts
```diff
@@ -0,0 +1,84 @@
...
+  ): FunctionCall {
+    const nodeType = getNodeType(node, this.ast.compilerVersion);
+    const inputType = getNodeType(nodeInput, this.ast.compilerVersion);
+
+    this.sourceUnit = this.ast.getContainingRoot(nodeInSourceUnit);
+    const name = this.getOrCreate(inputType, enumDef);
+    const functionStub = createCairoFunctionStub(
+      name,
+      [['arg', typeNameFromTypeNode(inputType, this.ast), DataLocation.Default]],
+      [['ret', typeNameFromTypeNode(nodeType, this.ast), DataLocation.Storage]],
```
DomHenderson, line 34: Why storage?

DomHenderson (changes requested):
This is more what I meant yes

src/cairoUtilFuncGen/enumInputCheck.ts
GuyMcComb: Ahh yes good point.

src/cairoUtilFuncGen/enumInputCheck.ts
GuyMcComb: Because Enums are stored in storage aren't they?

DomHenderson: no more than ints are

DomHenderson (approved):

PR #549, opened 2022-06-14
Update whitelist

DomHenderson:


yaziciahmet (changes requested):
fallback_return, inherited are moved into behaviour tests. short_data_calls_fallback is not supported. #551

PR #559, opened 2022-06-15
Replace --strict with --no-strict

DomHenderson:


JorikSchellekens (changes requested):
Flag changed but implementation not updated

JorikSchellekens (approved):

PR #583, opened 2022-06-21
Bug fix when memory treated as calldata is returned as memory treated as calldata

rodrigo-pino:


src/utils/utils.ts
```diff
@@ -416,7 +416,21 @@ export function isExternalCall(node: FunctionCall): boolean {
   );
 }
 
-// 'string_hash' function can not be user defined, due to mangling identifiers
+export function isExternalMemoryDynArray(node: Identifier, compilerVersion: string): boolean {
```
DomHenderson, line 419: Could you combine this with isExpectingSplit and then add comments to explain the cases by which we do/don't split identifiers? I find it difficult to be convinved we're hitting all edge cases otherwise

DomHenderson (changes requested):

PR #585, opened 2022-06-22
Unreachable functions remover

cicr99:
To remove unreachable functions from the code:
- A call graph is created: For each `FunctionDefinition`, it obtains a list of ids of the functions being referenced in its `FunctionCalls`
- Those ids are replaced by the actual `FunctionDefinition`. Ids of external function calls are ignored.
- For each contract, externally visible functions are collected and a `dfs` is performed from them in the previous graph to get all reachable functions
- All unreachable functions are removed

### DIFF
--- a/src/passes/functionPruner/callGraph.ts
+++ b/src/passes/functionPruner/callGraph.ts
@@ -1,12 +1,5 @@
 import assert from 'assert';
-import {
-  Expression,
-  FunctionCall,
-  FunctionDefinition,
-  Identifier,
-  IdentifierPath,
-  MemberAccess,
-} from 'solc-typed-ast';
+import { FunctionCall, FunctionDefinition } from 'solc-typed-ast';
 import { AST } from '../../ast/ast';
 import { ASTMapper } from '../../ast/mapper';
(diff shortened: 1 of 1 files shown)


src/passes/functionPruner/callGraph.ts
```diff
@@ -0,0 +1,72 @@
...
+      this.currentFunction !== undefined,
+      `Function Call ${printNode(node)} outside FunctionDefinition`,
+    );
+    const existingCalls = this.callGraph.get(this.currentFunction.id);
+    assert(
+      existingCalls !== undefined,
+      `${printNode(this.currentFunction)} should have been added to the map`,
+    );
+
+    const refFuncId = getReferencedFunctionId(node.vExpression);
```
DomHenderson, line 45: FunctionCall already has a vReferencedDeclaration member, any reason for having a custom function rather than using that?

DomHenderson (approved):

DomHenderson (changes requested):
Just putting a hold on this one for a moment for merge ordering. Will approve later today

DomHenderson (approved):

PR #592, opened 2022-06-24
Fix index access functions expected locations

yaziciahmet:
IndexAccess nodes without any return or VarDeclStmt parents get an expected location of undefined, which then causes the read function for it to be not generated. Smh like
```
bool[] storage arr;
if(arr[0]) {...}
```
crashes. Because it converts arr[0]  into `WARP_DARRAY0_felt_IDX(arr,0)` while it should have `WS0_READ_felt(WARP_DARRAY0_felt_IDX(arr,0))` . Expected locations are set from parent to child, so I was not able to fix this inside expectedLocationAnalyzer. Instead, I fixed it while replacing the IndexAccess with a FunctionCall.


src/passes/references/dataAccessFunctionaliser.ts
```diff
@@ -309,7 +309,14 @@ export class DataAccessFunctionaliser extends ReferenceSubPass {
     }
 
     if (replacement !== null) {
-      this.replace(node, replacement, undefined, actualLoc, expectedLoc, ast);
+      this.replace(
+        node,
+        replacement,
+        undefined,
+        actualLoc,
+        expectedLoc ?? DataLocation.Default,
```
DomHenderson, line 317: The issue here is specific to ifStatements, not indexAccesses. An unused indexAccess should still have an undefined expected location, the problem was that expectedLocationAnalyser wasn't setting the expected location of the if condition to default. Could you modify expectedLocationAnalyser to do so rather than doing this

DomHenderson (changes requested):

src/passes/references/dataAccessFunctionaliser.ts
yaziciahmet: Change is done.

PR #595, opened 2022-06-25
Cleanup contract name mangling

JorikSchellekens:
Uses a slightly shorter `__WC__` infix and defines a single source of truth for it.

### DIFF
--- a/src/passes/sourceUnitSplitter.ts
+++ b/src/passes/sourceUnitSplitter.ts
@@ -13,6 +13,9 @@ import { cloneASTNode } from '../utils/cloning';
 
 type Scoped = FunctionDefinition | ContractDefinition | VariableDeclaration | StructDefinition;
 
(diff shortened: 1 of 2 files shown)


src/passes/sourceUnitSplitter.ts
```diff
@@ -13,6 +13,9 @@ import { cloneASTNode } from '../utils/cloning';
 
 type Scoped = FunctionDefinition | ContractDefinition | VariableDeclaration | StructDefinition;
 
+export const CONTRACT_INFIX = '__WC__';
+export const FREE_SUFIX = '__WARP_FREE__';
```
swapnilraj, line 17: Why not just __WC_FREE__?

JorikSchellekens: Actually yeah `__WC_FREE`

swapnilraj (approved):

PR #596, opened 2022-06-25
Transpile multiple files

JorikSchellekens:
Naive way to compile any number of sol files. Handy for transpiling whole projects.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,7 +43,7 @@ type CliOptions = CompilationOptions & TranspilationOptions & PrintOptions & Out
 const program = new Command();
 
 program
-  .command('transpile <file>')
+  .command('transpile <files...>')
   .option('--compile-cairo')
   .option('--no-compile-errors')
   .option('--check-trees')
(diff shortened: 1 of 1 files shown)


DomHenderson (commented):
My only concern with this is that if you're trying to transpile just a few files then doing them one by one is fine, and if you're trying to transpile more than that then you really need some sort of pattern matching, so I'm not sure how much use this would get. Any thoughts?

JorikSchellekens:
Pattern matching works out of the box with globbing

src/index.ts
```diff
@@ -57,15 +57,20 @@ program
...
-    if (!isValidSolFile(file)) return;
-    try {
-      transpile(compileSolFile(file, options.warnings), options).map(([name, cairo, abi]) => {
-        outputResult(name, cairo, options, '.cairo', abi);
-      });
-    } catch (e) {
-      handleTranspilationError(e);
-    }
+  .action((files: string[], options: CliOptions) => {
+    if (files.some((file) => !isValidSolFile(file))) return;
```
swapnilraj, line 61: Shouldn't we check if we every sol file is valid, if not we should print an error about the invalid ones

JorikSchellekens: As in give errors for all failing ones instead of fast failing as we do here?

swapnilraj (approved):

PR #598, opened 2022-06-27
Excess functions

cicr99:
The purpose of this pass is to find functions like the following and remove them as a way to optimize the code:
```
function foo(uint a) returns (uint) {
  return a;
}
```
If these functions are not externally visible, it is possible to remove them and replace the calls to that function with the expression it returns, tidying up the references of course.
This pass also recognizes functions of this type that return more than one argument in the form of a `TupleExpression`, even if the identifiers are not in the same order as the function parameters.


tests/behaviour/expectations/behaviour.ts
```diff
@@ -3792,6 +3792,9 @@ export const expectations = flatten(
           Expect.Simple('test', [], []),
           Expect.Simple('returnTest', [], ['12', '0']),
         ]),
+        File.Simple('removeIdentityFunctions', [
+          Expect.Simple('f', ['10', '0'], ['3', '20', '0', '5', '0', '13', '0']),
+        ]),
```
rodrigo-pino, line 3797: Is there a way of testing that function `h` and `g` got pruned correctly. Like calling them and catching the error because they do not exist

rodrigo-pino (approved):
It looks good to me

tests/behaviour/expectations/behaviour.ts
cicr99: The thing is for those functions to be pruned they most not be externally visible, so even if there was no pruning at all, calling them should always throw an error. In other words, I think that trying to call those functions won't work as a way of testing this pass.

rodrigo-pino: Right, that's true

src/passes/functionPruner/index.ts
```diff
@@ -3,13 +3,7 @@ import { ASTMapper } from '../../ast/mapper';
 import { CallGraphBuilder } from './callGraph';
 import { FunctionRemover } from './functionRemover';
 
-export class UnreachableFunctionPruner extends ASTMapper {
-  // Function to add passes that should have been run before this pass
```
GuyMcComb, line 7: Why was this deleted?

src/passes/functionPruner/index.ts
```diff
@@ -3,13 +3,7 @@ import { ASTMapper } from '../../ast/mapper';
...
-  addInitialPassPrerequisites(): void {
```
GuyMcComb, line 8: Why was this removed?

GuyMcComb:
Sem tests running

src/passes/functionPruner/index.ts
cicr99: Oh I didn't realize it got deleted. It must have been when I had to fix some conflicts. I'll add it

GuyMcComb:
Sem test pass. Just waiting for deleted block to be added back

src/passes/identityFunctionRemover.ts
```diff
@@ -0,0 +1,125 @@
...
+  Identifier,
+  Return,
+  TupleExpression,
+} from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+import { printNode } from '../utils/astPrinter';
+import { isExternallyVisible, toSingleExpression } from '../utils/utils';
+
+export class IdentityFunctionRemover extends ASTMapper {
```
GuyMcComb, line 15: Looks good, just needs some documentation describing this pass.

src/passes/functionPruner/index.ts
GuyMcComb: Might be a good opportunity to add some documentation to this pass as well.

GuyMcComb (changes requested):

GuyMcComb:
Sem Tests passed with new commits

GuyMcComb (approved):

PR #599, opened 2022-06-27
Unifiy mangling + some maintenance

rodrigo-pino:
Moved all prefix/infix/suffix of generated/modified variable names to a single source file. This way is easier to notice name clashes and modify them all at once if needed.

Also verified it is not possible for name clash to occur:

- every user defined variable during the `IdentifierMangler` pass gets the prefix `__warp_usrId{i}` where `i` is different for each declaration.
- every generated variable use different infix/suffix/prefix
- even if there is some variable creation before the `IdentifierMangler` pass, and if adversarial solidity code is written, it is possible that user defined variables will have same name as generated one, but since they are different declarations and a reference system is used internally no clash will ocurr during this time. Also when `IdentifierMangler` is reach it will detect them as different variables and assigned them different prefixes (`__warp_usrId_i` and `__warp_usrId_j` where `i != j`) resolving the name clash when they get transpiled to Cairo.

Finally, updated some comments


src/utils/nameModifiers.ts
```diff
@@ -0,0 +1,45 @@
...
+// Update script name
+// ExternalArgModifier/MemoryRefInputModifier has a _mem suffix
+// If Functionalizer has a '_part' in between
+// The one with static index access to memory, clash is bound to happen
+
+// Used in TupleFiller in TupleFixes
+export const TUPLE_FILLER_PREFIX = '__warp_tf';
+
+// Used in SourceUnitSplitter
+export const FREE_FILE = '__WARP_FREE__';
```
JorikSchellekens, line 13: Can you update these to those specified in #595?

swapnilraj (approved):
Looks good!

PR #600, opened 2022-06-28
Use absolute path for logo

JorikSchellekens:
The logo is broken on npmjs.com because it tries to resolve the image relative to the url.
While this solution is not ideal for development the logo is unlikely to change in the near
future so I see no harm in setting this absolute path.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,4 @@
-<img src="./resources/WARP.svg" width="900" height="512" />
+<img src="https://raw.githubusercontent.com/NethermindEth/warp/main/resources/WARP.svg" width="900" height="512" />
 
 # Warp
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -1,4 +1,4 @@
-<img src="./resources/WARP.svg" width="900" height="512" />
+<img src="https://raw.githubusercontent.com/NethermindEth/warp/main/resources/WARP.svg./resources/WARP.svg" width="900" height="512" />
```
swapnilraj, line 1: the old path is still there

swapnilraj: is it meant to be like that?

axic:
> The logo is broken on npmjs.com because it tries to resolve the image relative to the url.

Someone should fix npmjs really 😬

readme.md
JorikSchellekens: no, undo caught me out

JorikSchellekens:
@axic seems like this is still a problem even with that suggestion:
https://github.com/npm/marky-markdown/pull/435#issuecomment-670591742
https://github.com/npm/marky-markdown/issues/449#issuecomment-670601860

swapnilraj (approved):

PR #602, opened 2022-06-29
Cairo v0.9

GuyMcComb:
Upgrading Warplib and testnet to support Cairo v0.9. 

Changes:

Removing function calls in return statements in Warplib.
Changing ContractDefintion to ContractClass in testnet wrapper.

Passed all behavior tests and semantic tests run on local.


swapnilraj:
I think you'll need to update the ci yml file as well

swapnilraj:
There is also a reference to cairo lang version in warp_venv.sh

swapnilraj (approved):

PR #606, opened 2022-07-11
Contract Factory Support

rodrigo-pino:
Transpile Solidity's `new Contract` expressions to Cairo's `deploy` system calls. 

- [x] Encode constructor arguments as raw felt calldata
- [x] Define a placeholder for class hash
- [x] Fill placeholder during behaviour tests
- [x] Fill placeholder during semantic tests
- [x] Fill placeholder during common warp usage
- [x] Allow using salt option for contract creation (only)
- [x] Behavior tests with new
- [x] Semantic tests with new
- [x] Gather and report unrelated bugs found during development:
       - Bug on `tests/behaviour/contracts/new/product.sol` when calling `getX` from arrays
       - Bug when `getX` returns a struct defined globally. Example in `tests/behaviour/conracts/new/many.sol`


src/transpiler.ts
```diff
@@ -62,6 +62,7 @@ import { createPassMap, parsePassOrder } from './utils/cliOptionParsing';
 import { TranspilationAbandonedError, TranspileFailedError } from './utils/errors';
 import { error, removeExcessNewlines } from './utils/formatting';
 import { printCompileErrors, runSanityCheck } from './utils/utils';
+import { NewToDeploy } from './passes/newToDeploy';
```
GuyMcComb, line 65: This can be moved to passes

src/passes/newToDeploy.ts
```diff
@@ -0,0 +1,215 @@
...
+      '',
+      true,
+      false,
+      varName,
+      sourceUnit.id,
+      false,
+      DataLocation.Default,
+      StateVariableVisibility.Internal,
+      Mutability.Constant,
+      'uint160',
```
GuyMcComb, line 206: @DomHenderson How do we feel about this being 160bits?

src/passes/implicitConversionToExplicit.ts
```diff
@@ -203,6 +203,9 @@ export class ImplicitConversionToExplicit extends ASTMapper {
         handleConcatArgs(node, ast);
         return;
       }
+      if (node.vFunctionName === 'encode') {
+        return;
+      }
```
rodrigo-pino, line 208: This is a left over. Marking it here for when adressing changes

src/passes/newToDeploy.ts
rodrigo-pino: I marked it as 160 bits thought in Cairo it can be bigger than that, it would never be an Uint256 (< 251 bits). Also it has the same size as an address. Perhaps making it an address type instead :thinking:

tests/behaviour/contracts/new/many.sol
```diff
@@ -0,0 +1,108 @@
...
+    function createUint256Contract(uint x_) public {
+        u256 = new Uint256Contract(x_);
+    }
+
+    function createUint8Contract(uint8 x_) public {
+        u8 = new Uint8Contract(x_);
+    }
+
+    function createDynArrayContract(uint[] memory x_) public {
+        dac = new DynArrayContract(x_);
```
GuyMcComb, line 74: Does the encode_to_felt work for calldata?

rodrigo-pino: It deals with any type. If not pass the right storage type a conversion is triggered.

GuyMcComb: Great!

src/passes/newToDeploy.ts
DomHenderson: Address might make the most sense I think. Not keen on it being 160 since with nethersolc there's no tie between the class hash and 160bits

tests/behaviour/expectations/semantic_whitelist.ts
```diff
@@ -1607,7 +1607,7 @@ const tests: string[] = [
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/constructor_with_params_inheritance_2.sol',
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/empty_contract.sol', // test suite cannot call non-existant function
     // 'tests/behaviour/solidity/test/libsolidity/semanticTests/constructor_inheritance_init_order_3_legacy.sol', // We don't support the legacy bug behaviour and instead implement viaYul behaviour.
-    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/interface_inheritance_conversions.sol', // REQUIRES CAIRO 0.9 uses new
+    // 'tests/behaviour/solidity/test/libsolidity/semanticTests/interface_inheritance_conversions.sol', // Address clash when deploying the same contracts
```
DomHenderson, line 1610: I'm a little uncomfortable with how many failing tests there are here for a PR up for merging (though this may also be because I find it hard to tell from these error messages which are expected failures (WILL NOT SUPPORT) and which are unexpected). Ultimately though, I'll leave it up to you and @GuyMcComb as to how many if any need fixing before this PR is merged

src/cairoUtilFuncGen/utils/encodeToFelt.ts
```diff
@@ -0,0 +1,283 @@
...
+    const resultStruct = this.externalArrayGen.getOrCreate(
+      typeNameToTypeNode(createBytesTypeName(this.ast)),
+    );
+
+    const cairoParams = parameters.join(',');
+    const funcName = `encode_as_felt${this.generatedFunctions.size}`;
+    const code = [
+      `func ${funcName}${IMPLICITS}(${cairoParams}) -> (calldata_array : ${resultStruct}):`,
+      `   alloc_locals`,
+      `   let total_size : felt = ${0}`,
```
DomHenderson, line 147: is there a reason for `${0}` rather than just `0` here?

src/cairoUtilFuncGen/utils/encodeToFelt.ts
```diff
@@ -0,0 +1,283 @@
...
+  isValueType,
+} from '../../utils/nodeTypeProcessing';
+import { mapRange, narrowBigIntSafe, typeNameFromTypeNode } from '../../utils/utils';
+import { CairoFunction, delegateBasedOnType, StringIndexedFuncGen } from '../base';
+import { ExternalDynArrayStructConstructor } from '../calldata/externalDynArray/externalDynArrayStructConstructor';
+
+const IMPLICITS = '';
+
+// TODO:
+// What happens with sructs defined in other contract different than the one using new
```
DomHenderson, line 40: Is it worth adding a test for this before merging?

DomHenderson (commented):
Looks good overall. The things I've raised are all fairly minor. The one overall suggestion I have is that this (much like most complex features in warp tbh) would benefit from the dense chunks of code being refactored into named and commented functions

tests/behaviour/expectations/semantic_whitelist.ts
GuyMcComb: Looks like all the tests that fail because of 'value option' should be changed to will not support? And I assume we have created bug tickets for "Fails because it WARP allows a contract to call itself during creation (And it shoudl not)"

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -157,10 +168,14 @@ export class RejectUnsupportedFeatures extends ASTMapper {
   visitNewExpression(node: NewExpression, ast: AST): void {
     if (
       !(node.vTypeName instanceof ArrayTypeName) &&
-      !(node.vTypeName instanceof ElementaryTypeName && node.vTypeName.name === 'bytes')
+      !(node.vTypeName instanceof ElementaryTypeName && node.vTypeName.name === 'bytes') &&
+      !(
+        node.vTypeName instanceof UserDefinedTypeName &&
+        node.vTypeName.vReferencedDeclaration instanceof ContractDefinition
```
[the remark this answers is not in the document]
rodrigo-pino: Yep, I think we cover them all.
There is no reference type that gets constructed with new other than arrays (& bytes). Neither mappings or structs.
And value types does not get constructed with new, except perhaps a contract type which is an address

src/passes/newToDeploy.ts
GuyMcComb: Could we have this as an Address and then we can confirm in the meeting on Wednesday

starknet-testnet/server.py
```diff
@@ -78,6 +79,39 @@ async def deploy():
...
+    state = await starknet_wrapper.get_state()
+    compiled_cairo = open(data["compiled_cairo"]).read()
+    contract_class = ContractClass.loads(compiled_cairo)
+
+    try:
+        execution_info = await state.declare(contract_class)
+        print("-----Declare info-----")
+        print(execution_info)
+        print("-----------\n")
+        # BENCHMARK needed here?
```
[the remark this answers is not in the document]
rodrigo-pino: Ok, I would prefer leaving it for another PR

tests/behaviour/expectations/semantic_whitelist.ts
rodrigo-pino: Modified those to `WILL NOT SUPPORT due to reason` to be more clear

src/cairoUtilFuncGen/utils/encodeToFelt.ts
rodrigo-pino: Yep, it is bugged when it's a global struct definition and it is used across contracts. Already added ticket for those on the Notion board

src/passes/references/expectedLocationAnalyser.ts
```diff
@@ -131,6 +131,11 @@ export class ExpectedLocationAnalyser extends ASTMapper {
         node.vArguments.forEach((arg) => this.expectedLocations.set(arg, DataLocation.Memory));
         return this.visitExpression(node, ast);
       }
+
+      if (node.vFunctionName === 'encode') {
```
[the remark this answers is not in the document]
rodrigo-pino: Yes it is

GuyMcComb (approved):

tests/behaviour/contracts/new/product.sol
```diff
@@ -0,0 +1,55 @@
...
+        products.push(p);
+    }
+
+    function getProduct(uint index) public view returns (
+        uint x, uint8 y, uint[] memory z, uint8[3] memory v, S memory w
+    ) {
+        Product p = products[index];
+        x = p.x();
+        y = p.y();
+        z = p.getZ(); // This are bugged
```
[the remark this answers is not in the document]
rodrigo-pino: I added the task on the notion board

PR #611, opened 2022-07-14
Adding block methods for .timestamp and .number

GuyMcComb:


RohitRanjangit (approved):

tests/behaviour/contracts/builtins/block.sol
```diff
@@ -0,0 +1,17 @@
...
+contract WARP {
+
+    function number() external view returns (bool) {
+        uint248 y = uint248(block.number);
+        return y > 0;
+    }
+ 
+    // Testnet returns 0 for get_block_timestamp so there is no test.
+    function timestamp() external view returns (uint248) {
+        uint248 y = uint248(block.timestamp);
```
GuyMcComb, line 14: This should be wrapped in felt to Uint256 in the Cairo world.

src/passes/builtinHandler/blockMethods.ts
```diff
@@ -0,0 +1,88 @@
...
+      node.vExpression instanceof Identifier &&
+      node.vExpression.name === 'block' &&
+      node.vExpression.vIdentifierType === ExternalReferenceType.Builtin
+    ) {
+      if (node.memberName === 'number') {
+        const replacementCall = createCallToFunction(
+          createCairoFunctionStub(
+            'get_block_number',
+            [],
+            // Chosen since this will be transformed to felt, so width does not matter.
```
DomHenderson, line 31: I find this comment quite misleading. It reads like it's saying that since the result is a felt and we treat uint8s as felts that inherently this is fine. When in actuality it's fine because the call with be wrapped with felt_to_uint256. It's also an important note because of the ongoing PR on structs for numeric types. Might it be a better idea to create a warplib function that returns a uint256 for the block number/timestamp?

src/passes/builtinHandler/blockMethods.ts
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      return;
+    }
+  }
+}
+
+function wrapInFeltToUint256(node: Expression, ast: AST): void {
+  const funcStub = createCairoFunctionStub(
+    'felt_to_uint256',
+    [['x', typeNameFromTypeNode(getNodeType(node, ast.compilerVersion), ast)]],
```
DomHenderson, line 77: I know it's nitpicky, but the parameters might as well be called in and out, or feltIn and uint256Out or w/e, just something so that someone who doesn't know how createCairoFunctionStub works by heart can tell what each line is

tests/behaviour/contracts/builtins/block.sol
```diff
@@ -0,0 +1,17 @@
...
+
+
+contract WARP {
+
+    function number() external view returns (bool) {
+        uint256 y = block.number;
+        return y > 0;
+    }
+ 
+    // Testnet returns 0 for get_block_timestamp so there is no test.
```
DomHenderson, line 12: If you only have this file up, this comment is confusing. I'd suggest spelling it out that there is no associated expectation and that it is here only to verify that the function can compile into compilable cairo

DomHenderson (changes requested):
Mostly looks fine, just left a few nitpicks. Might also be worth including the other block members in rejectUnsupportedFeatures, as afaik they're not there

src/passes/builtinHandler/blockMethods.ts
```diff
@@ -0,0 +1,54 @@
...
+import { createCairoFunctionStub, createCallToFunction } from '../../utils/functionGeneration';
+import { createUint256TypeName } from '../../utils/nodeTemplates';
+
+/*
+A subpass that replaces the block.timestamp and block.number methods.
+In Solidity these functions return uint256, but in cairo these will return felts.
+Therefore, we wrap the replacement functions in felt_to_uint256 warplib functions.
+*/
+export class BlockMethods extends ASTMapper {
+  visitMemberAccess(node: MemberAccess, ast: AST): void {
```
DomHenderson, line 13: This method needs to recurse if the member access doesn't match the expeceted pattern. Consider `f(block.timestamp).member`

DomHenderson (changes requested):
One more change request I'm afraid

DomHenderson (approved):

PR #612, opened 2022-07-14
Adding Literal Fix

GuyMcComb:
Fixing typo in Comparison function.

### DIFF
--- a/src/warplib/utils.ts
+++ b/src/warplib/utils.ts
@@ -11,6 +11,7 @@ import {
   FunctionCallKind,
   getNodeType,
   Identifier,
+  IntLiteralType,
   IntType,
   TypeNode,
 } from 'solc-typed-ast';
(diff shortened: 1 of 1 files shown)


RohitRanjangit:
I have been thinking, what if lhs is of type IntLiteralType instead of rhs?

GuyMcComb:
> I have been thinking, what if lhs is of type IntLiteralType instead of rhs? or Possibly both lhs and rhs
> 
> ```solidity
> contract WARP {
>   function a() public {
>     bool res = (address(0) != address(1));
>   }
> }
> ```

You are right, I will add an option for if both are literals. It is interesting that `address(0)` is transformed into a `IntLiteralType `node, but `0` is transformed into `IntType` Node.

DomHenderson (commented):
This is not actually a typo, it was written with the intention that both sides of the operator should have the same type. An assertion that both sides are the same type might be more appropriate, as well as a fix for address(0) becoming IntLiteralType

DomHenderson:
> I feel like the Comparison function (The function causing this error) should take the type of the LHS for both inputs if the TypeNode of the RHS is an IntLiteral and vice versa. If they are both LiteralIntTypes then we create the IntTypes TypeNodes for the CairoStub manually.

They should not be intliteral types by this point in the code. The fact that they are is a bug. ImplicitConversionToExplicit should wrap all literal types being used in cases like these in conversions with appropriate typestrings, and explicitToFunc should preserve the typestring of the conversion function in the cases that it replaces the call with its operand

PR #613, opened 2022-07-15
Prerequisite passes

RohitRanjangit:
https://www.notion.so/nethermind/Warp-2-0-Transpiler-af80a911dd8742b28604aa604fe6455f?v=775dc2465c794cf49412161ee3d0126c&p=9e88c6bb690848bd98bc87ce65cf5f17


src/utils/cliOptionParsing.ts
```diff
@@ -33,10 +37,39 @@ export function parsePassOrder(
     }
     const [key, nextPass] = foundPass;
     passesInOrder.push(nextPass);
+    keyPassesInOrder.push(key);
     if (key === until) break;
     remainingOrder = remainingOrder.slice(key.length);
   }
 
+  passesInOrder.forEach((element, index) => {
+    const prerequites = element._getPassPrerequites();
```
GuyMcComb, line 46: Spelling mistake here prerequites ->prerequisite.

src/utils/cliOptionParsing.ts
```diff
@@ -33,10 +37,39 @@ export function parsePassOrder(
...
+    const prerequitesKeys = [...keyPassesInOrder].slice(0, index);
```
GuyMcComb, line 47: I think we should rename this EarlierPassKeys, since these are the passes that are occurring before. Prerequisite means that it is needed before the current pass.

src/ast/mapper.ts
```diff
@@ -11,6 +11,27 @@ import { ASTVisitor } from './visitor';
 */
 
 export class ASTMapper extends ASTVisitor<void> {
+  // List of passes that should have been run before this one
+  prerequisite: Set<string> = new Set<string>();
+
+  addPassPrerequisite(pass_key: string) {
```
swapnilraj, line 17: you can make the logic easier and change the argument from to `pass_key : string` to `passKeys : Set<string>` and then just do `prerequisites.add(...[...passKeys])`

RohitRanjangit: My intent was that we can add a single key to pre-req list for a Pass through addPassPrerequisite function which can be used later. Also, unpacking like this is not supported in add method for a set.

GuyMcComb (approved):

PR #615, opened 2022-07-18
Support conditionals

cicr99:


src/passes/expressionSplitter/conditionalFunctionaliser.ts
```diff
@@ -165,3 +165,39 @@ export function addStatementsToCallFunction(
 export function getNodeVariables(node: Conditional) {
   return new Map([...collectUnboundVariables(node)].filter(([decl]) => !decl.stateVariable));
 }
+
+// There might be two different cases where conditionals return void:
+// - they are the expression of a return statement
+// - they are the expression of an expressionStatement
+// TODO - Check if there are any other cases where the conditional might be void
```
DomHenderson, line 172: They can also be nested in other void conditionals

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -59,12 +58,6 @@ export class RejectUnsupportedFeatures extends ASTMapper {
   visitErrorDefinition(node: ErrorDefinition, _ast: AST): void {
     throw new WillNotSupportError('User defined Errors are not supported', node);
   }
-  visitConditional(node: Conditional, _ast: AST): void {
```
DomHenderson, line 62: This should still reject non-void tuple conditionals as they aren't supported (right?). It should also reject tuple assignments that are children of anything other than an expression statement

tests/behaviour/contracts/conditionals/conditional.sol
```diff
@@ -0,0 +1,27 @@
+pragma solidity ^0.8.10;
+
+// SPDX-License-Identifier: MIT
+
+contract WARP {
+  uint256 public x;
+
+  function simpleScalar(
```
DomHenderson, line 8: I'd appreciate a test case that includes conditionals where both branches are literals. Ideally one for number literals as well as some for the various kinds of string literals.

DomHenderson (changes requested):

cicr99:
It was decided not to support conditionals yet due to many edge cases that can not be handled at the moment

PR #617, opened 2022-07-20
Unloading assignment bug

DomHenderson:
Previously, we handled compound assignments by converting x += 1 into x = x + 1 and leaving it at that. However this meant that we converted a[i++] += 1 into a[i++] = a[i++] + 1, which is not equivalent. This fixes this by extracting volatile subexpressions like i++, but still cloning things like index accesses. This is necessary because just extracting the lhs to a constant and cloning that would assign to the 'constant', not to the element in the array


src/passes/unloadingAssignment.ts
```diff
@@ -63,4 +82,48 @@ export class UnloadingAssignment extends ASTMapper {
...
+        `Unexpected empty index access in compound assignment ${printNode(node)}`,
+      );
+      if (!(node.vIndexExpression instanceof Literal)) {
+        const decl = this.extractIndividualSubExpression(node.vIndexExpression, ast);
+        return [decl, ...this.extractUnstableSubexpressions(node.vBaseExpression, ast)];
+      }
+    } else if (node instanceof MemberAccess) {
+      return this.extractUnstableSubexpressions(node.vExpression, ast);
+    } else if (node instanceof FunctionCall) {
+      return [this.extractIndividualSubExpression(node, ast)];
```
GuyMcComb, line 109: Would be interested to see an example of Function(i++) being used as an lvalue or am I missing what is going on here?

GuyMcComb (changes requested):

src/passes/unloadingAssignment.ts
```diff
@@ -63,4 +82,48 @@ export class UnloadingAssignment extends ASTMapper {
...
+  // node is the lvalue of a compound assignment
+  // this function extracts subexpressions that are not guaranteed to produce the same result every time
+  // to constants to evaluate once before the assignment
+  // this is needed because extracting to a 'constant' and then assigning to it would not modify the correct variable
+  // we also extract non-trivial expressions so they're only calculated once, though this is not technically necessary
+  // an array of created statements are returned so they can be visited
+  private extractUnstableSubexpressions(
+    node: Expression,
+    ast: AST,
+  ): VariableDeclarationStatement[] {
```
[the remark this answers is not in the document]
DomHenderson: The return value is used on line 29 when we iterate through the returned array and dispatchVisit each one. Without this you can extract increment operators from inside compound assignments and then not process them

src/passes/unloadingAssignment.ts
GuyMcComb: Oh I tough you were saying `getArray(i++)`

GuyMcComb (approved):

DomHenderson:
Converted to draft whilst I check one final case

DomHenderson:
Case I was concerned about is not actually valid (assignments are not lvalues), so nothing more to be added since the review

PR #619, opened 2022-07-20
Improve efficiency of control flow algorithms

DomHenderson:
The previous CFA algorithm was causing out of memory errors. It was used in two cases, neither of which required all the information it was gathering. I've replaced it with two much smaller algorithms with vastly better space usage


GuyMcComb (approved):

GuyMcComb:
Semantic Tests Passed

PR #620, opened 2022-07-21
Uniswap bump to 0.8.14

rodrigo-pino:
Initially tried bumping using `solidity-upgrade` tool but it did more harm than good producing invalid code on `NoDelegateCall` and no upgrades anywhere else.

So I ended doing the bump manually, which was mostly changing implicit to explicit conversions. Also checked the code for any other breaking changes (including silent ones) according to this list [here](https://docs.soliditylang.org/en/v0.8.15/080-breaking-changes.html#how-to-update-your-code) but found none

### DIFF
--- a/v3-core/contracts/libraries/TickBitmap.sol
+++ b/v3-core/contracts/libraries/TickBitmap.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity >=0.5.0;
+pragma solidity ^0.8.14;
 
 import './BitMath.sol';
 
@@ -13,7 +13,7 @@ library TickBitmap {
     /// @return bitPos The bit position in the word where the flag is stored
     function position(int24 tick) private pure returns (int16 wordPos, uint8 bitPos) {
         wordPos = int16(tick >> 8);
-        bitPos = uint8(tick % 256);
+        bitPos = uint8(uint24(tick % 256));
     }
 
     /// @notice Flips the initialized state for a given tick from false to true, or vice versa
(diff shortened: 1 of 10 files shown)


v3-core/contracts/libraries/TickBitmap.sol
```diff
@@ -71,8 +71,8 @@ library TickBitmap {
             initialized = masked != 0;
             // overflow/underflow is possible, but prevented externally by limiting both tickSpacing and tick
             next = initialized
-                ? (compressed + 1 + int24(BitMath.leastSignificantBit(masked) - bitPos)) * tickSpacing
-                : (compressed + 1 + int24(type(uint8).max - bitPos)) * tickSpacing;
+                ? (compressed + 1 + int24(uint24(BitMath.leastSignificantBit(masked) - bitPos))) * tickSpacing
+                : (compressed + 1 + int24(uint24(type(uint8).max - bitPos))) * tickSpacing;
```
GuyMcComb, line 75: What is going on with these wrappers?
What happens if the expression in the function call is negative now..

rodrigo-pino: The output type of the inner expressions are uint8, they are being increased in sized to uint24, and then cast to int24. There is no information lost. Also this the way it behaves in the same way solc 0.7.6 with implicit conversion, this is important because there are cases where there are two valid ways of casting but one produces different output than the others. For example in this same case for value `255` if instead of:
casting like `uint8 -> uint24 -> int24` which produces (255 -> 255 -> 255)
I  cast as `uint8 -> int8 -> int24` that produces instead (255 -> -128 -> -128), which is not the expected value in the original Uniswap contract

PR #621, opened 2022-07-21
Modify UniV3 ABI.ecode packed with PedersonHash Cairo Block

RohitRanjangit:
https://www.notion.so/nethermind/Modify-UniV3-ABI-ecode-packed-with-PedersonHash-Cairo-Block-08745a595d414fabbe183fbc38d5d5c2


v3-core/contracts/libraries/Position.sol
```diff
@@ -27,15 +27,53 @@ library Position {
...
+    /// }(
+    ///     __warp_usrid6_self : felt,
+    ///     __warp_usrid7_owner : felt,
+    ///     __warp_usrid8_tickLower : felt,
+    ///     __warp_usrid9_tickUpper : felt,
+    /// ) -> (__warp_usrid10_position : felt):
+    ///     from warplib.memory import wm_alloc, wm_write_felt
+    ///     alloc_locals
+    ///     let __warp_usrid10_position = 0
+    ///     let (__warp_se_0) = wm_alloc(Uint256(299, 0))
```
GuyMcComb, line 48: I think we would rather use wm_new since it creates a dynarray in memory. Then you can wm_write_felt the three values into the 3 new felt slots and then pass the initial location to warp_keccak.

GuyMcComb (changes requested):
I think change the method you are using to manipulate the memory from wm_alloc to wm_new and we should be good to go.

v3-core/contracts/libraries/Position.sol
```diff
@@ -45,12 +45,12 @@ library Position {
...
-    ///     wm_write_felt(__warp_se_0, __warp_usrid7_owner)
-    ///     wm_write_felt(__warp_se_0 + 251, __warp_usrid8_tickLower)
-    ///     wm_write_felt(__warp_se_0 + 275, __warp_usrid9_tickUpper)
+    ///     let (__warp_se_0) = wm_new(Uint256(3, 0), Uint256(1,0))
+    ///     wm_write_felt(__warp_se_0 + 2, __warp_usrid7_owner)
+    ///     wm_write_felt(__warp_se_0 + 3, __warp_usrid8_tickLower)
+    ///     wm_write_felt(__warp_se_0 + 4, __warp_usrid9_tickUpper)
     ///     let (__warp_se_1) = warp_keccak(__warp_se_0)
-    ///     let (__warp_se_2) = WS0_INDEX_Uint256_to_Info_9555281d(__warp_usrid6_self, __warp_se_1)
+    ///     let (__warp_se_2) = WS0_INDEX_Uint256_to_Info_d529aac3(__warp_usrid6_self, __warp_se_1)
```
DomHenderson, line 53: I'd be very careful with this, the name of that function is liable to change based on what else is in the file that's using it. It'd be safer to have a manual implementation of it here instead, but in simple cases this does work and we can modify it later if it needs using in more complex cases

DomHenderson (approved):

GuyMcComb (approved):

PR #624, opened 2022-07-26
Starknet 0.9.1 deploy command update

yaziciahmet:
`starknet deploy` command requires `--class_hash` option now, if `--no_wallet` option is not provided. Class hash is obtained from the command `starknet declare` . Stdout of `declare` is piped so it is not visible in the terminal.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -136,6 +136,7 @@ program
 interface IDeployProps_ {
   inputs?: string;
   use_cairo_abi: boolean;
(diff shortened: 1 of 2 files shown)


GuyMcComb (changes requested):
Could you add a change that checks for --inputs. Otherwise if you do not specify them the deploy command fails. We want the deploy command to not take --inputs as the default. Similar to how you have --use_wallet

PR #629, opened 2022-07-27
Reject Unsupported Feature Recursion

GuyMcComb:


BitBaseBit:
Pls explain more in the PR comment

GuyMcComb:
Adds a recursion so that we can inspect function options to see if there are any unsupported features. The example that brought this up was having abi.encode in the salt of a contract factory.

DomHenderson (approved):

PR #634, opened 2022-08-02
Semantic Tests for specific solc release

RohitRanjangit:
The solidity semantic tests are currently pulled from the develop branch and a whitelist used. This should be changed such that they are pulled from a specific release, and the whitelist format changed such that we can still easily test specific tests or groups during development, but we can also run a script to make sure that no tests are unaccounted for in the list

Now you can run the semantic tests from specfic solidity release: 
for example: `VERSION=0.8.14 yarn test:semantic` will pull solc-0.8.14 semantic tests and execute them

You can also use filter variable to filter group of tests / specific test. 
by :
`VERSION=0.8.14 FILTER=tests/behaviour/solidity/test/libsolidity/semanticTests/array yarn test:semantic`

also, you can enable compare flag by providing `--compare`, it'll compare the tests from `semantic_whitelist.ts` and solc VERSION semanticTests after applying the FILTER

UPDATE: we can only test for versions 0.8.5 or later, because of the isoltest dependency!


JorikSchellekens:
We'll need to run these on AWS before we merge

RohitRanjangit:
@GuyMcComb pls update ,once you run the script on AWS

GuyMcComb (approved):

GuyMcComb:
Does not break the old semantic tests, so fine for it to go in.

cicr99 (approved):
LGTM

PR #638, opened 2022-08-04
Fixes for `warp deploy` command

Glitch18:
The `warp deploy` command internally uses `starknet declare` before it runs the `starknet deploy` command. This currently fails because the declare command needs the `--network` flag which does not get passed to it. Added a fix for the same. 

( To recreate the issue, run 
```
bin/warp deploy warp_output/tests/benchmark/BaseJumpRateModelV2__WC__BaseJumpRateModelV2.cairo --inputs 0,0,0,0,0 --network alpha-goerli
```
 on the develop branch )

If we do not specify `--no_wallet`, the deploy command errors out with the following message:
```
Error: AssertionError: A wallet must be specified (using --wallet or the STARKNET_WALLET environment variable), unless specifically using --no_wallet.
starknet deploy failed
```
To fix this, the user can either set the `STARKNET_WALLET` env variable or pass `--wallet` in the `warp deploy` command.
Added support for the `--wallet` option for deploy command.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -148,6 +148,7 @@ interface IDeployProps_ {
   inputs?: string;
   use_cairo_abi: boolean;
   no_wallet: boolean;
+  wallet?: string;
 }
 
 export type IDeployProps = IDeployProps_ & IOptionalNetwork & IOptionalAccount & IOptionalDebugInfo;
@@ -163,6 +164,7 @@ program
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'Starknet network URL', process.env.STARKNET_NETWORK)
   .option('--no_wallet', 'Do not use a wallet for deployment.', false)
+  .option('--wallet <wallet>', 'Waller provider to use', undefined)
   .option('--account <account>', 'Account to use for deployment', undefined)
   .action((file: string, options: IDeployProps) => {
     runStarknetDeploy(file, options);
--- a/src/index.ts
+++ b/src/index.ts
@@ -164,7 +164,7 @@ program
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'Starknet network URL', process.env.STARKNET_NETWORK)
   .option('--no_wallet', 'Do not use a wallet for deployment.', false)
-  .option('--wallet <wallet>', 'Waller provider to use', undefined)
+  .option('--wallet <wallet>', 'Waller provider to use', process.env.STARKNET_WALLET)
   .option('--account <account>', 'Account to use for deployment', undefined)
   .action((file: string, options: IDeployProps) => {
     runStarknetDeploy(file, options);
--- a/src/index.ts
+++ b/src/index.ts
@@ -164,7 +164,7 @@ program
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'Starknet network URL', process.env.STARKNET_NETWORK)
   .option('--no_wallet', 'Do not use a wallet for deployment.', false)
-  .option('--wallet <wallet>', 'Waller provider to use', process.env.STARKNET_WALLET)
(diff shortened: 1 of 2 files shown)


src/starknetCli.ts
```diff
@@ -200,7 +206,11 @@ export async function runStarknetDeploy(filePath: string, options: IDeployProps)
     const classHash = compileResult.classHash;
     execSync(
       `${warpVenvPrefix} starknet deploy --network ${options.network} ${
-        options.no_wallet ? `--no_wallet --contract ${resultPath} ` : `--class_hash ${classHash}`
+        options.no_wallet
+          ? `--no_wallet --contract ${resultPath} `
+          : options.wallet === undefined
+          ? `--class_hash ${classHash}`
```
rodrigo-pino, line 212: (Related to previous comment)
And then you could scrap this  extra ternary (and any extra ternary of the future that needs the `wallet` option). For this to be fully functional you will need to verify at the beginning that `options.no_wallet || option.wallet !==undefined` expression is True

rodrigo-pino: Also nitpick:  Should we aloud the `options.no_wallet` and the `option.wallet` to be passed at the same time. Perhaps modifying the previous boolean expression to using a xor like `options.no_wallet xor option.wallet !==undefined`

rodrigo-pino (commented):
Overall it looks good to me

src/index.ts
```diff
@@ -163,6 +164,7 @@ program
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'Starknet network URL', process.env.STARKNET_NETWORK)
   .option('--no_wallet', 'Do not use a wallet for deployment.', false)
+  .option('--wallet <wallet>', 'Waller provider to use', undefined)
```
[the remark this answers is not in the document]
Glitch18: Added, thanks!

Glitch18:
I've added the environment variable of `STARKNET_WALLET` as a default value for the wallet option. 

Although I don't see how we can avoid using a double ternary. In this situation, we have three different cases, each requiring different options for the `starknet deploy` command. 

1. `--no-wallet` is passed => We pass --no-wallet in `starknet deploy` and omit `--wallet` 
2. Wallet is passed (through CLI or env var) => We pass it on to `starknet deploy`
3. Wallet is not passed (neither through CLI or env var) => We call `starknet deploy` without `--no-wallet` and without `--wallet` which gives the user a useful prompt. 

Would it be possible to handle three different cases without using a double ternary? The only other way I see it would be to use if/else or a switch statement.

Also, I think we should not add a check for `--wallet` and `--no-wallet` being present together. Because the user can have the `STARKNET_WALLET` environment variable set, but might want to use `--no-wallet` for a singular deployment. So if we throw an error if `--wallet` and `--no-wallet` are not mutually exclusive, it might be inconvenient for the user.

rodrigo-pino:
>     3. Wallet is not passed (neither through CLI or env var) => We call `starknet deploy` without `--no-wallet` and without `--wallet` which gives the user a useful prompt.

Yes you're right, I was missing that this was possible as well

rodrigo-pino (approved):

src/index.ts
```diff
@@ -163,6 +164,7 @@ program
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'Starknet network URL', process.env.STARKNET_NETWORK)
   .option('--no_wallet', 'Do not use a wallet for deployment.', false)
+  .option('--wallet <wallet>', 'Waller provider to use', process.env.STARKNET_WALLET)
```
cicr99, line 167: Shouldn't the description start with Wallet instead of Waller?

cicr99 (commented):
It looks good, just that small typo. Also, I agree with using the ternaries.

src/index.ts
Glitch18: Ah right, thanks for catching this!

cicr99 (approved):

PR #640, opened 2022-08-05
docusaurus files and gitaction workflows added(LATEST)

davidmitesh:
Documentation website using Docusaurus added to the /docs. Also, GitHub action workflow is added which gets triggered if the PR is merged to main.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+      - 'docs/**'
+      - '.github/workflows/deploy.yml'
```
Glitch18, line 8: I think we only need to list the develop branch here.

Glitch18 (commented):
Tested out the docusaurus website locally. Overall looks good to me.
Let's make the commented change in the Github Action Workflow. The deploy job should run for this PR after that and we can test out the github action.

.github/workflows/deploy.yml
davidmitesh: But Guy told me that the update should only trigger when the PR is merged to the main.

Glitch18: I see, okay. But what about the other two items? do we need them as target branches?

davidmitesh: no those are the target subdirectories and files, to which only if update is made then only the update takes place.

Glitch18: I think you should be able to remove the other two and only keep main as the target branch for this workflow.

Glitch18: Right, for that we would need to use the `paths` event under `push`. Something like:

```
on:
    push:
        paths:
          - 'docs/**'
```

Under `branches` it should only be main

Glitch18: See: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events

davidmitesh: sorry my bad, wanted to do that, mistakely wrote that under the branches. corrected now

Glitch18 (approved):

docs/blog/authors.yml
```diff
@@ -0,0 +1,15 @@
+maceo:
```
GuyMcComb, line 1: Please change this to your name.

docs/blog/2022-06-24-junopi/index.mdx
```diff
@@ -0,0 +1,127 @@
+---
+# slug: junopi
+# title: Running Juno from your Raspberry Pi
+title : Blog template 2 (WIP)
+# authors: [depablos, maceo]
+# tags: [juno, rpi, raspberry, pi, deploy]
+---
+
+<!-- Juno is a node which aims to help **decentralise StarkNet, a prominent Ethereum Layer 2**.
+
```
GuyMcComb, line 10: All of this will not appear right? Will it just say WIP?

docs/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "juno-docs",
```
GuyMcComb, line 2: This should change

GuyMcComb:
The Juno images are still present, please remove them.

docs/blog/authors.yml
davidmitesh: done

docs/blog/2022-06-24-junopi/index.mdx
davidmitesh: yeah

docs/package.json
davidmitesh: done

davidmitesh:
> 

All images from juno are commented I guess. Did you find any one left?

GuyMcComb:
> > 
> 
> All images from juno are commented I guess. Did you find any one left?

[docs/static/img/juno_rounded.png](https://github.com/NethermindEth/warp/pull/640/files#diff-f81e46760c856ac409e3f4d33cbce0be6575483ee9da9ff34f096067bc3cf581)

GuyMcComb (approved):

PR #646, opened 2022-08-09
upgrading packages

GuyMcComb:
Upgraded the js packages by running `yarn upgrade`. This is for the docusaurus site.


GuyMcComb:
Please wait for Sem tests to pass before merging.

PR #648, opened 2022-08-09
getContaining methods refactoring

cicr99:
There were many places in the code where the closest parent which was a `FunctionDefinition` was wanted and then there was an assertion checking this parent node was not `undefined`. Also there were two methods, which enclosed this behavior, so they were the same, in different parts of the code. This PR creates a method `getContainingFunction` for these cases, that is placed in the `utils` folder of the passes.
A similar situation occurred with `SourceUnit`, so a `getContainingSourceUnit` method was created the same way.


rodrigo-pino (approved):

GuyMcComb:
This must have been super satisfying. I will kick off Sem Tests and then this can go in.

GuyMcComb:
Sem tests pass.

GuyMcComb (approved):

PR #654, opened 2022-08-16
Check correct use of getNodeType methods

cicr99:
- Check that the node to parse the `typestring` is attached to a `SourceUnit` node, using `getContainingSourceUnit` method, which already has an assertion to check the result is not undefined
- Replace every use of `getNodeType` and `getNodeTypeInCtx` with the new functions

### DIFF
--- a/src/ast/ast.ts
+++ b/src/ast/ast.ts
@@ -7,7 +7,6 @@ import {
   Expression,
   FunctionCall,
   generalizeType,
-  getNodeType,
   Identifier,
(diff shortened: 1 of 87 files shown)


GuyMcComb:
Sem tests running.

GuyMcComb:
Sem tests pass.

GuyMcComb (approved):
GIANT PR! Good Work!

RohitRanjangit (approved):

PR #655, opened 2022-08-17
Interface call forwarder

rjnrohit:
https://www.notion.so/nethermind/Interface-Call-forwarder-3e2c2737847f4de997a491cd964a3b30

- [x] Handle Cairo Imports
- [x] structs
- [x] array type arguments
- [x] modify stub processor to support  cairo stub from contract documentation
- [x] generated sol writer
- [x] generate functions from cairo abi
- [x] testing thoroughly
- [x] rename solc proxy contract
- [x] take contract address as cli input
- [x] add contract address as an constructor parameter 
- [x] define functions inside proxy contract
- [ ] convert felt to uint256 *(in progress...)*
- [ ] array type argument automation 
- [ ] struct dependency graph
- [ ] import/struct automatic conflict resolution 
- [ ] re-testing


src/interfaceCallForwarder.ts
```diff
@@ -0,0 +1,395 @@
...
+      ),
+    );
+  });
+  functions.forEach((f: FunctionDefinition) => {
+    sourceUnit.appendChild(f);
+    ast.setContextRecursive(f);
+  });
+}
+
+class CairoContractSolWriter extends ASTNodeWriter {
```
JorikSchellekens, line 323: Contract should include function definitions

src/interfaceCallForwarder.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+}
+
+function getSolTypeName(cairoTypeString: string, ast: AST, struct_names: string[]): TypeName {
+  if (cairoTypeString.endsWith('*')) {
+    const baseTypeName: TypeName = getSolTypeName(cairoTypeString.slice(0, -1), ast, struct_names);
+    return createArrayTypeName(baseTypeName, ast);
+  }
+  if (cairoTypeString === 'felt') {
+    return createUintNTypeName(248, ast);
```
JorikSchellekens, line 383: Type is too small, instead we should use Uint256 and do runtime bounds checking.

src/interfaceCallForwarder.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+  if (cairoTypeString === 'Uint256') {
+    return createUint256TypeName(ast);
+  }
+  if (struct_names.includes(cairoTypeString)) {
+    // TODO: replace this with a proper struct type
+    // after creation of struct dependency ordering
+    return new CornerStructTypeName(ast.reserveId(), '', `<STRUCT:${cairoTypeString}>`);
```
JorikSchellekens, line 391: `getSolTypeName` should be called recursively on the structs and create their replacement type representation in Solidity along with their bounds checking functions.

src/interfaceCallForwarder.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+  // throw new Error(`Cairo type: ${cairoTypeString} Not Supported`);
+  return new CornerTypeName(ast.reserveId(), '', `<TYPE:${cairoTypeString}>`);
```
JorikSchellekens, line 394: We should never hit this case and if we do we should throw an error

JorikSchellekens (changes requested):
This still needs a lot of work. The user shouldn't have to do any changes to the generated files to use them in their solidity.

JorikSchellekens:
The output should be a Solidity contract with all the functions calls (without the address parameter) with the wrapping types and their runtime bounds checks along with all of the required struct definitions (linearised).

JorikSchellekens:
ping @RohitRanjangit

RohitRanjangit:
closed , alt to #756

PR #656, opened 2022-08-17
Update rejectUnsupportedFeatures.ts

davidmitesh:
A filter is added on the visitFunctionDefinition that checks if 'this' keyword is encountered in the constructor to call the external function, and abandons the transpilation if encounters such.


rodrigo-pino (changes requested):
You cannot ban the `this` identifier from a constructor. This example can be compiled and deployed:
```Solidity
contract Warp {
    uint public x = 10;
    constructor() {
        x = address(this).balance;
    }
}
```
I would suggest investigating which member accesses of `this` prevent deployment and banning those instead of the identifier.

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -174,6 +174,31 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
+
+    //Checking if the function defination is constructor and
+    // if it is, checking if 'this' keyword is used to call
+    // the external/public functions of the contract
+    const nodesWithThisIdentifier: ASTNode[] | undefined =
+      node.kind === FunctionKind.Constructor
+        ? node.vBody
+            ?.getChildren()
+            .filter((value, _) => value instanceof Identifier && value.raw.name === 'this')
+        : undefined;
```
rodrigo-pino, line 186: This is a nitpick. If the sole point of this variable is to be defined when the function definition is a constructor, I think code readability would improve by swapping the ternary expression for an  if:
```Typescript
if (node.kind === FunctionKind.Constructor) {
 // Declare variables and write any logic related with constructors here
}
```

GuyMcComb (changes requested):

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -174,6 +175,33 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
+    //Checking if the function defination is constructor and
+    // if it is, checking if 'this' keyword is used to call
+    // the external/public functions of the contract
+
+    if (node.kind === FunctionKind.Constructor) {
+      const nodesWithThisIdentifier = node.vBody
+        ?.getChildren()
+        .filter((value, _) => value instanceof Identifier && value.raw.name === 'this');
+
+      nodesWithThisIdentifier?.forEach((node: ASTNode) => {
```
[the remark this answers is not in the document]
davidmitesh: you mean to change  `"... .filter((value, _)  => ..."`   to something like  `"... .filter((node,_) => ..."`  right ?

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -174,6 +175,33 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
+        const parentNode = node?.parent;
+        if (
+          parentNode instanceof MemberAccess &&
+          parentNode?.parent instanceof FunctionCall &&
+          parentNode.parent.vReferencedDeclaration instanceof FunctionDefinition &&
+          (parentNode.parent.vReferencedDeclaration.visibility === FunctionVisibility.Public ||
```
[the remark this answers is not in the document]
davidmitesh: Util function is now utilized to check for external visibility.

src/passes/rejectUnsupportedFeatures.ts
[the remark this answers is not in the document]
davidmitesh: I have changed the variable name in the recent commit. Please confirm if that is the desired change.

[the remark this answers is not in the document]
GuyMcComb: Sorry, what I meant is that node should be reserved for the FunctionDefinition that is being visited by visitFunctionDefinition. The child nodes should be called something else.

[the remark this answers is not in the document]
davidmitesh: The changes are updated in the latest commit :raised_hands:

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -174,6 +174,31 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
+
+      nodesWithThisIdentifier?.forEach((identifierNode: ASTNode) => {
+        const parentNode = identifierNode?.parent;
+        if (
+          parentNode instanceof MemberAccess &&
+          parentNode?.parent instanceof FunctionCall &&
+          isExternalCall(parentNode.parent)
+        ) {
+          throw new WillNotSupportError(
+            `external function calls using this keyword not supported during contract deployment`,
```
GuyMcComb, line 195: Please change this => "this"

davidmitesh: I think it should also be slightly modified as : 

'external function calls using "this" keyword not supported in contract's constructor function'
What's your say on this?

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -174,6 +174,31 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
     }
+
+    //Checking if the function defination is constructor and
+    // if it is, checking if 'this' keyword is used to call
+    // the external/public functions of the contract
+
+    if (node.kind === FunctionKind.Constructor) {
+      const nodesWithThisIdentifier = node.vBody
+        ?.getChildren()
+        .filter((childNode, _) => childNode instanceof Identifier && childNode.raw.name === 'this');
```
GuyMcComb, line 185: Nitpick. childnode.raw.name => childnode.name.

GuyMcComb: Could we also have all this logic wrapped up in a single function, for readability?

davidmitesh: done

GuyMcComb:
Sem Tests Pass

GuyMcComb (approved):

rodrigo-pino (approved):

PR #658, opened 2022-08-18
Add Contract Hash to Contract Factories.

GuyMcComb:


src/utils/utils.ts
```diff
@@ -483,6 +485,19 @@ export function getSourceFromLocation(source: string, location: SourceLocation):
   return [...previousLines, ...currentLine, ...followingLines].join('\n');
 }
 
+export function callClassHashScript(filePath: string): string {
+  const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', 'warp_venv', 'bin')}:$PATH`;
+  const classHash = execSync(
+    `${warpVenvPrefix} python starknet-scripts/compute_class_hash.py ${filePath}`,
```
swapnilraj, line 491: The path to starknet-scripts folder here needs to be relative to the current file, so that it works properly when its distributed through a npm package, similar to how `warpVenvPrefix` is done

GuyMcComb:
Sem Tests Pass

rodrigo-pino (approved):
Elegantly done

src/index.ts
```diff
@@ -64,14 +66,21 @@ program
...
-        transpile(compileSolFile(file, options.warnings), options).map(([name, cairo, abi]) => {
-          outputResult(name, cairo, options, '.cairo', abi);
-        });
+        transpile(compileSolFile(file, options.warnings), options)
+          .map(([name, cairo, abi]) => {
+            outputResult(name, cairo, options, cairoSuffix, abi);
+            return createCairoFileName(name, cairoSuffix);
+          })
+          .forEach((file) => {
+            postProcessCairoFile(file, 'warp_output', contractToHashMap);
```
swapnilraj, line 82: The second argument here should be `options.outputDir`

src/utils/postCairoWrite.ts
```diff
@@ -1,18 +1,137 @@
...
+  During transpilation 0 is placed where the class hash would be. Now that all the needed files are
+  transpiled then we can calculate the class hashes of each and insert them into the file.
+  @param cairoFilePath: The path to the cairo File being processed.
+  @param pathPrefix: The prefix proceeding the the path to the file
+  @param contractHashToClassHash: A mapping that holds the contract path with out the pathPrefix and maps
+  it to the contracts class hash.
+ */
+export function postProcessCairoFile(
+  cairoFilePath: string,
+  pathPrefix = 'warp_output',
```
swapnilraj, line 23: We should remove this default value, I feel this sort of thing might catch us out later.
Especially since the decision of the default directory is decided by the cli part of the code

GuyMcComb: Yeah good spot!

PR #659, opened 2022-08-19
update deployment instruction in docs

jelilat:
The current deployment instruction `warp deploy <path to Cairo contract>` returns an error when you call it. I updated it with the right deployment command.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -106,6 +106,24 @@ bin/warp transpile example_contracts/ERC20.sol
 
 If you have used installation method 1 you can use the `warp` command in any folder. If you have used installation method 2, you will have to specify the path to the warp directory followed by `bin/warp` e.g `path_to_warp_repo/bin/warp ...`
 
+### StarkNet setup
+
+Select your network and wallet types. It's recommended to set these as
+environment variables but they can also be passed as explicit arguments to the
+Warp CLI and the StarkNet CLI.
+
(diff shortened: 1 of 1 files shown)


BitBaseBit (changes requested):
Please change the commit to alert the user in the CLI that we have used those parameters. And then put it in the docs somewhere.

jelilat:
I'm not sure I understand your comment @BitBaseBit

JorikSchellekens:
@jelilat you can create a "user" on StarkNet, check `warp deploy-account` which deploys a user and creates a file tracking them in your home directory. If there is a `__default__` user in this file it will be used by warp and `--no-wallet` is no longer required. Can you update the readme to include the account deployment steps instead of no-wallet? Also the gateway can be set in an env - would be nice to have that in the readme too.

Thank you for catching this!

JorikSchellekens:
I've updated the readme accordingly. Thanks for the initial pr @jelilat

JorikSchellekens (approved):
Going to be cheeky and approve this for myself :D

jelilat:
Thanks @JorikSchellekens 

Sorry I wasn't able to get to this sooner.

swapnilraj (approved):

JorikSchellekens:
Ah np, just appreciate the input!

PR #660, opened 2022-08-23
External call to other reference type bug fix

rodrigo-pino:
Previously it treated all reference types from cross contract calls as dynamic arrays, and tried to pack them inside a warp dynamic array struct, now it only does this for darrays. Add test for other reference types as well


GuyMcComb:
Sem Tests Running.

GuyMcComb (approved):

RohitRanjangit (approved):

PR #662, opened 2022-08-25
refac: starknetCli

syahirAmali:
- refactored existing functions
- created buildCairoCommand to generate starknet commands
- created callCairoCommand that executes the execSync
- broke down functions to be used for testing


src/starknetCli.ts
```diff
@@ -288,26 +366,65 @@ export async function runStarknetCallOrInvoke(
...
-    );
+    return processDeclareCLI(result, filePath);
   } catch {
-    logError(`starknet ${callOrInvoke} failed`);
+    logError('StarkNet declare failed');
   }
 }
 
-export function runStarknetDeclare(filePath: string) {
+export function runStarknetDeclare(filePath: string, options: IDeclareOptions) {
```
GuyMcComb, line 390: What do you think about splitting up the options in this function and passing them individually to the declare function? Could be either way.

src/starknetCli.ts
```diff
@@ -1,58 +1,105 @@
...
+    const result = execSync(cmd, options);
+
+    return result;
+  } catch (e) {
+    logError(`StarkNet ${command} failed`);
+  }
+}
+
+function buildCairoCommand(
+  options: any[] | Map<string, string | undefined>,
```
GuyMcComb, line 45: Avoid any at all costs.

src/starknetCli.ts
syahirAmali: I think using options would be better, it would be easier to add if there were any changes. And I guess that it is only taking 1 option, but for the rest of the commands using options seems to be the way its done, if we were to change declare, I think we should change the others.

src/starknetCli.ts
syahirAmali: I could change the options type from any [here](https://github.com/NethermindEth/warp/pull/662/commits/0dca7e9f54a2b5d924d14b87095acee55977020b), but for line 47 `multioptions?: any`, I could use some help with this

syahirAmali:
Just a couple of comments on the test suite made;

- not entirely sure how to export the commanderjs commands from index.ts to the testing file, so right now, it is simply a duplicate of the command and options
- console outputs from errors aren't being suppressed during testing so they will still be visible

src/starknetCli.ts
syahirAmali: Okay changes made

src/starknetCli.ts
```diff
@@ -97,39 +155,74 @@ async function compileCairoDependencies(
...
-    logError(
+    logCLIError(
       `Error: Exception: feeder_gateway_url must be specified with the "deploy" subcommand.\nConsider passing --network or setting the STARKNET_NETWORK environment variable.`,
     );
     return;
   }
   // Shouldn't be fixed to warp_output (which is the default)
   // such option does not exists currently when deploying, should be added
-  const dependencyGraph = getDependencyGraph(filePath, 'warp_output');
+  const dependencyGraph: Map<string, string[]> = getDependencyGraph(filePath, 'warp_output');
```
swapnilraj, line 225: Can you also include the fix for the comment above? You'll need to pass the outputDir from the cli options

swapnilraj (changes requested):

src/starknetCli.ts
```diff
@@ -164,75 +257,103 @@ export async function runStarknetDeploy(filePath: string, options: IDeployProps)
...
-        stdio: 'inherit',
-      },
-    );
-  } catch {
-    logError('starknet deploy failed');
-  }
+  return cmd;
+}
+
+export function starknetCallOrInvoke(
```
GuyMcComb, line 308: Could you add a suffix here like starknetCallOrInvokeCmd to show that this is where the command is being produced.

syahirAmali: ok changes made

src/starknetCli.ts
syahirAmali: option added and passed to functions

syahirAmali:
- Added command program factory for current commands
- Tests and index.ts now uses the program factory

GuyMcComb:
Do not forget to refactor --cairo-compile like it is in the current CLI.
Also, do not forget to removed the delcare from deploy, make that optional.

JorikSchellekens:
There have been some significant fixes to the cli in `src/index.ts` as a result of the Uniswap tests. They'll need to be reflected in this pr.

swapnilraj:
This looks abandoned so closing it, will re-open if someone wants to continue the work

PR #664, opened 2022-08-29
Library constants

cicr99:
Addressed problem: 
Usually, if the initial value of a state variable is different from a constant, it gets extracted into a block that is added into the constructor of the contract. In the case of library constants, that initialization also gets extracted, but since libraries can not have constructors, it never gets added anywhere, so the value of that variable is lost.

Implemented solution:
- Instead of adding the initialization block into a constructor, a private function is created using this block as the body. This is done for all cases, even libraries. This is handled inside the `StorageAllocator` pass. This pass also added a constructor for each contract, but this is handled in another pass now.
- Solving constructors in `inheritanceInliner` pass, is where a constructor is added for each contract. Also it looks for the variable initialization functions generated in the previous step, and adds a call to those functions inside the constructor's body


GuyMcComb:
Sem Tests Running

GuyMcComb:
Fails:  

- tests/behaviour/solidity/test/libsolidity/semanticTests/constructor_inheritance_init_order:
- tests/behaviour/solidity/test/libsolidity/semanticTests/constructor_inheritance_init_order_3_viaIR

GuyMcComb:
Sem Tests good to go.

GuyMcComb (approved):

RohitRanjangit (approved):

src/passes/inheritanceInliner/inheritanceInliner.ts
```diff
@@ -25,11 +25,21 @@ export class InheritanceInliner extends ASTMapper {
 
   // Function to add passes that should have been run before this pass
   addInitialPassPrerequisites(): void {
-    const passKeys: Set<string> = new Set<string>([]);
+    const passKeys: Set<string> = new Set<string>([
+      // StorageAllocator takes care of variable initialization, which
+      // are needed when solving constructorss
```
RohitRanjangit, line 30: nitpick, spelling mistake

RohitRanjangit (changes requested):

RohitRanjangit (approved):

PR #672, opened 2022-09-01
Fixing interface calls that have CallData Location

GuyMcComb:
Fixing data location of reference types in function calls.

The solidity compiler gives the wrong data location for calldata types in interface calls.
It converts the location from calldata to memory for no clear reason.
Have a look at the AST the following code provides:

```
pragma solidity ^0.8.14;

interface fooInterface  {
    function foo(bytes calldata x) external;
    function foo1(uint8[] calldata x) external;
    function foo2(bytes calldata x) external;   
}

// interface barInterface is fooInterface {}

contract testContract {
    function getSwapResult(
        address pool,
        bytes calldata x,
        uint8[] calldata y
    )
        external
    {
        fooInterface(pool).foo(x);
        fooInterface(pool).foo1(y);
    }

}

```
It is weird and there is nothing we can do about it.

Previously we fixed this by getting the data location from the function call, but the data location should come from the definition and not the call.


GuyMcComb:
Sem Tests Pass

GuyMcComb:
Sem Tests Pass again.

RohitRanjangit (approved):

swapnilraj (approved):

PR #675, opened 2022-09-02
'delete' keyword must be outermost expression

AlejandroLabourdette:


RohitRanjangit (commented):
Create a PR from this repo, not from your forked repo

src/passes/deleteHandler.ts
```diff
@@ -27,6 +28,15 @@ export class DeleteHandler extends ASTMapper {
...
     }
 
+    // The delete operation is expected to be the outermost expression of the line of
+    // code. Since delete does not return a value, it should not be replaced by an
+    // assignment (which does return a value), unless it satisfies the above property.
+    const parent = node.getClosestParentByType(ExpressionStatement);
+    if (parent === undefined || parent.vExpression !== node)
+      throw new WillNotSupportError(
+        'Delete operations have to be the outermost expression to be supported',
+      );
```
rodrigo-pino, line 38: This pass handles deletion from all types except  types in storage, since variables in storage behave different than in memory. 
In the existence of `a` where `a` is a storage variable, `delete a` does not produce an assignation but a custom function specially designed to erase `a`. This occurs on the `references` pass, on `delete` subpass. (src/passes/references/delete.ts)

My question is, does this delete behaviour should be applied to storage variables as well? Also if this is an unsupported feature, shouldn't it be implemented inside the `RejectUnsopportedFeatures` pass?

PR #679, opened 2022-09-05
'delete' keyword must be outermost expression

AlejandroLabourdette:


src/passes/deleteHandler.ts
```diff
@@ -38,6 +39,15 @@ export class DeleteHandler extends ASTMapper {
       return;
     }
 
+    // The delete operation is expected to be the outermost expression of the line of
+    // code. Since delete does not return a value, it should not be replaced by an
+    // assignment (which does return a value), unless it satisfies the above property.
+    const parent = node.getClosestParentByType(ExpressionStatement);
+    if (parent === undefined || parent.vExpression !== node)
+      throw new WillNotSupportError(
+        'Delete operations have to be the outermost expression to be supported',
```
swapnilraj, line 48: You can pass the node as a second argument to WillNotSupportError constructor so that the user gets a nice error message about what exact piece of code is not supported.

cicr99 (approved):
LGTM

swapnilraj:
Can you add a test where `WillNotSupportError` is thrown?

GuyMcComb:
Sem Tests Passed.

GuyMcComb (approved):

PR #681, opened 2022-09-05
Library import bug

cicr99:
The `ReferencedLibraries` pass was modified to correctly handle the case in which the referenced library refers to another and this in turn to another and so on

### DIFF
--- a/src/passes/referencedLibraries.ts
+++ b/src/passes/referencedLibraries.ts
@@ -1,11 +1,5 @@
 import assert from 'assert';
-import {
-  ASTNode,
-  ContractDefinition,
-  ContractKind,
-  FunctionCall,
-  MemberAccess,
-} from 'solc-typed-ast';
+import { ContractDefinition, ContractKind, FunctionCall, MemberAccess } from 'solc-typed-ast';
 import { AST } from '../ast/ast';
 import { ASTMapper } from '../ast/mapper';
 
@@ -25,7 +19,7 @@ export class ReferencedLibraries extends ASTMapper {
   visitFunctionCall(node: FunctionCall, ast: AST): void {
     const librariesById = new Map<number, ContractDefinition>();
     if (node.vExpression instanceof MemberAccess) {
-      //Collect all library nodes and their ids in the map 'contractDef'
+      // Collect all library nodes and their ids in the map 'librariesById'
       ast.context.map.forEach((astNode, id) => {
         if (astNode instanceof ContractDefinition && astNode.kind === ContractKind.Library) {
           librariesById.set(id, astNode);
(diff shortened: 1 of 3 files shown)


src/passes/referencedLibraries.ts
```diff
@@ -37,46 +31,64 @@ export class ReferencedLibraries extends ASTMapper {
...
-      //Checks if the Function is a referenced Library functions,
-      //if yes add it to the linearizedBaseContract list of parent ContractDefinition node
-      //free functions calling library functions are not yet supported
-      librariesById.forEach((library, _) => {
-        if (library.vFunctions.some((libraryFunc) => libraryFunc.id === calledDeclaration.id)) {
-          const parent = node.getClosestParentByType(ContractDefinition);
-          if (parent === undefined) return;
+      // Free functions calling library functions are not yet supported
+      const parent = node.getClosestParentByType(ContractDefinition);
+      if (parent === undefined) return;
```
RohitRanjangit, line 36: In that case, we should halt the transpilation with proper msg rather than ignoring it.

src/passes/referencedLibraries.ts
```diff
@@ -37,46 +31,64 @@ export class ReferencedLibraries extends ASTMapper {
...
+
+function getAllLibraries(
+  calledLibrary: ContractDefinition,
+  librariesById: Map<number, ContractDefinition>,
+  ids: number[],
+): void {
+  ids.push(calledLibrary.id);
 
   calledLibrary
     .getChildren()
```
RohitRanjangit, line 81: Considering every children of the library, the list ids would also have id's of unecessary libraries, Shouldn't we consider the function call which is inside the body of refrenced function ?

RohitRanjangit (changes requested):

GuyMcComb:
Sem Tests Running.

GuyMcComb:
Sem Tests Passed.

src/passes/referencedLibraries.ts
JorikSchellekens: Yes you are right, but I don't think we need it to get this working @GuyMcComb can you add this as a ticket to optimise and land the pr as is?

src/passes/referencedLibraries.ts
cicr99: The previous pass is in charge of adding the free functions to the contracts, therefore there is no problem with those functions that referenced libraries since in this pass they are going to be treated as functions of the contract. Since free functions are not removed from the ast, they should be ignored in this case.

RohitRanjangit (approved):

rodrigo-pino (approved):

PR #684, opened 2022-09-06
Fix rationalLiteral.ts

LuisLara-UH:
Added fraction reduction with gcd.
Refactored exp() method.
Added error when shifting fractions.


swapnilraj:
The code looks good, can you add some tests for your changes as well?

LuisLara-UH:
I added the behaviour tests for rational literals.

JorikSchellekens (approved):

PR #685, opened 2022-09-06
Abi encode warnings

rodrigo-pino:
Add warning pass and util functions with that purpose.
The warning pass currently have abi.encodePacked warnings for addresses

Try out with this contract:
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.14;

contract WARP {

    function f1(
        address a,
        address b,
        address c
    ) external pure returns (bytes memory) {
        return abi.encodePacked(a, b, c);
    }

    function middle() public pure returns (uint256) {
        uint8 a1 = 0;
        uint8 a2 = 0;
        uint8 a3 = 0;
        uint8 a4 = 0;
        uint8 a5 = 0;
        a1 = 0;
        a2 = 0;
        a3 = 0;
        a4 = 0;
        a5 = 0;
        return a1 + a2 + a3 + a4 + a5;
    }

    function f2(address a) external pure returns (bytes memory) {
        return abi.encodePacked(
            a,
            a,
            a
        );
    }
}

```


src/utils/utils.ts
```diff
@@ -484,6 +484,74 @@ export function getSourceFromLocation(source: string, location: SourceLocation):
...
 
+/**
+ * Given a source file and some nodes, prints them
+ * @param source solidity path to file
+ * @param locations nodes source locations
+ * @param chalkText function that highlight the nodes text locations
+ * @param surroundingLines lines surrounding highlighted lines[
+ * @returns text with highlights
+ */
+export function getSourceFromLocations(
```
swapnilraj, line 495: Can we use the function `getSourceFromLocation` to implement `getSourceFromLocations`, to avoid implementing the same logic twice.

We'll need to make to extract the linesAroundSource constant in `getSourceFromLocation`  into an argument and pass in a chalfText function (highlightingFunc is probably a better name) like `getSourceFomLocations`.

swapnilraj (approved):
One minor change apart from that lgtm!

src/passes/warnSupportedFeatures.ts
```diff
@@ -0,0 +1,70 @@
...
+      node.kind === FunctionCallKind.FunctionCall &&
+      node.vFunctionCallType === ExternalReferenceType.Builtin &&
+      ['encodePacked'].includes(node.vFunctionName)
+    ) {
+      node.vArguments
+        .filter((arg) => safeGetNodeType(arg, ast.compilerVersion) instanceof AddressType)
+        .forEach((arg) => this.addressesToAbiEncode.push(arg));
+    }
+
+    return this.visitExpression(node, ast);
```
cicr99, line 30: This is fine, but `commonVisit` is usually used instead. Plus, `visitExpression` does `commonVisit` on the node, so it might as well be used directly

cicr99 (approved):

src/utils/utils.ts
rodrigo-pino: Yes, you're right, actually found it easier for me to make `getSourceFromLocation` from `getSourceFromLocations`. And then realized there where very similar, so I left only the latter

PR #688, opened 2022-09-07
Optimize referenced libraries pass

cicr99:
The `ReferencedLibraries` pass adds all libraries referenced by a contract. To do so, when it encountered a function call, it would check to see if it was calling a declaration in a library, and then recursively check all function calls within that library, adding more libraries to the list. As a consequence, the resulting list might contain extra libraries, which were not actually being referenced.

The modification in the code of this pass corrects this situation. Instead of going through all the function calls inside the library, it goes through the function calls inside the referenced declaration only.

### DIFF
--- a/src/passes/referencedLibraries.ts
+++ b/src/passes/referencedLibraries.ts
@@ -16,6 +16,15 @@ import { ASTMapper } from '../ast/mapper';
 // linearizedBaselist of a contract/Library.
 
 export class ReferencedLibraries extends ASTMapper {
+  // Function to add passes that should have been run before this pass
+  addInitialPassPrerequisites(): void {
+    const passKeys: Set<string> = new Set<string>([
+      // Free Function inliner handles free functions that make library calls
+      'Ffi',
+    ]);
+    passKeys.forEach((key) => this.addPassPrerequisite(key));
+  }
+
   static map(ast: AST): AST {
     // Collect all library nodes and their ids in the map 'librariesById'
     const librariesById: Map<number, ContractDefinition> = new Map();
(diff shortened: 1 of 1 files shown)


src/passes/referencedLibraries.ts
```diff
@@ -53,45 +73,55 @@ export class ReferencedLibraries extends ASTMapper {
   }
 }
 
+function findDeclarationInLibrary(
+  calledDeclarationId: number,
+  librariesById: Map<number, ContractDefinition>,
+): [ASTNode, ContractDefinition] | undefined {
```
RohitRanjangit, line 79: Just one small suggestion,   declaration should be of type  CallableDefinition instead of generalised ASTNode

src/passes/referencedLibraries.ts
```diff
@@ -53,45 +73,55 @@ export class ReferencedLibraries extends ASTMapper {
...
-  librariesById: Map<number, ContractDefinition>,
-): ContractDefinition | undefined {
-  for (const library of librariesById.values()) {
-    if (library.vFunctions.some((f) => f.id == functionId)) return library;
-  }
-  return undefined;
-}
-
 function getAllLibraries(
+  declaration: ASTNode,
```
RohitRanjangit, line 101: ASTNode -> CallableDefinition

rjnrohit (changes requested):
lgtm, just few suggestions

rjnrohit (approved):

rodrodros (approved):
Looks good to me as well!

GuyMcComb (changes requested):
Fails: Transpile solidity
       tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/internal_library_function_bound_to_enum:

cicr99:
The PR was closed since it is not convenient to carry out such optimization.
When adding the library to the list of base contracts of a contract, all functions of that library are added to the contract in the `InheritanceInliner` pass, which means that all libraries that are referenced must be added, not just the ones that were found while reviewing the referenced declarations. In addition, any function that is not used or is not reachable from an externally visible function in the code will be removed in the `FunctionPruner` pass.

PR #691, opened 2022-09-07
Detect implicits from Cairo stubs

rodrigo-pino:
Given a function which is used as placeholder for a Cairo stub, its implicits (and the implicits of all the functions who call's it) are determined by the stub definition and not the solidity definition.

For example:
```Solidity
pragma solidity ^0.8.14;

contract C {

    ///warp-cairo
    ///func CURRENTFUNC(){bitwise_ptr : BitwiseBuiltin* }(lhs : felt, rhs : felt) -> (res : felt):
    ///from starkware.cairo.common.bitwise import bitwise_or
    ///    let (res) = bitwise_or(lhs, rhs)
    ///    return (res)
    ///end
    function f(uint8 a, uint8 b) internal pure returns (bytes32) {
        return keccak256(abi.encodePacked(a, b));
    }

    function g() external pure {
        f(10, 12);
    }
}
```
In this case `g` which interacts with `f` will have `bitwise_ptr` instead of `keccak_ptr` as its implicits after transpilation

### DIFF
--- a/src/passes/annotateImplicits.ts
+++ b/src/passes/annotateImplicits.ts
@@ -11,8 +11,10 @@ import { CairoFunctionDefinition, FunctionStubKind } from '../ast/cairoNodes';
 import { ASTMapper } from '../ast/mapper';
 import { ASTVisitor } from '../ast/visitor';
 import { printNode } from '../utils/astPrinter';
+import { TranspileFailedError } from '../utils/errors';
 import { Implicits, registerImportsForImplicit } from '../utils/implicits';
 import { isExternallyVisible, union } from '../utils/utils';
+import { getDocString, isCairoStub } from './cairoStubProcessor';
 
 export class AnnotateImplicits extends ASTMapper {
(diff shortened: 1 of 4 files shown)


src/passes/annotateImplicits.ts
```diff
@@ -129,3 +136,47 @@ class ImplicitCollector extends ASTVisitor<Set<Implicits>> {
...
+        `Implicit ${impl} defined on function stub (${printNode(node)}) is not known`,
+      );
+    }
+    result.add(impl);
+  });
+}
+
+function elementIsImplicit(e: string): e is Implicits {
+  const knownImplicits = [
+    'bitwise_ptr',
```
swapnilraj, line 170: The list of known implicits is already in `src/utils/implicits.ts` you can import from there and avoid code duplication

swapnilraj (approved):

JorikSchellekens (approved):

PR #694, opened 2022-09-08
Bug related to libraries and inheritance

cicr99:
When the set of referenced libraries was added to the list of linearized contracts of a contract, those libraries were not being added to contracts that inherit from that contract, which should happen.
Solution:
Add a method at the beginning of the inheritanceInliner pass that handles going through the list of linearized contracts of each one of the base contracts looking for all contract ids that were not added in the current linearization list, and add them


GuyMcComb:
Sem Tests Pass.

rodrigo-pino (approved):

GuyMcComb (approved):

PR #695, opened 2022-09-08
Fix interface from base contract not being written in final contract

swapnilraj:
This PR fixes the case where an interface referenced in the base contract was not being printed in the final contract


swapnilraj:
Succeeded by #699

PR #698, opened 2022-09-08
DeclarationNameMangler_audit_fixes

AlejandroLabourdette:
This pull request fix problems pointed by audit for declarationNameMangler.ts:
- Use a set for reserverdTerms to make queries faster
- Use Regex to find unsupportedCharacters in terms
- Use a unified counter to create id's for variable, internal function and new Type names
- Set a fixed width for id's in variables and internal function names to be able to sort them

Tickets:
https://www.notion.so/nethermind/AUDIT-CVF-35-declarationNameMangler-ts-de23ff0e7679484cb4f89ac58e9805aa
https://www.notion.so/nethermind/AUDIT-CVF-34-declarationNameMangler-ts-70b3c656cffc42efb6da34dce0e51701
https://www.notion.so/nethermind/AUDIT-CVF-33-declarationNameMangler-ts-1f8d615bac8a4994a0d530b97ec3e4ab
https://www.notion.so/nethermind/AUDIT-CVF-32-declarationNameMangler-ts-d69483bfcd5e4f828f1f049257354cc5

- [x] Sem Tests Running
- [x] Sem Tests Passed

### DIFF
--- a/src/passes/identifierManglerPass/declarationNameMangler.ts
+++ b/src/passes/identifierManglerPass/declarationNameMangler.ts
@@ -65,23 +65,23 @@ export function checkSourceTerms(term: string, node: ASTNode) {
   }
 
   // Creating the regular expression that match unsupportedCharacters
-  const regex_str = unsupportedCharacters.reduce(
+  const regexStr = unsupportedCharacters.reduce(
     (prevVal, val, i) => (i === 0 ? '\\' + val : (prevVal += '|\\' + val)),
     '',
   );
   // Looking for possible matches
-  const regex = RegExp(regex_str, 'g');
+  const regex = RegExp(regexStr, 'g');
   let match;
-  let unsupported_characters_found = '';
+  let unsupportedCharactersFound = '';
   while ((match = regex.exec(term)) !== null) {
     // Saving all chars founded
-    unsupported_characters_found += match[0];
+    unsupportedCharactersFound += match[0];
   }
-  if (unsupported_characters_found) {
+  if (unsupportedCharactersFound) {
     throw new WillNotSupportError(
       `${printNode(
         node,
-      )} ${term} contains unsupported character(s) "${unsupported_characters_found}"`,
+      )} ${term} contains unsupported character(s) "${unsupportedCharactersFound}"`,
     );
   }
 }
@@ -187,7 +187,7 @@ export class DeclarationNameMangler extends ASTMapper {
     if (match === null) {
       throw new TranspileFailedError(`Expected ${pattern} node name: ${name}`);
     }
-    const new_name = `${pattern}${match[1].padStart(lastIdSize, '0')}`;
-    return name.replace(new RegExp(`${pattern}([0-9]+)`), new_name);
+    const newName = `${pattern}${match[1].padStart(lastIdSize, '0')}`;
+    return name.replace(new RegExp(`${pattern}([0-9]+)`), newName);
   }
 }
--- a/src/passes/identifierManglerPass/declarationNameMangler.ts
+++ b/src/passes/identifierManglerPass/declarationNameMangler.ts
@@ -11,7 +11,7 @@ import { ABIEncoderVersion } from 'solc-typed-ast/dist/types/abi';
 import { AST } from '../../ast/ast';
 import { ASTMapper } from '../../ast/mapper';
 import { printNode } from '../../utils/astPrinter';
-import { WillNotSupportError } from '../../utils/errors';
+import { WillNotSupportError, TranspileFailedError } from '../../utils/errors';
 import {
   MANGLED_INTERNAL_USER_FUNCTION,
   MANGLED_LOCAL_VAR,
@@ -88,6 +88,10 @@ export function checkSourceTerms(term: string, node: ASTNode) {
 
 export class DeclarationNameMangler extends ASTMapper {
   lastUsedId = 0;
+  // Feel free to increase this value. The greater the value the less probably to
+  // execute the code to fix Id's width, but less legible variable names would be.
+  initialIdWidth = 2;
+  nodesNameModified: ASTNode[] = [];
 
   // This strategy should allow checked demangling post transpilation for a more readable result
   createNewExternalFunctionName(fd: FunctionDefinition): string {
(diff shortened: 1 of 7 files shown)


src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -85,17 +100,22 @@ export class DeclarationNameMangler extends ASTMapper {
...
   }
 
   createNewTypeName(existingName: string): string {
-    return `${MANGLED_TYPE_NAME}${this.lastUsedTypeId++}_${existingName}`;
+    return `${MANGLED_TYPE_NAME}_${this.getFormattedId()}_${existingName}`;
   }
 
   createNewVariableName(existingName: string): string {
-    return `${MANGLED_LOCAL_VAR}${this.lastUsedVariableId++}_${existingName}`;
+    return `${MANGLED_LOCAL_VAR}_${this.getFormattedId()}_${existingName}`;
```
rodrigo-pino, line 118: Does the extra '_' change to naming should go here or in the constants (i.e `MANGLED_LOCAL_VARS` ...)?. Also, I think it would be wise if these name mangling constants could be used as well in `CairoStubProcessor` pass regex expressions to have a single source of truth and avoid future headaches.

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -55,28 +55,43 @@ export const reservedTerms = [
...
+  // Creating the regular expression that match unsupportedCharacters
+  let regex_str = '\\' + unsupportedCharacters[0];
+  for (let index = 1; index < unsupportedCharacters.length; index++) {
+    regex_str += '|\\' + unsupportedCharacters[index];
+  }
+  // Looking for possible matches
+  const regex = RegExp(regex_str, 'g');
+  let match;
+  let unsupported_characters_found = '';
+  while ((match = regex.exec(term)) !== null) {
```
rodrigo-pino, line 76: Does `term` should get updated inside the body of the loop?

AlejandroLabourdette: The goal is to find all occurrence of unsupported characters in that term, the flag 'g' when created the regex do that. When no other match is found the regex.exec(term) returns null. Term is the string where the look is made. Why would you need it to change?

src/passes/identifierManglerPass/declarationNameMangler.ts
AlejandroLabourdette: That's a good idea.

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -141,5 +163,31 @@ export class DeclarationNameMangler extends ASTMapper {
...
+          node.name = this.updateName(node.name, MANGLED_LOCAL_VAR, lastIdSize);
+        } else {
+          throw new TranspileFailedError(`Not expected node to update name: ${node}`);
+        }
+      });
+    }
+    this.nodesNameModified = []; // Lose all references to avoid consuming unnecessary memory
+  }
+
+  // Set Id in a node name 'name' that fallows the pattern 'pattern' to a fixed width
```
rodrigo-pino, line 184: I am not sure I understand this commnet. fallows -> allows?

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -141,5 +163,31 @@ export class DeclarationNameMangler extends ASTMapper {
...
     this.commonVisit(node, ast);
+
+    // Checking if counter is greater than initialIdWidth digits. If so, names are
+    // modified to insert 0's to achieve id's fixed width.
+    const lastIdSize = this.lastUsedId.toString().length;
+    if (lastIdSize > this.initialIdWidth) {
+      this.nodesNameModified.forEach((node) => {
+        if (node instanceof FunctionDefinition) {
+          node.name = this.updateName(node.name, MANGLED_INTERNAL_USER_FUNCTION, lastIdSize);
+        } else if (node instanceof VariableDeclaration) {
```
rodrigo-pino, line 174: Does the `MANGLED_INTERNAL_USER_FUNCTION` should be used in both public and private functions?

src/utils/nameModifiers.ts
```diff
@@ -12,9 +12,9 @@ export const FREE_FILE_SUFFIX = '__WC_FREE';
 export const CONTRACT_INFIX = '__WC__';
 
 // Used in IdentifierManglerPass
```
rodrigo-pino, line 14: Could you add `CairoStubProcessor` here as a possible location

src/passes/identifierManglerPass/declarationNameMangler.ts
AlejandroLabourdette: Those node names can start with "__warp_usrfn_" or "__warp_usrid_". Let's say they are pattern1 and pattern2. If node name starts with "__warp_usrfn_" it fallows a pattern (pattern1 in this case). Then the id substring in that name is extended with 0's to get a specific length. The way to get the id substring is with regex using that pattern.

src/passes/identifierManglerPass/declarationNameMangler.ts
AlejandroLabourdette: No, that condition is over the `nodesNameModified` elements. And the only nodes of type `FunctionDefinition` added to that list are internal functions.

src/utils/nameModifiers.ts
AlejandroLabourdette: Ok.

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -85,17 +100,22 @@ export class DeclarationNameMangler extends ASTMapper {
       : fd.name;
   }
 
+  // Return a new id formatted to achieve the minimum lenght
```
GuyMcComb, line 103: typo

GuyMcComb:
Sem Tests Passed

cicr99 (approved):

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -55,28 +55,43 @@ export const reservedTerms = [
...
-  });
+  // Creating the regular expression that match unsupportedCharacters
+  const regex_str = unsupportedCharacters.reduce(
+    (prevVal, val, i) => (i === 0 ? '\\' + val : (prevVal += '|\\' + val)),
+    '',
+  );
+  // Looking for possible matches
+  const regex = RegExp(regex_str, 'g');
+  let match;
+  let unsupported_characters_found = '';
```
swapnilraj, line 75: Camel case for unsupported_chraracters_found

swapnilraj (changes requested):

src/passes/identifierManglerPass/declarationNameMangler.ts
swapnilraj: same for regex_str

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -141,5 +163,31 @@ export class DeclarationNameMangler extends ASTMapper {
...
+    this.nodesNameModified = []; // Lose all references to avoid consuming unnecessary memory
+  }
+
+  // Set Id in a node name 'name' that follows the pattern 'pattern' to a fixed width
+  updateName(name: string, pattern: string, lastIdSize: number) {
+    const match = new RegExp(`${pattern}([0-9]+)`).exec(name);
+    if (match === null) {
+      throw new TranspileFailedError(`Expected ${pattern} node name: ${name}`);
+    }
+    const new_name = `${pattern}${match[1].padStart(lastIdSize, '0')}`;
```
swapnilraj, line 190: camel case for new_name

rodrigo-pino (approved):

PR #699, opened 2022-09-09
fix interface being not carried out to final contract

RohitRanjangit:
This PR fixes the case where an interface referenced in the base contract was not being printed in the final contract, an alternate approach


RohitRanjangit:
#695

GuyMcComb (approved):

cicr99 (approved):

PR #700, opened 2022-09-09
test to see what happens

AlejandroLabourdette:
Is just to get tested the project with a first pass of expressionSplitter

### DIFF
--- a/src/transpiler.ts
+++ b/src/transpiler.ts
@@ -99,6 +99,7 @@ export function transform(ast: AST, options: TranspilationOptions & PrintOptions
 
 function applyPasses(ast: AST, options: TranspilationOptions & PrintOptions): AST {
   const passes: Map<string, typeof ASTMapper> = createPassMap([
+    ['E1', ExpressionSplitter],
     ['Tf', TupleFixes],
     ['Tnr', TypeNameRemover],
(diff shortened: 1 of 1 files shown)


src/transpiler.ts
```diff
@@ -99,6 +99,7 @@ export function transform(ast: AST, options: TranspilationOptions & PrintOptions
 
 function applyPasses(ast: AST, options: TranspilationOptions & PrintOptions): AST {
   const passes: Map<string, typeof ASTMapper> = createPassMap([
+    ['E1', ExpressionSplitter],
```
GuyMcComb, line 102: @cicr99 which pass would be best to move this after?

JorikSchellekens:
@AlejandroLabourdette I should have been more careful about how I detailed this. Can you put it after the tuple fixes pass?

src/transpiler.ts
AlejandroLabourdette: I keep looking it still prints errors I'm trying to see why

AlejandroLabourdette:
> @AlejandroLabourdette I should have been more careful about how I detailed this. Can you put it after the tuple fixes pass?
@JorikSchellekens one of the first errors was that the conversion that is done in that pass is necessary before. I'm working from there

PR #702, opened 2022-09-12
Transpile libraries

LuisLara-UH:
The idea here is to clone the ContractDefinition nodes of kind Library, and change its kind to Contract. This way, it does not interfere with the references to the Library node functions, and it transpiles the Library clone as a contract.


GuyMcComb:
Could you add a test?

src/passes/librariesConverter.ts
```diff
@@ -0,0 +1,41 @@
...
+      contractNode.kind = ContractKind.Contract;
+      this.externalizeFunctions(contractNode);
+      (node.parent as ASTNodeWithChildren<ASTNode>).appendChild(contractNode);
+    }
+  }
+
+  externalizeFunctions(node: ContractDefinition): void {
+    node.children.forEach((ch) => {
+      if (ch instanceof FunctionDefinition) {
+        (ch as FunctionDefinition).visibility = FunctionVisibility.External;
```
cicr99, line 37: Is it necessary to change the visibility of every function, or just the ones that are not externally visible? I would expect public functions to remain that way

src/utils/cloning.ts
```diff
@@ -480,6 +502,30 @@ function cloneASTNodeImpl<T extends ASTNode>(
       [...node.vParameters].map((p) => cloneASTNodeImpl(p, ast, remappedIds)),
       node.raw,
     );
+  } else if (node instanceof ContractDefinition) {
```
cicr99, line 505: I think declarations such as `ContractDefinition` and `EnumDefinition` should go in the previous section with the other ones, and leave `ParameterList` alone in `//ASTNodeWithChildren`

src/utils/nameModifiers.ts
```diff
@@ -7,6 +7,9 @@
 // Used in TupleFiller in TupleFixes
 export const TUPLE_FILLER_PREFIX = '__warp_tf';
 
+// Used in librariesConverter
+export const LIBRARY_CONTRACT_INFIX = 'library_';
```
cicr99, line 11: This is actually used as a prefix, should consider changing the name to `LIBRARY_CONTRACT_PREFIX`

src/passes/librariesConverter.ts
```diff
@@ -0,0 +1,41 @@
...
+    passKeys.forEach((key) => this.addPassPrerequisite(key));
+  }
+
+  visitContractDefinition(node: ContractDefinition, ast: AST): void {
+    if (node.kind === ContractKind.Library) {
+      const contractNode = cloneASTNode(node, ast);
+      contractNode.name = `${LIBRARY_CONTRACT_INFIX}${node.name}`;
+      contractNode.kind = ContractKind.Contract;
+      this.externalizeFunctions(contractNode);
+      (node.parent as ASTNodeWithChildren<ASTNode>).appendChild(contractNode);
```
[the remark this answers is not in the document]
LuisLara-UH: There I am cloning a Contract Definition Node. I thought this kind of node always has a SourceUnit parent node. But I could add some verification there if you want.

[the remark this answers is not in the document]
cicr99: You could use `getContainingSourceUnit` to do that verification. This method already contains an assertion.

src/passes/librariesConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (
+        ch instanceof FunctionDefinition &&
+        (ch.visibility === FunctionVisibility.Internal ||
+          ch.visibility === FunctionVisibility.Private)
+      ) {
+        ch.visibility = FunctionVisibility.External;
+      }
+    });
+  }
+}
```
rodrigo-pino, line 44: I don't understand why you're changing the new contract's visibility functions. I think they are meant to be private; by externalizing them security flaws can be created on the transpiled contract

LuisLara-UH: If I keep those functions private or internal, somewhere in the passes they get removed from the contract. As this contract is a cloned node, its components are not referenced by other nodes. I am not sure about what security flaws could be created.

Also, if I want to keep those functions internal or private, what can I do for they don't get removed?

rodrigo-pino: Function pruner removes all functions not visible from the outside. If this functions are private and internal and cannot be call by any form from an external function (directly or indirectly) then they are deleted. If you want them not to be deleted you need to design an external function which calls them.

LuisLara-UH:
I removed the externalize-functions, because this is only for transpile the internal and private functions which aren't referenced by public or external functions. After discussed with Rodrigo, I realized these functions are not necessary for the transpiled library, because they are not reachable.

rodrigo-pino (approved):

cicr99 (approved):

swapnilraj:
Are we waiting for a semantic test run or can we merge this?

PR #703, opened 2022-09-12
Fix member access interfaces bug

JorikSchellekens:
I can't seem to construct a minimal case which replicates this bug however it occurred as a result of transpiling https://github.com/Uniswap/v3-core/blob/main/contracts/test/TestUniswapV3Router.sol

This also fixes the issue that constructors were being added to interfaces.

### DIFF
--- a/src/passes/externalContractHandler/externalContractInterfaceInserter.ts
+++ b/src/passes/externalContractHandler/externalContractInterfaceInserter.ts
@@ -15,6 +15,7 @@ import { AST } from '../../ast/ast';
 import { CairoContract } from '../../ast/cairoNodes';
 import { ASTMapper } from '../../ast/mapper';
 import { cloneASTNode } from '../../utils/cloning';
+import { TranspileFailedError } from '../../utils/errors';
 import { safeGetNodeType } from '../../utils/nodeTypeProcessing';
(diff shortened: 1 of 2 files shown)


src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -47,8 +48,13 @@ export class ExternalContractInterfaceInserter extends ASTMapper {
...
     const nodeType = safeGetNodeType(node.vExpression, ast.compilerVersion);
     if (nodeType instanceof UserDefinedType && nodeType.definition instanceof ContractDefinition) {
+      // nodeType.definition will get the old ContractDefinition, see the note
+      // in storageAllocator on hotfixing getNodeType
+      // To fix this we lookup the contract definition id in the context
+      if (node.context === undefined)
+        throw new TranspileFailedError('Node context should not be undefined');
       importExternalContract(
-        nodeType.definition,
+        node.context.locate(nodeType.definition.id) as ContractDefinition,
```
RohitRanjangit, line 57: I was not aware of this hotfix/problem with getNodeType, when I made the PR #699

RohitRanjangit (approved):

GuyMcComb (approved):

PR #711, opened 2022-09-19
Hop bridge for Cairo 0.10

rodrodros:


rodrigo-pino:
Compiles succesfully, only missing part is Connector, which should be a simple contract made in native Cairo which will be the responsible of inter-blockchain message passing.

PR #714, opened 2022-09-26
Fix class hash zero bug

swapnilraj:
The class_hash was computed after the compile so the compiled cairo file had the class_hash as zero

### DIFF
--- a/src/utils/postCairoWrite.ts
+++ b/src/utils/postCairoWrite.ts
@@ -17,12 +17,13 @@ export const HASH_OPTION = 'sha256';
   @param pathPrefix: The prefix proceeding the the path to the file
   @param contractHashToClassHash: A mapping that holds the contract path with out the pathPrefix and maps
   it to the contracts class hash.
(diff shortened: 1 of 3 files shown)


rodrigo-pino (approved):

src/utils/postCairoWrite.ts
```diff
@@ -16,29 +16,29 @@ export const HASH_OPTION = 'sha256';
   @param cairoFilePath: The path to the cairo File being processed.
   @param pathPrefix: The prefix proceeding the the path to the file
   @param contractHashToClassHash: A mapping that holds the contract path with out the pathPrefix and maps
-  it to the contracts class hash.
+  @returns files: A list of files that were edited to include the class_hash constant it to the contracts class hash.
```
rodrigo-pino, line 19: Minor fix to the comments here

swapnilraj: Fixed

GuyMcComb (approved):

GuyMcComb:
No changing of Core compiler logic, skipping CI.

PR #724, opened 2022-10-05
Split Declare and Deploy in CLI

RohitRanjangit:


GuyMcComb:
Fails locally when StarkNet CLI tries to simulate the transaction. Where we were getting a local nonce less than the expected nonce on StarkNet we are now getting nonce too high compared to the local StarkNet implementation.

swapnilraj:
The README examples need to be updated to take this into account

readme.md
```diff
@@ -145,16 +145,24 @@ warp deploy_account
...
 ```bash
 warp transpile <path to Solidity contract>
 ```
 
-To deploy a Cairo contract:
+To declare a StarkNet contract:
 
 ```bash
-warp deploy <path to Cairo contract>
+warp declare <path to StarkNet contract>
```
swapnilraj, line 157: Sorry to be annoying but can you put the warp deploy_account and stuff as well
We should have intstructions that if a user follows them verbatim they should be able to deploy

rodrigo-pino (commented):
* Not specifically related to this PR but taking advantage on the work on cli. There is no longer a need to compile contract  dependencies when doing deploys, where dependencies were contract that were deployed using the `deploy` sys call and needed to be compiled to know their declaration hash, but this is already executed during  warp transpilation. Could you remove that part (at line 133). It should be replaced by a normal cairoCompile call.

* Also (I write it here because I cannot mark it), since our users are not top devs, I would suggest to make a small parser that detects when `warp deploy` errors because of the use of a nonexistent(or yet to be declared) class hash, and explicitly ask for them to call `warp declare ...` first

readme.md
GuyMcComb: It is mentioned above. Can we make another PR for ReadME changes?

rodrigo-pino (approved):

GuyMcComb:
Does not touch core logic therefor skipping semantic tests

GuyMcComb (approved):

PR #726, opened 2022-10-05
Add missing implicit in abi encode

swapnilraj:


GuyMcComb (approved):

rodrigo-pino (approved):

GuyMcComb:
I feel we don't need to run sem tests on this one

PR #744, opened 2022-10-16
Avoid retranspilation

rodrigo-pino:
Optimization to avoid retranspilation of files when excuting 
```bash
warp transpile <file_list>
```
Previous behaviour:
```bash
warp transpile x.sol y.sol
```
if x.sol imports y.sol, then during `x` transpilation, `y` would get transpiled as it's a dependency
Then `y` would get transpiled (again!)

Current behaviour:
`y` is transpiled only once, during the process of the transpilation of `x`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -87,7 +87,7 @@ program
         const otherFiles = new Set<string>(
           otherAST.roots.map((sourceUnit) => sourceUnit.absolutePath),
         );
-        return files.every(otherFiles.has);
+        return files.every((f) => otherFiles.has(f));
       });
     });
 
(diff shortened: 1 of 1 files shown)


JorikSchellekens (approved):

JorikSchellekens:
Nice, this was on my todo! What would result in even less redundancy would be to flatten all the ASTs into one somehow (taking care of scopes) so that everything only gets included exactly once.

rodrigo-pino:
> Nice, this was on my todo! What would result in even less redundancy would be to flatten all the ASTs into one somehow (taking care of scopes) so that everything only gets included exactly once.

Yes that would be optimal. I thought of merging the ASTs but that could get messy. Other more hacky way I thought of is making (or simulating the existence of) a sol file which  only imports the files to transpile and then compiling it, leaving `solc-typed-ast` do all the work

PR #745, opened 2022-10-19
Add reference types for abi decoding

rodrigo-pino:


src/cairoUtilFuncGen/abi/base.ts
```diff
@@ -45,3 +47,22 @@ export abstract class AbiBase extends StringIndexedFuncGenWithAuxiliar {
...
+ *    uint8[20] -> uint8[]
+ *    uint[][8] -> uint[][]
+ *    uint[10][15] -> uint[10][]
+ *  @param type a static ArrayType
+ *  @returns static array without length information
+ */
+export function removeSizeInfo(type: ArrayType): string {
+  assert(type.size !== undefined, 'Expected an ArrayType with known type (a solc static array)');
+  const typeString = type.pp();
+  const reversedTypeString = typeString.split('').reverse().join('');
```
swapnilraj, line 64: `stringVal.replaceAll(/\[\d+\]/g, '[]')`
Why not something like this ^?

swapnilraj (changes requested):

warplib/dynamic_arrays_util.cairo
```diff
@@ -108,6 +108,18 @@ func byte_array_to_uint256_value{
...
 }
 
+// Unsafe copy of elements from an array to the other from index to last_index
+func memory_dyn_array_copy{bitwise_ptr: BitwiseBuiltin*, range_check_ptr, warp_memory: DictAccess*}(
+    a_index: felt, a_ptr: felt, b_index: felt, b_last_index: felt, b_ptr: felt
+) {
+    if (b_index == b_last_index) {
+        return ();
+    }
+    let (elem) = dict_read{dict_ptr=warp_memory}(a_ptr + 2 + a_index);
```
swapnilraj, line 118: is the constant offset of 2 to skip the length encoding?

src/cairoUtilFuncGen/abi/abiDecode.ts
```diff
@@ -129,21 +147,42 @@ export class AbiDecode extends StringIndexedFuncGenWithAuxiliar {
 
   public getOrCreateDecoding(type: TypeNode): string {
     const unexpectedType = () => {
-      throw new TranspileFailedError(`Encoding ${printTypeNode(type)} is not supported yet`);
+      throw new TranspileFailedError(`Encoding of ${printTypeNode(type)} is not valid`);
```
swapnilraj, line 150: encoding -> decoding

src/cairoUtilFuncGen/abi/base.ts
rodrigo-pino: I made this to have equal typestrings from static arrays, that receive the length as an input so we can use the same generated function for both

rodrigo-pino: The purpose of the function is to replace the length from static arrays.
So that generated functions for T[x] && T[y] could be the same and the length is used as a paremeter. This way we avoid the generation of similar functions with just a single value change (x -> y)

I do all the reversing and stuff so that I only take the last one:
T[z][x] -> T[z][].

I don't want to remove them all because I still need to keep the static length for nested types. So that the generated function of T[z][] can call to the generated function of T[z] (which will be T[]) but with size `z`.

rodrigo-pino: The purpose of the function is to replace the length from static arrays.
So that generated functions for T[x] && T[y] could be the same and the length is used as a paremeter. This way we avoid the generation of similar functions with just a single value change (x -> y)

I do all the reversing and stuff so that I only take the last one:
T[z][x] -> T[z][].

I don't want to remove them all because I still need to keep the static length for nested types. So that the generated function of T[z][] can call to the generated function of T[z] (which will be T[]) but with size `z`.

warplib/dynamic_arrays_util.cairo
rodrigo-pino: Yes

src/cairoUtilFuncGen/abi/base.ts
swapnilraj: I see it makes sense

swapnilraj (approved):

swapnilraj:
Can you update the readme, the unsupported features list

src/cairoUtilFuncGen/abi/abiEncode.ts
```diff
@@ -342,15 +342,7 @@ export class AbiEncode extends AbiBase {
...
-    let key = 'inline ' + type.pp();
-    // Modifiy key to generate same function for diferent size (but same element type)
-    // static arrays
-    if (type.size !== undefined) {
-      const reversedKey = key.split('').reverse().join('');
-      // swapping '[<num>]' for '[]' but since string is reversed we are swapping ']<num>[' for ']['
-      const parsedKey = reversedKey.replace(/\][0-9]+\[/, '][');
-      key = parsedKey.split('').reverse().join('');
-    }
+    const key = 'inline ' + (type.size !== undefined ? type.pp() : removeSizeInfo(type));
```
cicr99, line 345: In order to execute the function call `removeSizeInfo(type)`, the condition `type.size !== undefined` must be false. Then the first statement of this function is an assert with `type.size !== undefined` as the condition, which means that everytime this function call is executed, it throws an assertion error. Is this what was intended? I'm asking because that's not what was happening in the code that was removed.

tests/behaviour/expectations/semantic_whitelist.ts
```diff
@@ -678,7 +678,7 @@ const tests: string[] = [
       // 'tests/behaviour/solidity/test/libsolidity/semanticTests/functionCall/gas_and_value_basic.sol', // WILL NOT SUPPORT gas
       // 'tests/behaviour/solidity/test/libsolidity/semanticTests/functionCall/gas_and_value_brace_syntax.sol', // WILL NOT SUPPORT gas
       // 'tests/behaviour/solidity/test/libsolidity/semanticTests/functionCall/member_accessors.sol', // WILL NOT SUPPORT keccak
-      // 'tests/behaviour/solidity/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol', // WILL NOT SUPPORT precompile ext call. abi.decode
+      // 'tests/behaviour/solidity/test/libsolidity/semanticTests/functionCall/precompile_extcodesize_check.sol', WILL NOT SUPPORT  member of addresses
```
cicr99, line 681: The `//` is missing before `WILL NOT SUPPORT ...`

src/cairoUtilFuncGen/abi/abiEncode.ts
rodrigo-pino: Good find, I overlooked. This wouldn't make the code to break but to generate extra functions

src/cairoUtilFuncGen/abi/base.ts
```diff
@@ -45,3 +47,22 @@ export abstract class AbiBase extends StringIndexedFuncGenWithAuxiliar {
...
+  // swapping '[<num>]' for '[]' but since string is reversed we are swapping ']<num>[' for ']['
+  const parsedTypeStting = reversedTypeString.replace(/\][0-9]+\[/, '][');
+  return parsedTypeStting.split('').reverse().join('');
```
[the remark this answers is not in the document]
rodrigo-pino: Yes, the code can be improved. I took a third approach

cicr99 (approved):
LGTM

PR #746, opened 2022-10-19
Add devnet

JorikSchellekens:
The new Dockerfile is compatible with the [old single container setup](https://github.com/NethermindEth/warp/tree/devnet-docs#docker-whale) so the instructions doesn't need to change

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,8 +32,8 @@ It's best to copy the contract/repo to the warp directory so it is available in 
 ### Deploy to devnet
 
 ```bash
-docker-compose exec warp starknet-compile warp_output/example__contracts/ERC20__WC__WARP.cairo --out erc20.json --abi erc20.abi.json
-docker-compose exec warp starknet deploy --no_wallet --contract erc20.json --gateway_url http://devnet:5050
+docker-compose exec warp warp compile warp_output/example__contracts/ERC20__WC__WARP.cairo
+docker-compose exec warp starknet deploy --no_wallet --contract warp_output/example__contracts/ERC20__WC__WARP_compiled.json --gateway_url http://devnet:5050
 ```
 
 ## Documentation 📖


swapnilraj (approved):

USING-DEVNETS.md
```diff
@@ -0,0 +1,30 @@
...
+
+export STARKNET_GATEWAY_URL=http://localhost:5050
+export STARKNET_FEEDER_GATEWAY_URL=htp://localhost:5050
+export STARKNET_ACCOUNT_DIR=~/.starknet_accounts_devnet
+```
+
+It's a good idea to then launch the devnet with the following
+
+```
+starknet-devent --timeout 10000 --seed 0
```
piwonskp, line 15: Typo

USING-DEVNETS.md
```diff
@@ -0,0 +1,30 @@
+Here are some useful env variables to have set if you are targeting a local StarkNet devnet
+
+```
+export STARKNET_NETWORK=alpha-goerli
+export STARKNET_WALLET=starkware.starknet.wallets.open_zeppelin.OpenZeppelinAccount
+
+export STARKNET_GATEWAY_URL=http://localhost:5050
```
swapnilraj, line 7: Its just GATEWAY_URL and FEEDER_GATEWAY_URL no prefix neede

swapnilraj:
Took the liberty to address the comments

USING-DEVNETS.md
```diff
@@ -0,0 +1,17 @@
+To use starknet devnet you have to export evironment variables first:
+
+```bash
+export $(cat development.env | xargs)
```
swapnilraj, line 4: `source development.env`

development.env
```diff
@@ -0,0 +1,7 @@
+STARKNET_NETWORK=alpha-goerli
+STARKNET_WALLET=starkware.starknet.wallets.open_zeppelin.OpenZeppelinAccount
+STARKNET_ACCOUNT_DIR=~/.starknet_accounts_devnet
```
swapnilraj, line 3: If we set `STARKNET_ACCOUNT_DIR=$PWD` then we will not need to copy the account json file and this way we also avoid copying over the original account json

USING-DEVNETS.md
piwonskp: Source creates shell variables instead of env variables as source expects the file to be a shell script

piwonskp: Ok, there is however an option to do it with source and still preserve the .env syntax

readme.md
```diff
@@ -5,6 +5,37 @@
...
+
+Alternatively for warp developers:
+
+```bash
+docker-compose exec warp npx ts-node src transpile example_contracts/ERC20.sol
+```
+
+It's best to copy the contract/repo to the warp directory so it is available in container via volume. Use contract's paths relative to warp root. For example, assuming you've copied your project to `warp/projects/myproject` you can replace `example_contracts/ERC20.sol` with `projects/myproject/mycontract.sol` in the above commands.
+
+### Deploy to devnet
```
swapnilraj, line 32: We should move this to the dev section, there is a command for compiling with warp
`warp compile` we should use that.

In a follow up PR can you also make it possible to deploy to the devnet using warp deploy and then replace starknet deploy with warp deploy

piwonskp: Not sure what you mean by dev section. If [that](https://github.com/NethermindEth/warp#warp-installation-method-2-from-sourcefor-devs) then the section is not describing docker-compose usage

PR #748, opened 2022-10-25
Add note - env variables not working

piwonskp:


USING-DEVNETS.md
```diff
@@ -1,4 +1,5 @@
 Here are some useful env variables to have set if you are targeting a local StarkNet devnet
+> **_NOTE:_**  `GATEWAY_URL` and `FEEDER_GATEWAY_URL` don't work (Cairo 0.10.0, 0.10.1) and therefore have to be passed via command line arguments.
```
swapnilraj, line 2: We should just remove the exported variables if they don't work?
To confirm neither GATEWAY_URL nor STARKNET_GATEWAY_URL work right?

piwonskp:
The work is continued on #746

PR #749, opened 2022-10-25
Add format flag

rodrodros:


GregVardz (commented):
Please change the flag to '--format-cairo' instead of the inverse. I'd rather have people explicitly opt in instead of opt-out.

swapnilraj:
Let's also enable formatting with the dev flag as well

src/index.ts
```diff
@@ -58,6 +59,9 @@ program
   .option('--compile-cairo')
   .option('--no-compile-errors')
   .option('--check-trees')
+  // for development mode
+  .option('--dev', 'Run AST sanity checks on every pass instead of the final AST only', false)
+  .option('--no-format-cairo', "Don't format cairo output")
```
cicr99, line 64: Use `'` instead of `"` for the option message in order to keep the same format.

rodrigo-pino: You're right, the only reason it's different is because the use of a `'` inside the string were the formatter prefer "Don't" against 'Don\'t'.  Nonetheless when changed form `no-format-cairo` to `format-cairo` this will be solved

cicr99: 👌

swapnilraj (approved):

swapnilraj:
@rodrigo-pino can you resolve the conflicts and then we can merge

rodrigo-pino:
Closed because the addition here where merged with PR #755

PR #751, opened 2022-10-27
Add include-path option to execute warp commands

AlejandroLabourdette:
This PR is to allow executing warp commands specifying compiler options for solc to execute with. So far are included:
- include-paths
- base-path

This allow to solve the error importing libraries with @. You can simply include the directory were the library is placed (most scenarios node_modules folder).


src/solCompile.ts
```diff
@@ -97,13 +102,19 @@ function cliCompile(
...
+
   return {
     result: JSON.parse(
-      execSync(`${solcCommand} --standard-json ${allowedPaths}`, {
-        input: JSON.stringify(input),
-        maxBuffer: MAX_BUFFER_SIZE,
-        stdio: ['pipe', 'pipe', 'ignore'],
-      }).toString(),
+      execSync(
+        `${solcCommand} --standard-json ${allowedPaths} ${includePathOptions} --base-path .`,
```
JorikSchellekens, line 111: We should also just add base path as an option. Actually can you do an overview of all the options solc provides and make sure all the useful one's are threaded through correctly?

JorikSchellekens (changes requested):

src/solCompile.ts
```diff
@@ -82,6 +83,7 @@ function formatInput(fileName: string): SolcInput {
 function cliCompile(
   input: SolcInput,
   solcVersion: string,
+  options?: CompilationOptions,
```
swapnilraj, line 86: Why is options optional here?

AlejandroLabourdette: For compatibility, we call that in *compileSolFileAndExtractContracts* function. I kind of didn't like the idea of modifying the call to pass some empty object there, so in cliCompile if nothing is passed then it just add no option to the solc exec command

swapnilraj (approved):

PR #752, opened 2022-10-27
User defined identifiers are not mangled

cicr99:
All user defined names now have a common prefix


src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -87,46 +83,38 @@ export function checkSourceTerms(term: string, node: ASTNode) {
...
+    this.commonVisit(node, ast);
   }
 
   visitStructDefinition(_node: StructDefinition, _ast: AST): void {
-    // struct definitions should already have been mangled at this point
-    // by visitContractDefinition and visitSourceUnit
+    // struct definitions should have already been mangled at this point
+    // by visitSourceUnit and mangleContractDefinition
     return;
   }
```
rodrigo-pino, line 98: If this function does nothing, I think it is safe to erase it. I don't think it affects readability because by non existing you can infer structs mangling get handled somewhere else.

swapnilraj (changes requested):

rodrodros (commented):
Does the rejectPass is necessary, since everything will get prefixed with `__warp_` even vars which are named `__warp_` resulting in `__warp___warp_` and no naming clash is possible?

src/transpiler.ts
```diff
@@ -171,6 +172,9 @@ function applyPasses(ast: AST, options: TranspilationOptions & PrintOptions): AS
   printPassName('Input', options);
   printAST(ast, options);
 
+  // Reject code that contains identifiers starting with certain patterns
+  ast = RejectPrefix.map(ast);
```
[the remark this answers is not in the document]
swapnilraj: The idea is that later we will not mangle user defined identifiers, that's why need to reject the prefixes. This is all so that we improve readability of generated code.
It needs to before the passes because the ast sanity checks would fail if there is a pass before the tuple fixes
Yea we don't need the assignment

src/passes/identifierManglerPass/declarationNameMangler.ts
cicr99: Well, if you remove this function, then it will execute the code of the `visitStructDefinition` function in the base class, which does `commonVisit` on its children, so at the end, the node would be visited again and thus the mangling will be done again

src/transpiler.ts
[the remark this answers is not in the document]
cicr99: `TupleFixes` passes handles things such as: `(a, ) = (3, s[7][]);`. When the Sanity Checks are executed, they use `getNodeType` method which fails on this expression `(a, )` in `solc-typed-ast`. This means that adding `RejectPrefix` pass before `TupleFixes` will cause the Sanity Checks to fail. But still, `RejectPrefix` pass should be run first because `TupleFixes` already generates new variable names with the prefix `__warp_`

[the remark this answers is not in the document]
JorikSchellekens: How much work would it be to fix this in solc-typed-ast?

src/passes/identifierManglerPass/declarationNameMangler.ts
cicr99: With the last changes made, it's true that this function is no longer necessary, so I removed it as suggested.

src/passes/rejectPrefix.ts
```diff
@@ -0,0 +1,67 @@
...
+  FunctionDefinition,
+  StructDefinition,
+  VariableDeclaration,
+} from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+import { getErrorMessage, WillNotSupportError } from '../utils/errors';
+import { MANGLED_WARP } from '../utils/nameModifiers';
+
+export class RejectPrefix extends ASTMapper {
```
swapnilraj, line 14: can you add an export for RejectPrefix in the export file

src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -87,47 +83,46 @@ export function checkSourceTerms(term: string, node: ASTNode) {
...
-  lastUsedId = 0;
-  // Feel free to increase this value. The greater the value the less probably to
-  // execute the code to fix Id's width, but less legible variable names would be.
-  initialIdWidth = 2;
-  nodesNameModified: ASTNode[] = [];
-
-  // This strategy should allow checked demangling post transpilation for a more readable result
-  createNewFunctionName(fd: FunctionDefinition, ast: AST): string {
-    return !isNameless(fd) ? `${fd.name}_${safeCanonicalHash(fd, ast)}` : fd.name;
+  nameCounter = 0;
```
swapnilraj, line 86: nit: newline

src/cairoUtilFuncGen/memory/memoryStruct.ts
```diff
@@ -84,12 +84,12 @@ export class MemoryStructGen extends StringIndexedFuncGen {
       code: [
         `func ${funcName}{range_check_ptr, warp_memory: DictAccess*}(${argString}) -> (res:felt){`,
         `    alloc_locals;`,
-        `    let (start) = wm_alloc(${uint256(structType.width)});`,
+        `    let (__warp_start) = wm_alloc(${uint256(structType.width)});`,
```
rodrodros: I think a better solution here wouldn't be mangling our vars in generated functions but mangling input parameters. In this case like instead of just using the name of the struct, have like member_structname or smth like that

rodrodros (approved):

swapnilraj (approved):

PR #753, opened 2022-10-27
Fix classhash placeholder path

rodrigo-pino:


src/passes/newToDeploy.ts
```diff
@@ -50,8 +50,9 @@ import { getContainingSourceUnit } from '../utils/utils';
  *
  *   Since solidity stores the bytecode of the contract to create, and cairo uses
  *   a declaration address(`class_hash`) of the contract on starknet and the latter
- *   can only be known when interacting with starknet, empty placeholders are used
- *   for compiling, and then filled during deployement with the right address.
+ *   can only be known when interacting with starknet, placeholders are used to
+ *   store this hashes, nonetheless at this stage these hashes cannot be known
```
cicr99, line 54: `this` -> `these`

swapnilraj:
I see the salt bytes size is changed,
this will need adjustment as well then
https://github.com/NethermindEth/UniStark/blob/main/test/shared/utilities.ts#L50

tests/behaviour/expectations/types.ts
```diff
@@ -16,10 +18,10 @@ export class AsyncTest {
     return `${this.name}.sol`;
   }
   get cairo() {
-    return `warp_output/${mangleContractFilePath(manglePath(this.name), this.contract)}.cairo`;
+    return `${OUTPUT_DIR}/${mangleContractFilePath(manglePath(this.name), this.contract)}.cairo`;
```
swapnilraj, line 21: minor nit: path.resolve or path.join instead of /

src/passes/warnSupportedFeatures.ts
```diff
@@ -50,6 +75,15 @@ export class WarnSupportedFeatures extends ASTMapper {
       [...addresses.entries()].forEach(([path, nodes]) => warn(path, nodes));
     }
 
+    if (deploySalts.size > 0) {
+      console.log(
+        `${warning(
+          'Warning',
```
swapnilraj, line 81: @rodrigo-pino can we make a note of this in the docs as well

swapnilraj (approved):

src/passes/warnSupportedFeatures.ts
rodrigo-pino: There was already mention of it in `doc/features/contract_factories.mdx` so I updated that part

JorikSchellekens (approved):

PR #754, opened 2022-10-28
Adding address notes

GuyMcComb:


rodrigo-pino (approved):

cicr99 (approved):

docs/docs/solidity_equivalents/addresses.mdx
```diff
@@ -1,3 +1,9 @@
 ---
-title: Addresses (WIP)
+title: Addresses
 ---
+
+Addresses on Starknet are longer than on Ethereum. On Starknet they take up the whole felt space i.e 251 bits.
```
swapnilraj, line 5: bigger instead of longer

swapnilraj: Can we add information about how the addresses are represented with 256 bits in our flavour of solidity

docs/docs/solidity_equivalents/addresses.mdx
```diff
@@ -1,3 +1,9 @@
...
+This is significantly longer than the Ethereum equivalent of 160 bits.
+
+Please note that the precompile `ecrecover` should be used to recover Ethereum addresses only and not StarkNet addresses.
+This function will only return 160 bits.
```
swapnilraj, line 9: This can be better stated by saying that ecrecover returns a uint160 instead of address now

docs/docs/solidity_equivalents/addresses.mdx
GuyMcComb: Great point!

docs/docs/solidity_equivalents/addresses.mdx
```diff
@@ -1,3 +1,12 @@
...
+title: Addresses
 ---
+
+Addresses on Starknet are larger than on Ethereum.
+On Starknet they take up the whole felt space (251 bits) rather than 160 bits.
+
+To support StarkNet addresses we have slightly modified the solc compiler.
+As a result, addresses in our flavour of Solidity are represented with 256 bits instead of 160 bits.
+
+It is also important to note that the `ecrecover` precompile will not work with StarkNet addresses,
```
JorikSchellekens, line 11:
```suggestion
It is also important to note that the `ecrecover` precompile does not work with StarkNet addresses since it uses Ethereum's curve. The return type of `ercrecover` has been changed from `address` to `uint160` to reflect this. 
```

JorikSchellekens: That said, should we change it StarkNet's curve since we've changed address anyway?

JorikSchellekens (changes requested):

JorikSchellekens (approved):

PR #757, opened 2022-10-31
Decode warp output

LuisLara-UH:
- [x] Sem Tests Running
- [ ] Sem Tests Passed


src/transcode/utils.ts
```diff
@@ -0,0 +1,198 @@
...
+    )
+      throw new CLIError('Solidity abi in file is not a list of function signatures');
+  } else {
+    throw new CLIError('Solidity abi in file is not a list of function signatures.');
+  }
+}
+
+export function addSignature(node: SourceUnit, ast: AST, signature: string, returns = '') {
+  const abi = ast.abi.get(node.id);
+  const signature_with_return: [string, string] = [signature, returns];
```
swapnilraj, line 166: camel case

src/starknetCli.ts
```diff
@@ -210,12 +211,17 @@ export async function runStarknetCallOrInvoke(
...
-      { stdio: 'inherit' },
-    );
+    ).toString('utf-8');
+
+    if (!options.use_cairo_abi) {
+      warpOutput = (
+        await decodeOutputs(filePath, options.function, warpOutput.toString().split(' '))
+      ).toString();
+    }
+    console.log(warpOutput);
```
rodrigo-pino, line 224: Is this logging intentional or a leftover?

src/transcode/utils.ts
```diff
@@ -0,0 +1,198 @@
...
+  } else if (type.components.length !== 0) {
+    // struct
+    return type.components.reduce((acc, ty) => {
+      return acc + getWidthInFeltsOf(ty);
+    }, 0);
+  }
+  throw new Error('Not Supported ' + type.baseType);
+}
+
+export function divmod(x: bigint, y: bigint): [bigint, bigint] {
```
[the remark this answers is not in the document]
LuisLara-UH: Found it. I will update it

src/transcode/utils.ts
LuisLara-UH: Python traumas xD. Will update it

swapnilraj:
@LuisLara-UH for all the new files that are added can you create an export file
similar to these https://github.com/NethermindEth/warp/blob/develop/src/passes/export.ts
https://github.com/NethermindEth/warp/blob/develop/src/export.ts

src/starknetCli.ts
```diff
@@ -210,18 +211,23 @@ export async function runStarknetCallOrInvoke(
...
-
   try {
-    execSync(
+    let warpOutput: string = execSync(
       `${warpVenvPrefix} starknet ${callOrInvoke}  --address ${options.address} --abi ${abiPath} --function ${funcName} --network ${options.network} ${wallet} ${account} ${inputs}`,
-      { stdio: 'inherit' },
-    );
+    ).toString('utf-8');
+
+    if (!options.use_cairo_abi) {
```
swapnilraj, line 219: can you add a check if this was a call, when invoking we don't get a result that we can decode

src/transcode/encode.ts
```diff
@@ -0,0 +1,163 @@
...
+    ? `--inputs ${transcodeCalldata(funcSignature, parseParam(rawInputs.join(' ')))
+        .map((i) => i.toString())
+        .join(' ')}`
+    : '';
+
+  return [funcName, inputs];
+}
+
+export function transcodeCalldata(funcSignature: string, inputs: Param): bigint[] {
+  return parse(funcSignature)(inputs);
```
swapnilraj, line 43: In a future PR can you delete the functionSIgnatureParser and use the encode from below and get the ParamType from the ethers library

src/transcode/decode.ts
```diff
@@ -0,0 +1,187 @@
...
+    type.components.forEach((member, i) => {
+      namedMembers[member.name] = indexedMembers[i];
+    });
+
+    return { ...namedMembers, ...indexedMembers } as Result;
+  }
+  throw Error(`Complex type not supported ${type.type}`);
+}
+
+export function getWidthOf(type: ParamType): number {
```
swapnilraj, line 154: we can delete this as its unsused

src/transcode/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+  const signatureWithReturn: [string, string] = [signature, returns];
+  if (abi === undefined) {
+    ast.abi.set(node.id, new Set([signatureWithReturn]));
+  } else {
+    abi.add(signatureWithReturn);
+  }
+}
+
+export function returnSignature(funcDef: FunctionDefinition): string {
+  const return_params = funcDef.vReturnParameters.vParameters
```
swapnilraj, line 171: oops another camel case

swapnilraj (changes requested):

src/transcode/encode.ts
LuisLara-UH: Ok, I will push this change on another PR

src/transcode/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+  } else if (type.baseType.startsWith('int')) {
+    const width = parseInt(type.baseType.slice(3), 10);
+    return width < 256 ? 1 : 2;
+  } else if (type.baseType.startsWith('address')) {
+    return 1;
+  } else if (type.baseType.startsWith('bool')) {
+    return 1;
+  } else if (/byte\d*$/.test(type.baseType)) {
+    const width = parseInt(type.baseType.slice(4), 10);
+    return width * 8;
```
rodrigo-pino, line 27: Is there an error in the regex here? Shoudn't it be:
```typescript
/bytes\d*$/
```
Also the felt size for fixed bytes is 2 if fixed size is 32 and 1 in other case

src/transcode/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+
+function validateParam(param: unknown) {
+  if (param instanceof Array) {
+    param.map(validateParam);
+    return;
+  }
+  if (param instanceof String) return;
+  if (param instanceof Number) return;
+  if (typeof param === 'string') return;
+  if (typeof param === 'number') return;
```
rodrigo-pino, line 128: Nit: this could be `or`s in a single if

src/transcode/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+    // struct
+    return type.components.reduce((acc, ty) => {
+      return acc + getWidthInFeltsOf(ty);
+    }, 0);
+  }
+  throw new Error('Not Supported ' + type.baseType);
+}
+
+export function isPrimitiveParam(type: ParamType): boolean {
+  // because why use types in a sensible manner?
```
swapnilraj, line 50: this comment is out of sync and can be removed

src/transcode/utils.ts
LuisLara-UH: I just realized method `getWidthInFeltsOf` containing this is not being called. I am not sure what is it for. However, I will fix this and ask Swapnil

swapnilraj:
#764 changes the abi output to a file, so we can remove abiExtractor and parse the abi from output abi file

LuisLara-UH:
> #764 changes the abi output to a file, so we can remove abiExtractor and parse the abi from output abi file

Can we do this on another PR? We could merge this PR if the transcoding functionality is complete, and when #764 is merged we create another PR for removing abiExtractor

swapnilraj:
> > #764 changes the abi output to a file, so we can remove abiExtractor and parse the abi from output abi file
> 
> Can we do this on another PR? We could merge this PR if the transcoding functionality is complete, and when #764 is merged we create another PR for removing abiExtractor

Yea we can do it in a follow-up pr

src/starknetCli.ts
```diff
@@ -210,18 +211,23 @@ export async function runStarknetCallOrInvoke(
...
       `${warpVenvPrefix} starknet ${callOrInvoke}  --address ${options.address} --abi ${abiPath} --function ${funcName} --network ${options.network} ${wallet} ${account} ${inputs}`,
-      { stdio: 'inherit' },
-    );
+    ).toString('utf-8');
+
+    if (isCall && !options.use_cairo_abi) {
+      warpOutput = (
+        await decodeOutputs(filePath, options.function, warpOutput.toString().split(' '))
+      ).toString();
+    }
```
swapnilraj, line 223: since decodeOuputs can return objects calling .toString on retuns [object Object]
you'll need a function that converts a Result type to string

rodrodros (approved):

swapnilraj:
You can test with this call
`bin/warp call warp_output/test__WC__Warp.cairo --address 0x057efbf7928d3d4648b41d163ef5328cff4650bb8ff03f804c8d270793b459e4 --function testStruct --network alpha-goerli`

```
pragma solidity ^0.8.0;

contract Warp {

  function testPrimitive() external pure returns (uint256) {
    return 2^255- 1;
  }

  struct S {
    uint256 a;
    uint56 b;
    bool c;
  }

  function testStruct() external pure returns (S memory) {
    return S(12, 23, true);
  }

  function testArray() external pure returns (uint256[] memory) {
    uint256[] memory g = new uint256[](4);
    g[0] = 0;
    g[1] = 56;
    g[2] = 4353;
    g[3] = 4353465;
    return g;
  }
}
```

LuisLara-UH:
> You can test with this call `bin/warp call warp_output/test__WC__Warp.cairo --address 0x057efbf7928d3d4648b41d163ef5328cff4650bb8ff03f804c8d270793b459e4 --function testStruct --network alpha-goerli`
> 
> ```
> pragma solidity ^0.8.0;
> 
> contract Warp {
> 
>   function testPrimitive() external pure returns (uint256) {
>     return 2^255- 1;
>   }
> 
>   struct S {
>     uint256 a;
>     uint56 b;
>     bool c;
>   }
> 
>   function testStruct() external pure returns (S memory) {
>     return S(12, 23, true);
>   }
> 
>   function testArray() external pure returns (uint256[] memory) {
>     uint256[] memory g = new uint256[](4);
>     g[0] = 0;
>     g[1] = 56;
>     g[2] = 4353;
>     g[3] = 4353465;
>     return g;
>   }
> }
> ```

Tested, outputs:
```12 0 23 1

**Comment by swapnilraj** [comment]
> > You can test with this call `bin/warp call warp_output/test__WC__Warp.cairo --address 0x057efbf7928d3d4648b41d163ef5328cff4650bb8ff03f804c8d270793b459e4 --function testStruct --network alpha-goerli`
> > ```
> > pragma solidity ^0.8.0;
> > 
> > contract Warp {
> > 
> >   function testPrimitive() external pure returns (uint256) {
> >     return 2^255- 1;
> >   }
> > 
> >   struct S {
> >     uint256 a;
> >     uint56 b;
> >     bool c;
> >   }
> > 
> >   function testStruct() external pure returns (S memory) {
> >     return S(12, 23, true);
> >   }
> > 
> >   function testArray() external pure returns (uint256[] memory) {
> >     uint256[] memory g = new uint256[](4);
> >     g[0] = 0;
> >     g[1] = 56;
> >     g[2] = 4353;
> >     g[3] = 4353465;
> >     return g;
> >   }
> > }
> > ```
> 
> Tested, are these outputs correct?
> 
> ```
> 12 0 23 1
> ```

The output should be 12 23 true,
The first parameter in the struct is a Uint256, so its split into low and high but we should not expose that to the user.
we should format is better as well but that can be done later when we use the solidity output from nethersolc because then we will have the member names as well.

**Comment by LuisLara-UH** [comment]
> we should format is better as well
Sorry I don't get this

**Comment by LuisLara-UH** [comment]
I think in another PR we should add some more tests to encoding

**Review by swapnilraj — APPROVED** [review]

File: tests/cli.test.ts [hunk 12]
```diff
@@ -33,7 +34,7 @@ const tests: encodeTest[] = [
   ],
   ['mint(bool,bool,bool,bool)', ['true', 'false', '1', '0'], ['0x1', '0x0', '0x1', '0x0']],
   ['mint(bytes8)', ['0xffffffffffffffff'], ['0xffffffffffffffff']],
-  ['byte_me(bytes)', [['0xff', '0xff', '0xff', '0xff']], ['0x4', '0xff', '0xff', '0xff', '0xff']],
+  ['byte_me(bytes[])', [['0xff', '0xff', '0xff', '0xff']], ['0x4', '0xff', '0xff', '0xff', '0xff']],
```
swapnilraj, line 37: why did this behaviour change?

LuisLara-UH: The behaviour of `bytes` is similar to `string`: its an array of `byte`, but when you are going to parse a value hexadecimal you write it as a complete hexadecimal string, not as an array of hexadecimal values that can be parsed into one byte. 
An example of that is the previous function that receives `bytes8`. In this case, it is also an array of `byte`, but of fixed-size(8). The value you pass to it is `'0xffffffffffffffff'`, not `['0xff', '0xff', '0xff', '0xff']`.
Even if we would want to encode it that way, it may be confusing at the time of parsing. It would be similar to pass `string` values as explicit arrays of characters, like `['s', 't', 'r', 'i', 'n', 'g']`. In this case it would parse it as an array of `string`, not as an array of `char`.

GuyMcComb:
Sem failures

-  tests/behaviour/solidity/test/libsolidity/semanticTests/libraries/using_library_mappings_return

LuisLara-UH:
The semantic test is failing on `abiExtractor`. Given that we remove it in `remove-abi-extractor` PR, there this semantic test is passing. The `remove-abi-extractor` PR has all commits from here, so I think we could just take that one for merging into develop, and close this

PR #763, opened 2022-11-05
Replace if-functionalisation with tempvar postpends

JorikSchellekens:
This replaces if functionalisation by appending tempvars for all live variables
at the end of each if branch and if statement. This is explained in more depth
in the `ifStatementTempVarPrepender` pass.

I believe that StarkWare should just fix their ifs in Cairo and if this results
in a significant performance boost will be lobbying them to just get it fixed
so we can drop this pass altogether.


src/utils/nodeTemplates.ts
```diff
@@ -269,7 +276,7 @@ export function createDefaultConstructor(node: ContractDefinition, ast: AST): Fu
 
 export function createVariableDeclarationStatement(
   varDecls: (VariableDeclaration | null)[],
-  intitalValue: Expression,
+  intitalValue: Expression | undefined,
```
swapnilraj, line 279: I don't think this is needed

JorikSchellekens: That is the type of `initialValue` for a `VariableDeclarationStatement`

JorikSchellekens:
Got an ack from @GuyMcComb that this branch passed the semantic tests

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+      if (!this.liveVars.delete(implicit)) {
+        throw new TranspileFailedError(
+          `Implicit variable dead. ${implicit} in function ${node.name}.`,
+        );
+      }
+    });
+    node.vParameters.vParameters.forEach((p) => {
+      if (!this.liveVars.delete(p.name) && node.functionStubKind === FunctionStubKind.None) {
+        node.functionStubKind;
+        // console.log(`Dead parameter found ${p.name} in function ${node.name}`);
```
GuyMcComb, line 91: console.log

GuyMcComb: Other than that, it looks good to go!

src/ast/visitor.ts
```diff
@@ -68,6 +68,7 @@ import {
 import { CairoAssert, CairoContract, CairoFunctionDefinition } from './cairoNodes';
 
 import { AST } from './ast';
+import { CairoTempVar } from './cairoNodes/tempvar';
```
cicr99: I think you can just add `CairoTempVar` to the imports in line 68

src/ast/cairoNodes/tempvar.ts
```diff
@@ -0,0 +1,13 @@
+import { Statement, StructuredDocumentation } from 'solc-typed-ast';
+
+export class CairoTempVar extends Statement {
```
cicr99: I think an export for `CairoTempVars` in the export file should be added as well

src/cairoWriter.ts
```diff
@@ -80,6 +80,7 @@ import {
   CairoFunctionDefinition,
   FunctionStubKind,
 } from './ast/cairoNodes';
+import { CairoTempVar } from './ast/cairoNodes/tempvar';
```
cicr99: This import can be added to the imports from `'./ast/cairoNodes'`

src/passes/identityFunctionRemover.ts
```diff
@@ -33,7 +33,7 @@ export class IdentityFunctionRemover extends ASTMapper {
   // Function to add passes that should have been run before this pass
   addInitialPassPrerequisites(): void {
     const passKeys: Set<string> = new Set<string>([
-      'If', // Most of the functions this pass removes are generated by the IfFunctionaliser
+      // 'If', // Most of the functions this pass removes are generated by the IfFunctionaliser
```
cicr99: This can be removed now

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+ ********************
+ */
+
+export class IfStatementTempVarPostpender extends ASTMapper {
+  private liveVars: Set<string> = new Set();
+
+  // Function to add passes that should have been run before this pass
+  addInitialPassPrerequisites(): void {
+    const passKeys: Set<string> = new Set<string>([
+      'An', // Most of the functions this pass removes are generated by the IfFunctionaliser
```
cicr99: The comment in here is not correct for the `AnnotateImplicits` pass

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+    ]);
+    passKeys.forEach((key) => this.addPassPrerequisite(key));
+  }
+
+  visitCairoFunctionDefinition(node: CairoFunctionDefinition, ast: AST): void {
+    this.liveVars = new Set<string>();
+    [...node.implicits].forEach((implicit) => this.liveVars.add(implicit));
+    // node.vReturnParameters.vParameters.forEach((p) => this.liveVars.add(p.name));
```
cicr99: Is this supposed to stay as a comment? If so, it should be removed.

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+    [...node.implicits].forEach((implicit) => {
+      if (!this.liveVars.delete(implicit)) {
+        throw new TranspileFailedError(
+          `Implicit variable dead. ${implicit} in function ${node.name}.`,
+        );
+      }
+    });
+    node.vParameters.vParameters.forEach((p) => {
+      if (!this.liveVars.delete(p.name) && node.functionStubKind === FunctionStubKind.None) {
+        node.functionStubKind;
```
cicr99: What's the purpose of this line?

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+        // console.log(`Dead parameter found ${p.name} in function ${node.name}`);
+      }
+    });
+    this.liveVars = new Set<string>();
+  }
+
+  visitIdentifier(node: Identifier, _ast: AST): void {
+    if (
+      node.vReferencedDeclaration &&
+      node.vReferencedDeclaration.getClosestParentByType(CairoFunctionDefinition)
```
cicr99: I assume this declaration should be in the same function definition as the identifier, but this condition does not ensure that. I think I would like some kind of assertion here to ensure it.

src/utils/astChecking.ts
```diff
@@ -669,6 +670,8 @@ export function checkSane(unit: SourceUnit, ctx: ASTContext): void {
       checkDirectChildren(node, 'vSubExpression');
     } else if (node instanceof CairoAssert) {
       checkDirectChildren(node, 'vExpression');
+    } else if (node instanceof CairoTempVar) {
+      // TODO make sure vars are live after and declared before
```
cicr99: What about this TODO?

src/passes/identityFunctionRemover.ts
rodrodros: This line can go

src/passes/ifStatementTempVarPostpender.ts
rodrodros: I think this comment is a left over since `IfFunctionaliser` no longer exists. I would put instead as a requirement for having cairoFunctionDefintions which are created by the `AnotateImplicits`

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+    const trueHasPathWithoutReturn = hasPathWithoutReturn(node.vTrueBody);
+    if (trueHasPathWithoutReturn) {
+      for (const liveVar of liveVarsBefore) {
+        const child = createCairoTempVar(liveVar, ast);
+        (node.vTrueBody as Block).appendChild(child);
+        ast.registerChild(child, node.vTrueBody);
+      }
+    }
+    const falseHasPathWithoutReturn =
+      (node.vFalseBody && hasPathWithoutReturn(node.vFalseBody)) || !node.vFalseBody;
```
rodrodros: nit: refactor to:
`node.vFalseBody === undefined || hasPathWithoutReturn(node.falseBody)`

src/ast/cairoNodes/tempvar.ts
rodrodros: I think appending `Statement` to the class name would be good for readability purposes. When reading the code later it's clear we are creating a Statement. Also this way, we follow the same pattern as solc-typed-ast nodes

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+function ensureBothBranchesAreBlocks(node: IfStatement, ast: AST): void {
+  if (!(node.vTrueBody instanceof Block) && !(node.vTrueBody instanceof UncheckedBlock)) {
+    node.vTrueBody = createBlock([node.vTrueBody], ast);
+    ast.registerChild(node.vTrueBody, node);
+  }
+
+  if (
+    node.vFalseBody &&
```
rodrodros: opinionated nitpick: don't leave to JavaScript engine to coerce undefined to boolean or vice versa, instead declare it explicitly like `node.vFalseBody !== undefined`

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,200 @@
...
+      this.dispatchVisit(node.vFalseBody, ast);
+      liveVarsAfterFalseBody = new Set(this.liveVars);
+    }
+    this.liveVars = new Set(liveVarsBefore);
+    this.dispatchVisit(node.vCondition, ast);
+    this.liveVars = new Set([
+      ...this.liveVars,
+      ...liveVarsAfterFalseBody,
+      ...liveVarsAfterTrueBody,
+    ]);
```
rodrodros, line 153: I think this last assignation to `liveVars` is unnecesary, no further node visits are done after this point

JorikSchellekens: This is a class-level record of the `liveVars` at a point in the code. The visited ifStatement might be in a block with statements which proceed it. These statements will not have been visited yet. The live vars of an if is the union of the live vars of the condition and the two branches. This line is correct.

src/utils/astChecking.ts
JorikSchellekens: This was left as a todo for a sanity check. It should check that the defined tempvar is indeed live. However, this node will never actually be hit because it should only introduced by this pass. I would be happy to have it removed.

src/passes/ifStatementTempVarPostpender.ts
JorikSchellekens: Lost line from an old debug

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  if (
+    node.vFalseBody !== undefined &&
+    !(node.vFalseBody instanceof Block) &&
+    !(node.vFalseBody instanceof UncheckedBlock)
+  ) {
+    node.vFalseBody = createBlock([node.vFalseBody], ast);
+    ast.registerChild(node.vFalseBody, node);
+  } else if (!node.vFalseBody) {
```
piwonskp: Could you use a more strict condition? Also it seems the if-else block could be restructured. E.g. wouldn't putting the else block before eliminate the need of `node.vFalseBody !== undefined`?

rodrigo-pino: Well the second if is actually a bug!

The else block was removing any kind of sentence in the `FalseBody`. I wonder why no test failed

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,193 @@
...
+  IfStatement,
+  UncheckedBlock,
+  VariableDeclarationStatement,
+} from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+import { TranspileFailedError } from '../utils/errors';
+import { CairoFunctionDefinition } from '../ast/cairoNodes';
+import { createBlock, createCairoTempVar } from '../utils/nodeTemplates';
+import { hasPathWithoutReturn } from '../export';
```
cicr99: This import should be from the source file not the export file

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,193 @@
...
+ ********************
+ */
+
+export class IfStatementTempVarPostpender extends ASTMapper {
+  private liveVars: Set<string> = new Set();
+
+  // Function to add passes that should have been run before this pass
+  addInitialPassPrerequisites(): void {
+    const passKeys: Set<string> = new Set<string>([
+      'An', // An creates cairoFunctionDefinition which this pass use
```
cicr99: I would use the name of the pass in the comment instead of `An` again, for more readability

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,193 @@
...
+          `Implicit variable dead. ${implicit} in function ${node.name}.`,
+        );
+      }
+    });
+    this.liveVars = new Set<string>();
+  }
+
+  visitIdentifier(node: Identifier, _ast: AST): void {
+    if (
+      node.vReferencedDeclaration &&
```
cicr99: I think it's better to be specific, as in `node.vReferencedDeclaration !== undefined`

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+    }
+    const falseHasPathWithoutReturn =
+      node.vFalseBody === undefined || hasPathWithoutReturn(node.vFalseBody);
+
+    if (falseHasPathWithoutReturn) {
+      for (const liveVar of liveVarsBefore) {
+        const child = createCairoTempVar(liveVar, ast);
+        (node.vFalseBody as Block).appendChild(child);
+        // @ts-ignore ensureBothBranchesAreBlocks forces vFalseBody to be defined;
```
cicr99: Wouldn't it be better to use an assertion right after `ensureBothBranchesAreBlocks` is called, making sure `node.vFalseBody !== undefined`?. This way there is no need to be asking about it all the time, and this comment won't be needed either.

src/utils/astChecking.ts
```diff
@@ -669,6 +669,9 @@ export function checkSane(unit: SourceUnit, ctx: ASTContext): void {
       checkDirectChildren(node, 'vSubExpression');
     } else if (node instanceof CairoAssert) {
       checkDirectChildren(node, 'vExpression');
+    } else if (node instanceof CairoTempVarStatement) {
+      // Not being checked because this node does not gets affected by any
```
cicr99: gets -> get

rodrodros: grammar police :policeman:

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,194 @@
...
+    ensureBothBranchesAreBlocks(node, ast);
+    assert(node.vFalseBody instanceof Block);
+
+    const liveVarsBefore = new Set(this.liveVars);
+    this.dispatchVisit(node.vTrueBody, ast);
+    const liveVarsAfterTrueBody = new Set(this.liveVars);
+
+    this.liveVars = new Set(liveVarsBefore);
+    let liveVarsAfterFalseBody = new Set<string>();
+    if (node.vFalseBody) {
```
cicr99: With the previous assertion this condition should always be true, so I don't think it is necessary

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,194 @@
...
+    const trueHasPathWithoutReturn = hasPathWithoutReturn(node.vTrueBody);
+    if (trueHasPathWithoutReturn) {
+      for (const liveVar of liveVarsBefore) {
+        const child = createCairoTempVar(liveVar, ast);
+        (node.vTrueBody as Block).appendChild(child);
+        ast.registerChild(child, node.vTrueBody);
+      }
+    }
+    const falseHasPathWithoutReturn =
+      node.vFalseBody === undefined || hasPathWithoutReturn(node.vFalseBody);
```
cicr99: Same here, `node.vFalseBody === undefined` should always be false, so you should be able to leave it just like `const falseHasPathWithoutReturn = hasPathWithoutReturn(node.vFalseBody);`

cicr99 (approved):

swapnilraj (approved):

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+      'An', // AnotateImplicit pass creates cairoFunctionDefinitions which this pass use
+    ]);
+    passKeys.forEach((key) => this.addPassPrerequisite(key));
+  }
+
+  visitCairoFunctionDefinition(node: CairoFunctionDefinition, ast: AST): void {
+    this.liveVars = new Set<string>();
+    [...node.implicits].forEach((implicit) => this.liveVars.add(implicit));
+    // analyse the block
+    if (node.vBody) {
```
piwonskp: A more strict condition would be useful

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+      this.dispatchVisit(node.vBody, ast);
+    }
```
piwonskp, line 80: Imo a `filter` and `map` would be a more elegant choice here unless the above and below foreach loops are merged

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+    // remove the parameters and implicits
+    [...node.implicits].forEach((implicit) => {
+      if (!this.liveVars.delete(implicit)) {
+        throw new TranspileFailedError(
+          `Implicit variable dead. ${implicit} in function ${node.name}.`,
+        );
+      }
+    });
```
piwonskp: Wouldn't it be better to try to delete all, have a list of dead variables, and then print them all at once if there are any?

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+      ...this.liveVars,
+      ...liveVarsAfterFalseBody,
+      ...liveVarsAfterTrueBody,
+    ]);
+
+    const trueHasPathWithoutReturn = hasPathWithoutReturn(node.vTrueBody);
+    if (trueHasPathWithoutReturn) {
+      for (const liveVar of liveVarsBefore) {
+        const child = createCairoTempVar(liveVar, ast);
+        (node.vTrueBody as Block).appendChild(child);
```
piwonskp: Assert that `vTrueBody` is a block similarly to `vFalseBody`

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+    const falseHasPathWithoutReturn = hasPathWithoutReturn(node.vFalseBody);
+
+    if (falseHasPathWithoutReturn) {
+      for (const liveVar of liveVarsBefore) {
+        const child = createCairoTempVar(liveVar, ast);
+        node.vFalseBody.appendChild(child);
+        ast.registerChild(child, node.vFalseBody);
+      }
```
piwonskp: The for loop could be extracted into a separate method `setBlockTempVars(ast, liveVars, block)` and used both by true and false bodies

src/passes/ifStatementTempVarPostpender.ts
```diff
@@ -0,0 +1,192 @@
...
+    ast.registerChild(node.vTrueBody, node);
+  }
+
+  if (node.vFalseBody === undefined) {
+    node.vFalseBody = createBlock([], ast);
+    ast.registerChild(node.vFalseBody, node);
+  } else if (!(node.vFalseBody instanceof Block) && !(node.vFalseBody instanceof UncheckedBlock)) {
+    node.vFalseBody = createBlock([node.vFalseBody], ast);
+    ast.registerChild(node.vFalseBody, node);
+  }
```
piwonskp, line 188: Could be refactored to:
```
node.vFalseBody = ensureWrappedWithBlock(ast, node.node.vFalseBody)
ast.registerChild(node.vFalseBody, node);
```

rodrodros: Would `ensureWrappedWithBlock` be a new function definition?

src/passes/ifStatementTempVarPostpender.ts
rodrodros: I am not sure I follow

src/passes/ifStatementTempVarPostpender.ts
rodrodros: If yes, then I think it would be overkill, code repetition is so little

src/passes/ifStatementTempVarPostpender.ts
piwonskp: Oh my bad. I've read if clause as an element of foreach loop

src/passes/ifStatementTempVarPostpender.ts
piwonskp: Yea, new function, I was not entirely sure which solution I prefer as well. That's just a suggestion. It might be a useful utility function as there might be similar cases with other statements. And its slightly more readable

piwonskp (approved):

PR #764, opened 2022-11-05
Output solidity ABI file

swapnilraj:
Don't mangle output file paths!!
storage code moved below

In a future PR we can remove abiExtractor as well


JorikSchellekens (approved):
LGTM!

.gitignore
```diff
@@ -134,9 +134,9 @@ solc-v0.7.6
 
 # output from warp
 warp_output/
-example__contracts
-example_contracts/**/*.cairo
-example_contracts/**/*.json
+warp_test/
+warp_test/**/*.cairo
```
JorikSchellekens, line 138: The two globed paths are not required

src/testing.ts
```diff
@@ -257,10 +256,12 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
   if (force) {
     postTestCleanup();
   } else if (!preTestChecks()) return;
+  console.log('findSolSourceFilePaths');
```
JorikSchellekens, line 259: Console log

src/testing.ts
```diff
@@ -276,12 +277,12 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
 }
 
 function preTestChecks(): boolean {
-  if (!checkNoCairo('example__contracts')) {
-    console.log('Please remove example__contracts, or run with -f to delete it');
+  if (!checkNoCairo(path.join('warp_test', 'example_contracts'))) {
+    console.log('Please remove example_contracts, or run with -f to delete it');
```
JorikSchellekens, line 281: Should probably read `warp_test/example_contracts` instead

JorikSchellekens (changes requested):

src/testing.ts
swapnilraj: oopsie

src/testing.ts
swapnilraj: why?

src/testing.ts
swapnilraj: fixed

.gitignore
swapnilraj: fixed

src/testing.ts
JorikSchellekens: Because that's the folder that should be removed

swapnilraj: right I see, I was confused between the code and string literal

JorikSchellekens (approved):

src/passes/filePathMangler.ts
```diff
@@ -1,15 +1,11 @@
-import { ImportDirective, SourceUnit } from 'solc-typed-ast';
```
rodrigo-pino, line 1: I think we can safely delete this pass now, and move the checkPath to rejectUnsupportedFeatures

PR #765, opened 2022-11-08
Remove abi extractor

LuisLara-UH:
- [X] Sem Tests Passed


docs/docs/features/cairo_stubs.mdx
```diff
@@ -102,7 +102,7 @@ namespace Example:
     end
 end
 
-# Original soldity abi: ["constructor()","recursiveAdd(uint8,uint8)"]
+# Original solidity abi: [["constructor()", ""],["recursiveAdd(uint8,uint8)", "uint8"]]
```
swapnilraj: We can delete this

docs/docs/getting_started/a-usage-and-installation.mdx
```diff
@@ -234,7 +234,7 @@ namespace foo:
     end
 end
 
-# Original soldity abi: ["constructor(uint256)",""]
+# Original solidity abi: [["constructor(uint256)",""]]
```
swapnilraj: this as well

src/ast/ast.ts
```diff
@@ -51,9 +51,8 @@ import { CairoFunctionDefinition } from './cairoNodes';
 export class AST {
   // SourceUnit id -> CairoUtilFuncGen
   private cairoUtilFuncGen: Map<number, CairoUtilFuncGen> = new Map();
-  // TODO we can delete this stuff below
-  // SourceUnit id -> function signatures
-  abi: Map<number, Set<string>> = new Map();
+  // SourceUnit id -> function [signatures, returns]
+  abi: Map<number, Set<[string, string]>> = new Map();
```
swapnilraj: if the abi extractor is deleted there should be no writes into this; so we can delete this

src/io.ts
```diff
@@ -5,7 +5,7 @@ import { TranspileFailedError, logError } from './utils/errors';
 import { execSync } from 'child_process';
 import { AST } from './ast/ast';
 
-export const solABIPrefix = '// Original soldity abi:';
+export const solABIPrefix = '// Original solidity abi:';
```
swapnilraj: delete

src/starknetCli.ts
```diff
@@ -198,7 +213,7 @@ export async function runStarknetCallOrInvoke(
   let funcName, inputs: string;
   try {
     [funcName, inputs] = await encodeInputs(
-      filePath,
+      solAbiPath ?? '',
```
swapnilraj: we should try to regenerat the solidity abi if its not found;
we have the path to solidity file

src/utils/utils.ts
```diff
@@ -569,3 +569,7 @@ export function getContainingSourceUnit(node: ASTNode): SourceUnit {
   assert(root !== undefined, `Unable to find root source unit for ${printNode(node)}`);
   return root;
 }
+
+export function dumpABI(node: SourceUnit, ast: AST): string {
```
swapnilraj: we don't need this since the abi mapping will empty now

tests/cli.test.ts
```diff
@@ -48,7 +49,7 @@ const tests: encodeTest[] = [
 describe('Solidity abi parsing and decode tests', function () {
   tests.map(([signature, input, output]) =>
     it(`parses ${signature}`, () => {
-      const result = parse(signature)(input);
+      const result = encode(FunctionFragment.fromString(signature).inputs, input);
```
swapnilraj, line 52: nice!

src/starknetCli.ts
LuisLara-UH: The solidity abi file is always generated unless it fails to compile, in which case it returns immediately and it gets not here. The same happens with abiPath, which is being used in the command later. That undefined checking is pure formality with JS. However, if you want I can put the code for generating the path there

src/starknetCli.ts
```diff
@@ -210,18 +225,26 @@ export async function runStarknetCallOrInvoke(
...
       `${warpVenvPrefix} starknet ${callOrInvoke}  --address ${options.address} --abi ${abiPath} --function ${funcName} --network ${options.network} ${wallet} ${account} ${inputs}`,
-      { stdio: 'inherit' },
-    );
+    ).toString('utf-8');
+
+    if (isCall && !options.use_cairo_abi) {
+      const decodedOutputs = await decodeOutputs(
+        solAbiPath ?? '',
+        options.function,
+        warpOutput.toString().split(' '),
```
swapnilraj, line 237: Now that we have the components of a struct in the abi we can improve the formatting and put the key names for a struct and for arrays we should wrap them in []

tests/cli.test.ts
```diff
@@ -33,7 +34,7 @@ const tests: encodeTest[] = [
   ],
   ['mint(bool,bool,bool,bool)', ['true', 'false', '1', '0'], ['0x1', '0x0', '0x1', '0x0']],
   ['mint(bytes8)', ['0xffffffffffffffff'], ['0xffffffffffffffff']],
-  ['byte_me(bytes)', [['0xff', '0xff', '0xff', '0xff']], ['0x4', '0xff', '0xff', '0xff', '0xff']],
+  ['byte_me(bytes[])', [['0xff', '0xff', '0xff', '0xff']], ['0x4', '0xff', '0xff', '0xff', '0xff']],
```
swapnilraj, line 37: are the cli tests run in CI?

swapnilraj (changes requested):

tests/cli.test.ts
LuisLara-UH: Yup, I realized of the encoding bugs because they were failing in CI

GuyMcComb:
Sem Tests Failing:
tests/behaviour/solidity/test/libsolidity/semanticTests/array/copying/array_copy_different_packing

LuisLara-UH:
`tests/behaviour/solidity/test/libsolidity/semanticTests/array/copying/array_copy_different_packing`

This semantic test is failing in develop too: the problem is that there is a contract named `c`, which is also the name of a variable inside. It is easily patched just changing the name of the contract. I think it is not an issue from this branch, and we could fix it on another PR.

LuisLara-UH:
I added the new formatting for the decoded outputs, where I added the special formatting for structs and arrays. I tested it with a few cases, but I think we can implement some testing for decoding, like the one made for encoding (https://github.com/NethermindEth/warp/blob/develop/tests/cli.test.ts#L9)

swapnilraj:
> `tests/behaviour/solidity/test/libsolidity/semanticTests/array/copying/array_copy_different_packing`
> 
> This semantic test is failing in develop too: the problem is that there is a contract named `c`, which is also the name of a variable inside. It is easily patched just changing the name of the contract. I think it is not an issue from this branch, and we could fix it on another PR.

That's the wrong fix, we can't change the contract to make it work for our tests. Can you share the test result?
@GuyMcComb

GuyMcComb (approved):

swapnilraj (approved):

PR #767, opened 2022-11-09
Add testing for underscore integers

LuisLara-UH:


swapnilraj (approved):

rodrodros:
Found an awkward case where it breaks:
```solidity
if (3_000 < 5_000) { ... }
```
It is easily patched, but it's odd to me why different behaviour if the literal is in different places. If the source of it can be found, I think it would good

LuisLara-UH:
Patched it and added the testing. I am going to create a ticket for investigating this later.

rodrodros (approved):

tests/behaviour/contracts/underscore_integers/underscore_integers.sol
```diff
@@ -0,0 +1,18 @@
...
+    function a() public pure returns (uint256) {
+        if(5_000 > 3_000)
+            return x;
+        return 0;
+    }
+
+    function b() public pure returns (int) {
+        return y;
+    }
+}
```
GuyMcComb: Empty Line

PR #768, opened 2022-11-09
Inline assembly blocks

piwonskp:


src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+  Block,
+  Expression,
+  ExpressionStatement,
+  VariableDeclaration,
+  UncheckedBlock,
+} from 'solc-typed-ast';
+import assert from 'assert';
+
+class YulTransformer {
+  vars: { [name: string]: VariableDeclaration };
```
rodrodros: I would suggest using a `Mapping` here to be on par with project style

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+    return createExpressionStatement(this.ast, assignment as Expression);
+  }
+}
+
+export class InlineAssemblyTransformer extends ASTMapper {
+  visitInlineAssembly(node: InlineAssembly, ast: AST): void {
+    let transformer = new YulTransformer(ast, node.externalReferences, node.context!.map);
+    const statements = node.yul!.statements.map((yul: YulNode) => {
+      let astNode = transformer.run(yul);
```
rodrodros: Avoid using `let` when possible

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    const blockRefs: [number, ASTNode][] = [...contextMap].filter(([id, _]) =>
+      blockRefIds.includes(id),
+    );
+    this.vars = Object.fromEntries(
+      blockRefs.map(([id, ref]) => [(ref as Identifier).name, ref as VariableDeclaration]),
+    );
+  }
+
+  run(node: YulNode): ASTNode {
+    if (!Object.hasOwn(this, node.nodeType)) throw `${node.nodeType} is not supported`;
```
rodrodros: I am not sure `Object.hasOwn` is 100% right in this scenario, also it currently does not compile with `yarn tsc`.

Also please throw warp defined exception, like `NotSupportedYet`. Exceptions are defined in `src/utils/errors.ts`

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -93,9 +92,6 @@ export class RejectUnsupportedFeatures extends ASTMapper {
     }
     this.visitExpression(node, ast);
   }
-  visitInlineAssembly(node: InlineAssembly, _ast: AST): void {
-    this.addUnsupported('Yul blocks are not supported', node);
-  }
```
rodrodros, line 97: I think we should still notify about yul blocks which haven't only the supported operations

src/transpiler.ts
```diff
@@ -100,6 +101,7 @@ export function transform(ast: AST, options: TranspilationOptions & PrintOptions
 
 function applyPasses(ast: AST, options: TranspilationOptions & PrintOptions): AST {
   const passes: Map<string, typeof ASTMapper> = createPassMap([
+    ['Iat', InlineAssemblyTransformer],
```
rodrodros: I think this pass should go after RejectUnsupportedFeatures

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+  ast: AST;
+
+  constructor(ast: AST, externalReferences: any[], contextMap: Map<number, ASTNode>) {
+    this.ast = ast;
+    const blockRefIds = externalReferences.map((ref) => ref.declaration);
+    const blockRefs: [number, ASTNode][] = [...contextMap].filter(([id, _]) =>
+      blockRefIds.includes(id),
+    );
+    this.vars = Object.fromEntries(
+      blockRefs.map(([id, ref]) => [(ref as Identifier).name, ref as VariableDeclaration]),
```
rodrodros: I think treating  `ref` first as `Identifer` and then later as `VariableDeclaration` hurts readability. I think `blockeRef` type since it's creation should be restricted to it's most strict type, `[number, VariableDeclaration][]` in this case.

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+        leftExpr,
+        rightExpr,
+        node,
+      );
+    }
+    throw 'Only binary operations are supported';
+  }
+
+  YulAssignment(node: YulNode): ExpressionStatement {
+    let vars: Expression;
```
rodrodros: nit: more useful name to `vars`. I think `yulVar`  or `yulExpr` (or smth like that) would match better in this context

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    const blockRefIds = externalReferences.map((ref) => ref.declaration);
+    const blockRefs: [number, ASTNode][] = [...contextMap].filter(([id, _]) =>
+      blockRefIds.includes(id),
+    );
+    this.vars = Object.fromEntries(
+      blockRefs.map(([id, ref]) => [(ref as Identifier).name, ref as VariableDeclaration]),
+    );
+  }
+
+  run(node: YulNode): ASTNode {
```
rodrodros: If `run` result is always going to be tread as an `Expression`, I would suggest reflecting it on the function signature

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    return method.bind(this)(node);
+  }
+
+  YulIdentifier(node: YulNode): Identifier {
+    return createIdentifier(this.vars[node.name], this.ast);
+  }
+
+  YulFunctionCall(node: YulNode): BinaryOperation {
+    const binary_ops: { [name: string]: string } = { add: '+', mul: '*', div: '/', sub: '-' };
+    if (Object.keys(binary_ops).includes(node.functionName.name)) {
```
rodrodros: bynary_op -> bynaryOp
nit: I would suggest using a mapping in here as well. I think overall it would improve readability

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    const assignment = new Assignment(
+      this.ast.reserveId(),
+      node.src,
+      vars.typeString,
+      '=',
+      vars,
+      value,
+      node,
+    );
+    return createExpressionStatement(this.ast, assignment as Expression);
```
rodrodros: No need here to treat assignmet as an expression

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+  }
+
+  run(node: YulNode): ASTNode {
+    if (!Object.hasOwn(this, node.nodeType)) throw `${node.nodeType} is not supported`;
+    const method = this[node.nodeType as keyof YulTransformer] as (node: YulNode) => ASTNode;
+    return method.bind(this)(node);
+  }
+
+  YulIdentifier(node: YulNode): Identifier {
```
rodrodros: I would suggest renaming this funcs: `YulIdentifier`, `YulAssignment etc... to: `createYulIdentifier`, `createYulAssignment`

src/passes/inlineAssemblyTransformer.ts
piwonskp: Good catch, I forgot to commit tsconfig.json

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,125 @@
...
+  Assignment,
+  Block,
+  Expression,
+  ExpressionStatement,
+  VariableDeclaration,
+  UncheckedBlock,
+  Literal,
+} from 'solc-typed-ast';
+import assert from 'assert';
+import { WillNotSupportError } from '../export';
```
cicr99: Import this error from the file where it is defined instead of `export` file
Also, this pass should also be added to the `export` file

src/passes/inlineAssemblyTransformer.ts
piwonskp: I considered naming it vars and values or lhs and rhs

src/passes/inlineAssemblyTransformer.ts
piwonskp: Imo object syntax is more readable than a list of lists. It's easier to read an object than a bunch of square brackets and commas. And we don't care about the order of items here either unlike e.g passes Map

src/passes/rejectUnsupportedFeatures.ts
piwonskp: It's done in the `YulTransformer` class

src/passes/inlineAssemblyTransformer.ts
piwonskp: My initial idea was to be able to easily unit test the class thus the ASTNode type so you could pass any YulNode and receive any ASTNode. However many of util functions require `ast` and `node` so it will be hard to isolate the class anyway. I guess I will just use `Statement` as a return type

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,124 @@
...
+      node,
+    );
+    return createExpressionStatement(this.ast, assignment);
+  }
+}
+
+export class InlineAssemblyTransformer extends ASTMapper {
+  visitInlineAssembly(node: InlineAssembly, ast: AST): void {
+    const transformer = new YulTransformer(ast, node);
+    const statements = node.yul!.statements.map((yul: YulNode) => {
```
cicr99: An assertion to make sure `node.yul` is different from undefined could be added before this line with an error message. Then there would be no need to use `!` after `node.yul`

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,124 @@
...
+    return createNumberLiteral(node.value, this.ast);
+  }
+
+  createYulIdentifier(node: YulNode): Identifier {
+    return createIdentifier(this.vars.get(node.name)!, this.ast);
+  }
+
+  createYulFunctionCall(node: YulNode): BinaryOperation {
+    const binaryOps: { [name: string]: string } = { add: '+', mul: '*', div: '/', sub: '-' };
+    if (Object.keys(binaryOps).includes(node.functionName.name)) {
```
swapnilraj: its enough to check if the result of `binaryOps[node.functionName.name]` is not undefined

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,124 @@
...
+    const method = this[methodName as keyof YulTransformer] as (node: YulNode) => ASTNode;
+    return method.bind(this)(node);
+  }
+
+  createYulLiteral(node: YulNode): Literal {
+    return createNumberLiteral(node.value, this.ast);
+  }
+
+  createYulIdentifier(node: YulNode): Identifier {
+    return createIdentifier(this.vars.get(node.name)!, this.ast);
```
swapnilraj: we should add a check if the lookup of node.name is not undefined

src/passes/rejectUnsupportedFeatures.ts
rodrodros: Can you make it the same way as `RejectUnsupportedPass` which collects all the errors first and at the end it throws them all. Since this pass is going to be run after `Reject`, you could define error collection there, reusing function in this class

src/passes/inlineAssemblyTransformer.ts
rodrodros: It doesn't feel right to me to use the result of `run` as an Expression or Statement depending on the context. They have different semantic meanings, the result of `run` is either an expression or a statement. Treating this way as it can be what you want depending on where the function was call, in my opinion makes the code unreadable for others and for the future you which won't remember

src/passes/inlineAssemblyTransformer.ts
rodrodros: I meant overall, in the way that the if condition later would have become `bynaryOps.has(<element>)`

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,124 @@
...
+      blockRefIds.includes(id),
+    ) as [number, VariableDeclaration][];
+    this.vars = new Map(blockRefs.map(([id, ref]) => [ref.name, ref]));
+  }
+
+  run(node: YulNode): Statement {
+    const methodName = `create${node.nodeType}`;
+    if (!(methodName in this))
+      throw new WillNotSupportError(`${node.nodeType} is not supported`, this.assemblyRoot, false);
+    const method = this[methodName as keyof YulTransformer] as (node: YulNode) => ASTNode;
```
piwonskp, line 51: It seems to be inconsistent with [https://github.com/NethermindEth/warp/blob/develop/src/ast/visitor.ts#L87](https://github.com/NethermindEth/warp/blob/develop/src/ast/visitor.ts#L87) as pointed out by @rodrigo-pino. I prefer naming methods as `node.nodeType` (without the create prefix) and map them as it is implemented above

src/solCompile.ts
```diff
@@ -55,17 +52,17 @@ type SolcInput = {
     };
   };
   settings?: {
-    viaIR: boolean;
+    optimizer?: any;
```
swapnilraj: what's happening here?

piwonskp: Strange, I've only merged origin/main

piwonskp: Oh sure, I should have merged origin/develop probbably 😅

piwonskp: Unfortunately git revert made the PR unreadable. I guess I will reconstruct the branch and open a new PR

piwonskp:
Replaced by #776 to have a clearer diff

src/passes/inlineAssemblyTransformer.ts
swapnilraj: I think we should keep the create prefix, because its a verb instead of just the nodeType which is a noun.
The verb makes its obvious what the function will do

cicr99: I prefer your current approach rather than the one we use in `ASTVisitor` with all of the `if`s. I do think that the whole project should be consistent, so I would say to change the approach on `ASTvisitor` for this one (I don't see atm any reason why we couldn't do so), but that should be left to do in a different PR.

About the way you do the mapping, I like to use a prefix as well, but I would use something like `transform` instead of `create`. First, because it would then be similar to the other methods like `transformNode`. Second, because you are not actually "creating" a yulNode but "transforming" it into an ASTNode

PR #769, opened 2022-11-10
Fix source unit path for contracts in "include-path"

swapnilraj:


src/passes/sourceUnitPathFixer.ts
```diff
@@ -0,0 +1,31 @@
...
+      }
+    }
+  }
+
+  static map_(ast: AST, includePaths: string[]): AST {
+    ast.roots.forEach((root) => {
+      const mapper = new SourceUnitPathFixer(includePaths);
+      mapper.dispatchVisit(root, ast);
+    });
+    return ast;
```
rodrodros: Is the `map_` function needed? It has the same definition as `map` from `ASTMapper`

src/passes/sourceUnitPathFixer.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  visitSourceUnit(node: SourceUnit, _ast: AST): void {
+    if (fs.existsSync(node.absolutePath)) {
+      for (const prefix of this.includePaths) {
+        const filePath = path.join(prefix, node.absolutePath);
+        if (fs.existsSync(filePath)) {
+          node.absolutePath = filePath;
+          break;
+        }
```
rodrodros: Shouldn the most outer `if` condition be negated like:
```typescript
if (!fs.existSync(node.absolutePath)) { ... }
```
Would it be very complex to write a test to check this?

GuyMcComb:
@AlejandroLabourdette Looks like this includes your work?

src/index.ts
```diff
@@ -20,7 +20,7 @@ import { runTests } from './testing';
 import { postProcessCairoFile } from './utils/postCairoWrite';
 
 export type CompilationOptions = {
-  warnings: boolean;
+  warnings?: boolean;
```
AlejandroLabourdette, line 23: Why do we make this property nullable? Isn't the default behavior just to take it as false? If that's the case we avoid "options.warnings || false" further in the code

AlejandroLabourdette:
> @AlejandroLabourdette Looks like this includes your work?

Ye, it does

src/passes/sourceUnitPathFixer.ts
swapnilraj: Yea its needed because I need to pass the includePaths, this was the fastest way to do it. I don't think there are many better ways

src/index.ts
swapnilraj: I had to make nullable to satisfy the typechecker, in testing.ts we pass some options we to transpile which don't have the warning key; making it nullable makes it okay to not have a warnings option

src/passes/sourceUnitPathFixer.ts
swapnilraj: I agree we should have tests, I was looking at our cli tests and they are just for input encoding/decoding. We don't tests that invoke from the cli

GuyMcComb: The majority of the work has been done in this PR: https://github.com/NethermindEth/warp/pull/662

AlejandroLabourdette (approved):

GuyMcComb (approved):

GuyMcComb:
@AlejandroLabourdette will this work without the user passing additional CLI options? If not, are there some default arguments we can include?

swapnilraj:
> @AlejandroLabourdette will this work without the user passing additional CLI options? If not, are there some default arguments we can include?

I will include a little utility to search for the node modules folder as a default arg but its a bit tricky since I need to find the root folder of the project for that.
Let's merge this as is and then I can make another PR to add the default node_modules arg

AlejandroLabourdette:
> @AlejandroLabourdette will this work without the user passing additional CLI options? If not, are there some default arguments we can include?

We can transpile a file without specifying any of these options but then no path would be included. If our contract import from a path that is not inside the directory we are executing it will prompt an error, otherwise it will run as before these changes.
About using a default value for include-paths we need to make some assumptions about the path where we are calling warp. An example:
```
     dirA
         \_ dirB
             \_ dirC
                 \_ file_to_execute 
             \_ directory_to_include
                 \_ files_to_include
```
If we call warp from dirB then the path to include is "directory_to_include", but if we call it from dirA then the path to included would be "dirB/directory_to_include".

PR #770, opened 2022-11-11
fix implicit issue: uint256(addr)

rjnrohit:
-------   **BUG FIX** -------
this contract
```solidity
pragma solidity ^0.8.14;

contract temp {
    function address2uint256(address _addr) public returns (uint256) {
        return uint256(_addr);
    }
    function ad2uint(address _addr) public returns (uint256) {
        return address2uint256(_addr);
    }
}
```
was produces invalid cairo code


rodrodros (approved):

example_contracts/addr2uint256.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity ^0.8.14;
+
+contract WARP {
+    function address2uint256Util(address _addr) public returns (uint256) {
+        return uint256(_addr);
+    }
+    function address2uint256(address _addr) public returns (uint256) {
+        return address2uint256Util(_addr);
+    }
+}
```
GuyMcComb: extra line

tests/behaviour/contracts/conversions/addr2uint256.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity ^0.8.14;
+
+contract WARP {
+    function address2uint256Util(address _addr) public returns (uint256) {
+        return uint256(_addr);
+    }
+    function address2uint256(address _addr) public returns (uint256) {
+        return address2uint256Util(_addr);
+    }
+}
```
GuyMcComb: extra line

GuyMcComb:
Looks good to go!
Just add the lines

example_contracts/addr2uint256.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity ^0.8.14;
```
cicr99: I don't think we need this test case in here, since it's also being added in the behavior tests. The expectation is for it to be `'Successful'`, the behavior test already checks it doesn't fail at transpilation or compilation, so this would be checked twice. Other than that, it looks good to me.

rjnrohit: Yeah, Github CI runs will check this twice but the idea is to do syntax check on the output when a person only runs `bin/warp test` or if they only run`yarn test:behaviour`

JorikSchellekens (approved):
Don't think we need the tests in `example_contracts`. As Carmen pointed out it's being checked twice. That said I appreciate the sentiment. I would like it if the example_contracts directory was repurposed to only have interesting examples for a user of warp and all the tests were made into 'behaviour tests'. The behavior tests should just be renamed to 'tests'. Anyway g2g otherwise

cicr99 (approved):

PR #771, opened 2022-11-11
[BUG FIX] : add missing import

rjnrohit:
`tests/behaviour/contracts/inheritance/constructors/order_of_eval.sol` was producing cairo code with missing import.

### DIFF
--- a/src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
+++ b/src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
@@ -167,6 +167,9 @@ export class DynArrayPushWithArgGen extends StringIndexedFuncGen {
     });
     this.requireImport('starkware.cairo.common.uint256', 'Uint256');
     this.requireImport('starkware.cairo.common.uint256', 'uint256_add');
+    this.requireImport('starkware.cairo.common.cairo_builtins', 'BitwiseBuiltin');
+    this.requireImport('starkware.cairo.common.dict_access', 'DictAccess');
+    this.requireImport('starkware.cairo.common.cairo_builtins', 'HashBuiltin');
     return funcName;
   }
 }


GuyMcComb (approved):

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
```diff
@@ -167,6 +167,9 @@ export class DynArrayPushWithArgGen extends StringIndexedFuncGen {
     });
     this.requireImport('starkware.cairo.common.uint256', 'Uint256');
     this.requireImport('starkware.cairo.common.uint256', 'uint256_add');
+    this.requireImport('starkware.cairo.common.cairo_builtins', 'BitwiseBuiltin');
+    this.requireImport('starkware.cairo.common.dict_access', 'DictAccess');
+    this.requireImport('starkware.cairo.common.cairo_builtins', 'HashBuiltin');
```
rodrodros, line 172: The bug here is not missing imports, but the consequence of a bigger problem which is that generated functions always get printed no matter if their definition is deleted by the function pruner. (This applies to all generated warp functions). Which contract threw this error? General pattern is:
```solidity
function generated_function() { ... }
function f() private { 
    generated_function()
}
```
`f` gets removed in function pruner and `generated_function` does as well. Then annotate implicit pass is run, but since neither function exists no builtin import is added. The odd parts is when we write the Cairo code, because we handle generated functions different as normal functions they'll get printed anyways and if they were the only function that required certain import it won't be in the transpiled contract. And because of this, there is an error during on compilation.

I think the optimal solution would be to change part of the project structure and link generated function strings with their definitions (instead of handling separately). This is not completely straightforward to implement because there are some edge cases (like generated functions of generated functions) and there is a lot to change (all `cairoUtilFuncWriter`!).

I believe that a better patch for the moment would be non-deleting generated functions, since they'll definition will still exists, when the annotate pass gets executed it will include the missing imports and we will produce cairo code with unused code (which we already do) but without any of the bugs!

Edit: Improve explanation

rodrodros (changes requested):

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
swapnilraj: Thanks for the explanation, this bug has reared its head a few times now. I think its the right time to tackle it properly maybe we make a new node for generated function definitions and attach the source code for the function in the node so that we can manipulate it properly. @rodrigo-pino how does this sound to you?

rodrodros: Yes, I think that is the best approach, and I think  this new node could be used as well for cairo stubs, which we currently handle some times as solidity function and other times as cairo stubs (and I think it should be always handled as cairo stubs)

rjnrohit:
Added a ticket to the notion page : https://www.notion.so/nethermind/fix-generated-functions-pruner-bug-eb9bd85066284e5689089d10dee705ea

PR #775, opened 2022-11-14
ignore indexed arg

rjnrohit:
Ignore indexed event: Ignore the fact that args are indexed

### DIFF
--- a/src/passes/warnSupportedFeatures.ts
+++ b/src/passes/warnSupportedFeatures.ts
@@ -9,9 +9,11 @@ import {
   FunctionCall,
   FunctionCallKind,
   FunctionCallOptions,
+  ParameterList,
   NewExpression,
   parseSourceLocation,
   UserDefinedTypeName,
+  SourceUnit,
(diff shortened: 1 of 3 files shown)


rodrodros (commented):
I think it would be useful to throw a warning (in the warning pass) that indexed parameters are being omitted until proper support is added

GuyMcComb (approved):

src/passes/warnSupportedFeatures.ts
```diff
@@ -84,6 +96,12 @@ export class WarnSupportedFeatures extends ASTMapper {
       [...deploySalts.entries()].forEach(([path, nodes]) => warn(path, nodes));
     }
 
+    if (indexedArgs.size > 0) {
+      console.log(
+        `${warning('Warning:')} Indexed parameters are not supported yet. They will be ignored.`,
```
GuyMcComb: *indexing will be ignored

GuyMcComb (approved):

swapnilraj (approved):

PR #776, opened 2022-11-14
Inline assembly arithmetic operations

piwonskp:
Replaces #768


src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+}
+
+export class InlineAssemblyTransformer extends ASTMapper {
+  visitInlineAssembly(node: InlineAssembly, ast: AST): void {
+    assert(node.yul != undefined, 'Attribute yul of the InlineAssembly node is undefined');
+    const verifier = new YulVerifier();
+    node.yul.statements.map((node: YulNode) => verifier.verifyNode(node));
+
+    if (verifier.errors.length) {
```
swapnilraj: nit:  we try to be explicit and not rely on JS's truthy/falsy semantics so length > 0 would be better

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,159 @@
...
+
+class YulVerifier {
+  errors: string[] = [];
+
+  verifyNode(node: YulNode) {
+    const methodName = `create${node.nodeType}`;
+    if (!(methodName in YulTransformer.prototype))
+      this.errors.push(`${node.nodeType} is not supported`);
+    const method = this[methodName as keyof YulVerifier] as (node: YulNode) => void;
+    if (method != undefined) return method.bind(this)(node);
```
swapnilraj: we should replace all != with their strict counterpart !==
you can add this rule https://eslint.org/docs/latest/rules/eqeqeq and make it error so that they are not added in the future

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.ast = ast;
+    const blockRefIds = assemblyRoot.externalReferences.map((ref) => ref.declaration);
+    assert(assemblyRoot.context !== undefined, `Assembly root context not found.`);
+    const blockRefs = [...assemblyRoot.context.map].filter(([id, _]) =>
+      blockRefIds.includes(id),
+    ) as [number, VariableDeclaration][];
+    this.vars = new Map(blockRefs.map(([_, ref]) => [ref.name, ref]));
+  }
+
+  toSolidityStatement(node: YulNode): Statement {
```
piwonskp, line 103: I've used an explicit naming. It could be prefixed with `build` or `convert`

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,157 @@
...
+    return this.toSolidityNode(node);
+  }
+
+  toSolidityExpr(node: YulNode, ...args: unknown[]): Expression {
+    return this.toSolidityNode(node, ...args) as Expression;
+  }
+
+  toSolidityNode(node: YulNode, ...args: unknown[]): ASTNode {
+    const methodName = `transform${node.nodeType}`;
```
piwonskp: Maybe it should be prefixed `visit` and use the same method names across transformer and verifier? WYT?

piwonskp: Would be less explicit but more unified and consistent, also with the rest of the codebase

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,176 @@
...
+  VariableDeclaration,
+  UncheckedBlock,
+  Literal,
+  Statement,
+  ASTContext,
+} from 'solc-typed-ast';
+import assert from 'assert';
+import { getErrorMessage, WillNotSupportError } from '../utils/errors';
+
+const binaryOps: { [name: string]: string } = { add: '+', mul: '*', div: '/', sub: '-' };
```
swapnilraj: We are missing operations like shr shl etc
https://docs.soliditylang.org/en/v0.8.17/yul.html#evm-dialect

piwonskp:
Features missing as of `c1eac4929c82ccdfb636208370d55cf0432afb9a`: 
```solidity
contract C {
    function g(uint x) public pure returns (uint y) {
        uint8 a = 1;
        assembly {
            y := add(1, add(2, a))
        }
    }
}
```

src/passes/inlineAssemblyTransformer.ts
GuyMcComb: No issue here.

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -87,10 +86,6 @@ export class RejectUnsupportedFeatures extends ASTMapper {
     this.visitExpression(node, ast);
   }
 
-  visitInlineAssembly(node: InlineAssembly, _ast: AST): void {
-    this.addUnsupported('Yul blocks are not supported', node);
```
GuyMcComb, line 90: Could we check for YUL operators searching for mem and storage?

piwonskp: With YulNodes implemented we would probably move the whole verifier here. I don't think it's a good idea to reject memory and storage functions here and have verification partitioned between unsupported features pass and the verifier. The idea of the verifier was to print all errors regarding the inline assembly block at once

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,176 @@
...
+    let lhs: Expression;
+    if (node.variableNames.length === 1) {
+      lhs = this.transformYulIdentifier(node.variableNames[0]);
+    } else {
+      lhs = createTuple(
+        this.ast,
+        node.variableNames.map((i: YulNode) => this.transformYulIdentifier(i)),
+      );
+    }
+    const rhs = this.toSolidityExpr(node.value, lhs.typeString);
```
piwonskp, line 183: Seems a little hackish to me. Is there a better solution? What about expressions that are not a part of an assignment?

src/passes/inlineAssemblyTransformer.ts
piwonskp: Ignore it for now, wait for the YulNode PR. Once it's completed it will need to be refactored anyway

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,239 @@
...
+      this.errors.push(`${node.functionName.name} is not supported`);
+    node.arguments.forEach((arg: YulNode) => this.verifyNode(arg));
+  }
+}
+
+class YulTransformer {
+  vars: Map<string, VariableDeclaration>;
+  ast: AST;
+
+  constructor(ast: AST, externalReferences: any[], context: ASTContext) {
```
rodrodros, line 93: Is it possible to specify what `externalReference` type is going to be. Or what do you expect it to be?

piwonskp: A more precise type is only possible after filtering which is done in the constructor currently. The type is the same as in [InlineAssembly.externalReferences](https://github.com/ConsenSys/solc-typed-ast/blob/master/src/ast/implementation/statement/inline_assembly.ts#L12)

rodrodros (approved):

src/utils/nodeTemplates.ts
```diff
@@ -138,6 +138,13 @@ export function createEmptyTuple(ast: AST): TupleExpression {
   return node;
 }
 
+export function createTuple(ast: AST, nodes: Expression[]): TupleExpression {
+  const typeString = `tuple(${nodes.map((node) => node.typeString)})`;
+  const node = new TupleExpression(ast.reserveId(), '', typeString, false, nodes, nodes);
```
cicr99: Why are you passing `nodes` as the value for the `raw` parameter? I'm not sure about what raw stores. Usually when creating a new node and there is no other node from where to extract its raw value, this is just left as undefined.

src/utils/nodeTemplates.ts
```diff
@@ -138,6 +138,13 @@ export function createEmptyTuple(ast: AST): TupleExpression {
   return node;
 }
 
+export function createTuple(ast: AST, nodes: Expression[]): TupleExpression {
```
cicr99, line 139: Since there is this function now, I don't think we need to also have `createEmptyTuple` as the only difference would passing an empty list of nodes

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,239 @@
+import { ASTMapper } from '../ast/mapper';
```
cicr99, line 1: Should this file be added to the exports file as well?

src/passes/inlineAssemblyTransformer.ts
```diff
@@ -0,0 +1,239 @@
...
+
+class YulVerifier {
+  errors: string[] = [];
+
+  verifyNode(node: YulNode) {
+    const transformerName = generateTransformerName(node.nodeType);
+    if (!(transformerName in YulTransformer.prototype))
+      this.errors.push(`${node.nodeType} is not supported`);
+
+    const verifierName = `check${node.nodeType}`;
```
cicr99, line 71: Should we use here a method like `generateTransformerName`? Maybe this method can also take the prefix as a parameter and then be used for both cases

piwonskp: Let's just leave it as it is. Major changes are coming to YulNodes so it will be dropped anyway in the near future

cicr99 (approved):

PR #782, opened 2022-11-18
Indexed Events

rjnrohit:
support indexed events

The logs will contain array (keys and data) of 248 bit numbers ( 31 byte felts packed into a felt).


warplib/dynamic_arrays_util.cairo
```diff
@@ -231,3 +231,32 @@ func bytes_upper_bound{range_check_ptr: felt}(number: felt) -> (upper_bound: fel
...
     return (32 + result,);
 }
+
+// Utility functions for indexed events
+
+// Memory felt array to felt array copy
+// ---- mem_loc: warp memory location of the array
+// ---- array: felt array to write to
+// ---- index : the location at which to start writing to array
+func memory_array_to_felt_array{range_check_ptr: felt, warp_memory: DictAccess*}(
```
swapnilraj, line 241: https://github.com/NethermindEth/warp/blob/develop/warplib/memory.cairo#L357
we have an implementation of this function

swapnilraj: https://github.com/NethermindEth/warp/blob/develop/warplib/memory.cairo#L357
This function is very similar to your implementation, can you make the one in memory.cairo a wrapper around your implementation

src/cairoUtilFuncGen/abi/indexEncode.ts
```diff
@@ -0,0 +1,427 @@
...
+      `  alloc_locals;`,
+      `  let bytes_index : felt = 0;`,
+      `  let (bytes_array : felt*) = alloc();`,
+      ...encodings,
+      `  let (max_length256) = felt_to_uint256(bytes_index);`,
+      `  let (mem_ptr) = wm_new(max_length256, ${uint256(1)});`,
+      `  felt_array_to_warp_memory_array(0, bytes_array, 0, mem_ptr, bytes_index);`,
+      `  return (mem_ptr,);`,
+      `}`,
+    ].join('\n');
```
piwonskp, line 78: Template literals may be multiline

src/cairoUtilFuncGen/abi/indexEncode.ts
```diff
@@ -0,0 +1,427 @@
...
+    this.requireImport('starkware.cairo.common.alloc', 'alloc');
+    this.requireImport('starkware.cairo.common.cairo_builtins', 'BitwiseBuiltin');
+    this.requireImport('starkware.cairo.common.uint256', 'Uint256');
+    this.requireImport('warplib.maths.utils', 'felt_to_uint256');
+    this.requireImport('warplib.memory', 'wm_new');
+    this.requireImport('warplib.dynamic_arrays_util', 'felt_array_to_warp_memory_array');
+
+    const cairoFunc = { name: funcName, code: code };
+    this.generatedFunctions.set(key, cairoFunc);
+    return cairoFunc.name;
```
piwonskp, line 89: The method could use some split imo. E.g imports block could be placed in its own `addImports` method

src/cairoUtilFuncGen/abi/indexEncode.ts
```diff
@@ -0,0 +1,427 @@
...
+  ): string {
+    const funcName = this.getOrCreateEncoding(type, padding);
+    if (isDynamicallySized(type, this.ast.compilerVersion) || isStruct(type)) {
+      return [
+        `let (${newIndexVar}) = ${funcName}(`,
+        `  bytes_index,`,
+        `  bytes_array,`,
+        `  ${varToEncode}`,
+        `);`,
+      ].join('\n');
```
piwonskp, line 145: Multiline

src/cairoUtilFuncGen/abi/indexEncode.ts
```diff
@@ -0,0 +1,427 @@
...
+      assert(type.size !== undefined);
+      return [
+        `let (${newIndexVar}) = ${funcName}(`,
+        `  bytes_index,`,
+        `  bytes_array,`,
+        `  0,`,
+        `  ${type.size},`,
+        `  ${varToEncode},`,
+        `);`,
+      ].join('\n');
```
piwonskp, line 159: Same

src/cairoUtilFuncGen/event.ts
```diff
@@ -0,0 +1,186 @@
...
+          else
+            keysInsertions.push(
+              this.generateComplexEncodingCode(paramType, 'keys', `param${index}`),
+            );
+        } else {
+          dataInsertions.push(this.generateSimpleEncodingCode(paramType, 'data', `param${index}`));
+        }
+        return [params, keysInsertions, dataInsertions];
+      },
+      [new Array<{ name: string; type: string }>(), new Array<string>(), new Array<string>()],
```
piwonskp, line 111: A lot of ifs, elses and nested functions. Perhaps it could be extracted to its own function? Consider that you can transform any if-else to:
```
() => {
  if {
    return;
  }
  return;
}
```

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: I'm just following the standard that rodrigo has used for abiEncode implementation

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: multline template string may introduce extra whitespaces at the beginning that we want to avoid.

src/cairoUtilFuncGen/event.ts
rjnrohit: will do

src/cairoUtilFuncGen/abi/indexEncode.ts
RohitRanjangit: same convention

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: can you elaborate more about it?

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: multline template string may introduce extra whitespaces at the beginning that we want to avoid.

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: multline template string may introduce extra whitespaces at the beginning that we want to avoid. And even identation would look bad 
```js
        node.canonicalSignature(ABIEncoderVersion.V2),
      ),
      ...keysInsertions,
`   // keys: pack 31 byte felts into a single 248 bit felt,
   let (keys_len: felt, keys: felt*) = pack_bytes_felt(${BYTES_IN_FELT_PACKING}, ${BIG_ENDIAN}, keys_len, keys);,
   // data arrays,
   let data_len: felt = 0;,
   let (data: felt*) = alloc();`,
      ...dataInsertions,
      `   // data: pack 31 bytes felts into a single 248 bits felt`,
      `   let (data_len: felt, data: felt*) = pack_bytes_felt(${BYTES_IN_FELT_PACKING}, ${BIG_ENDIAN}, data_len, data);`,
      `   emit_event(keys_len, keys, data_len, data);`,
      `   return ();`,
      `}`,
```

src/cairoUtilFuncGen/abi/indexEncode.ts
rjnrohit: I got it, this is  also what I was thinking to add  , but since there is not much of repetitive import locations , so I avoided it.

src/cairoUtilFuncGen/abi/indexEncode.ts
piwonskp: Good point. We could think of using https://www.npmjs.com/package/endent or https://www.npmjs.com/package/common-tags to fix the issue.

rjnrohit: we can do it in a separate PR. Because we've to change the template string standard throughout the code base. can you create a notion ticket for it?

piwonskp: I agree, but I don't want to be hasty. First we have to discuss that with @rodrigo-pino if he has established the convention :) I don't want to arbitrarily decide on code style conventions

piwonskp: I'll create notion ticket if Rodrigo also agrees that is the preferred solution

piwonskp (approved):

src/cairoUtilFuncGen/abi/indexEncode.ts
rodrodros: @piwonskp  @RohitRanjangit  I did it like that because I was following the existing  conventions too :sweat_smile: , `cairoUtilFuncWriter`  is fully written like this.

I don't mind changing it if it would improve readability. My only two drawbacks are the need for an extra library and that there is a lot to change :hot_face: 

Also, I find it pretty readable as it is, but I've done considerable work in this area and I am surely biased.

AlejandroLabourdette (approved):

PR #783, opened 2022-11-18
Fix collision of identifiers with contract definition

swapnilraj:
- [x] Sem Tests passed


src/passes/identifierManglerPass/declarationNameMangler.ts
```diff
@@ -120,8 +121,24 @@ export class DeclarationNameMangler extends ASTMapper {
...
+    const parentScope = node.vScope;
+
+    if (parentContract !== undefined) {
+      return (
+        parentContract.name === node.name ||
+        parentScope
+          .getChildrenByType(Identifier, true)
+          .some((identifier) => identifier.name === node.name)
+      );
+    }
```
piwonskp: Imo `if (parentContract === undefined) return false;` would be more straightforward than inequality comparison

PR #784, opened 2022-11-21
Base path fix

rjnrohit:
map file location relative to current working directory
- [X] Sem Tests Pass


swapnilraj:
@GuyMcComb can we run the semantic tests as well

src/index.ts
```diff
@@ -94,7 +94,12 @@ program
       options.basePath = options.basePath || defaultBasePath;
     }
 
-    const ast = compileSolFiles(files, options);
+    // map file location relative to current working directory
+    const mFiles = files.map((file) => {
+      return path.relative(process.cwd(), file);
+    });
```
piwonskp: Can be shortened by removing brackets and return
```suggestion
    const mFiles = files.map(file => path.relative(process.cwd(), file));
```

rjnrohit: great, I'll do

piwonskp (approved):

tests/behaviour/test_calldata.ts
```diff
@@ -5165,7 +5375,7 @@ const tests: { [file: string]: ITestCalldata[] | null } = {
           '0x000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000003',
         expectations: '0x',
         failure: false,
-        signature: 'constructor(uint256[3])',
```
GuyMcComb: Whats going on here @RohitRanjangit

rjnrohit: Nothing

GuyMcComb (approved):

PR #792, opened 2022-11-29
Fix curly braces in structured documentation

piwonskp:


src/cairoWriter.ts
```diff
@@ -394,6 +394,12 @@ function writeContractInterface(node: ContractDefinition, writer: ASTWriter): Sr
   const functions = node.vFunctions.map((v) => {
     const resultLines = writer
       .write(v)
+      /* TODO: It's a quickfix for now. Implement that as a pass maybe.
```
rjnrohit, line 397: may be we can use `/// warp-cairo end` as flag in the comments that denotes end of the stub part from the comments. This way we can distinguish the comments and stub parts

piwonskp: If we want to keep original comments then ofc, but right now it seems that we don't do it so a pass that remove any StructuredDocumentation from Interface would suffice. Not sure if we want to keep this behaviour in the long term though. That's one of the reasons why I decided to abandon implementing it in pass for now

rodrodros (approved):

rjnrohit (approved):

PR #794, opened 2022-11-30
Minor improvements to `TypeInformationCalculator`

rodrodros:
- Fix a bug with type info of contracts where it didn't throw an error
- Always throw an error when there is `type()` case which is not being handled


piwonskp (approved):

GuyMcComb:
Got some `.any` type warnings in there.

rjnrohit (approved):

rodrodros:
> Got some `.any` type warnings in there.

They are from unrelated code. I would suggest a separate PR for fixing those.

PR #796, opened 2022-12-05
Add semantic tests batches to Github CI and fix implicit issue indexed events

rjnrohit:
This PR introduces the semantic test CI runner on Github, from now onwards it doesn't require manual effort to run semantic tests on AWS. Github CIs runs semantic tests in three batches to reduce the overall execution time.  

- [x] It passed 
- [x] See if Github CI semantic runner is capable of catching errors during semantic tests  execution

For that I deliberately introduced a error 
when I did: 
Screenshot 2022-12-12 at 6 46 06 PM

It actually fails,  proof: Screenshot 2022-12-12 at 6 42 02 PM

Hence, Github CI semantic runner is working as expected.


FalcoXYZ (approved):
LGTM!

.github/workflows/semantic.yml
```diff
@@ -65,10 +127,16 @@ jobs:
...
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+          sudo apt install -y gcc-11 gcc-11-multilib g++-11 g++-11-multilib
+          gcc-11 --version
+
       - name: Test code execution
         run: |
-          yarn testnet:benchmark &
-          node build/autoRunSemanticTests.js
+          yarn testnet:benchmark_semantic &
+          SM_BATCH=3 node build/autoRunSemanticTests.js
```
piwonskp: No need to copy-paste code. It is possible to reuse code via custom actions, reusable workflows or matrices. In this case I'd suggest using matrix

rjnrohit: good point! done

.github/workflows/semantic.yml
```diff
@@ -65,63 +42,19 @@ jobs:
...
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+          sudo apt install -y gcc-11 gcc-11-multilib g++-11 g++-11-multilib
+          gcc-11 --version
+
       - name: Test code execution
         run: |
-          yarn testnet:benchmark &
-          node build/autoRunSemanticTests.js
+          yarn testnet:benchmark_semantic &
+          SM_BATCH=${{ matrix.SM_BATCH }} node build/autoRunSemanticTests.js
```
piwonskp: There's `env` property that can be used to set env variables

rjnrohit: done!

piwonskp (approved):

PR #803, opened 2022-12-13
Conditionals and operators

agbanusi:


rjnrohit (commented):
Please add behaviour tests for this.

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -99,7 +99,7 @@ function evaluateLiteral(node: Literal): RationalLiteral | boolean | null {
 
 function evaluateUnaryLiteral(node: UnaryOperation): RationalLiteral | boolean | null {
   const op = evaluateLiteralExpression(node.vSubExpression);
-  if (op === null) return null;
+  if (!op) return null;
```
rodrodros: Nitpick: could you revert this change back. I believe that by explicitly stating that you want it to be different than null makes the code safer and more readable

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,16 +128,15 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+    const notNullMember = left ?? right;
     if (notNullMember === null) {
       return null;
-    }
-    if (typeof notNullMember === 'boolean') {
+    } else if (typeof notNullMember === 'boolean') {
       switch (node.operator) {
         case '&&': // false && x = false
-          return notNullMember ? null : false;
+          return notNullMember && false;
```
cicr99: This should be changed back to the way it was before. The way this is right now always returns false, which is unwanted if `notNullMember` is true

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,16 +128,15 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
         case '||': // true || x = true
-          return notNullMember ? true : false;
+          return notNullMember || true;
```
cicr99: Same thing here, this always returns true

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -236,6 +226,14 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
       case '>>':
         return left.shiftRight(right);
+      case '&':
+        return left.and(right);
+      case '|':
+        return left.or(right);
+      case '^':
+        return left.xor(right);
+      case '~':
+        return left.not();
```
cicr99: This operator can be removed from this switch because `~` is a unary operator, and this is for binary operators

src/passes/literalExpressionEvaluator/rationalLiteral.ts
```diff
@@ -137,6 +137,28 @@ export class RationalLiteral {
...
+  }
+  xor(other: RationalLiteral): RationalLiteral {
+    return new RationalLiteral(
+      this.numerator ^ other.numerator,
+      this.denominator ^ other.denominator,
+    );
+  }
+  not(): RationalLiteral {
+    return new RationalLiteral(~this.numerator, this.denominator);
+  }
```
cicr99: This functionality is already defined here as `bitwiseNegate`, so no need for this, it can be removed.

It is also a fine idea to add the "bitwise" prefix to these functions as in: `bitwiseAnd`, `bitwiseOr`, `bitwiseXor`; so it can be different from the logical operators

tests/behaviour/contracts/expressions/rationalLiterals.sol
```diff
@@ -52,4 +52,20 @@ contract WARP {
...
+    return 18|6;
+  }
+
+  function xor() pure public returns (int8) {
+    return 18^6;
+  }
+
+  function not() pure public returns (int8) {
+    return ~3;
+  }
```
cicr99: This can be removed as well, since there is a case above for it

cicr99 (changes requested):

agbanusi:
> Please add behaviour tests for this.

Done this

cicr99 (approved):

piwonskp (approved):

rodrodros (approved):

PR #804, opened 2022-12-16
Re-structure cairo writer

rodrodros:
Split cairoWriter script (over 1000 lines) into several scripts


rjnrohit (approved):

AlejandroLabourdette (approved):

piwonskp:
Ci does not pass. Btw consider a directory structure similiar to [solc-typed-ast](https://github.com/ConsenSys/solc-typed-ast/tree/master/src/ast/implementation). Perhaps a separate directory for statements/expressions would be helpful?

rodrodros:
> Btw consider a directory structure similiar to [solc-typed-ast](https://github.com/ConsenSys/solc-typed-ast/tree/master/src/ast/implementation). Perhaps a separate directory for statements/expressions would be helpful?

I would generally agree, but in this case I think it's not that beneficial because writers are not that different at implementation level.

They all define a `writeInner` and a `writeOuter` methods and they don't receive  different treatments depending if they are declarations, statements or expressions.

src/cairoWriter/index.ts
```diff
@@ -0,0 +1,155 @@
+import { AST } from '../export';
```
cicr99: make sure to always import from the source file instead of the export file.

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -0,0 +1,84 @@
+import assert from 'assert';
+import { ASTWriter, ContractKind, SourceUnit, SrcDesc } from 'solc-typed-ast';
+import { getStructsAndRemappings, removeExcessNewlines, TEMP_INTERFACE_SUFFIX } from '../../export';
+import { CairoASTNodeWriter } from '../base';
+import { writeImports } from '../utils';
+
+// Used by CairoContractWriter
```
cicr99: `functionCallWriter` also uses `interfaceNameMappings`

cicr99 (commented):
Really nice work!!

src/cairoWriter/index.ts
cicr99: This happens on other files as well

piwonskp (approved):

PR #808, opened 2022-12-21
fix documentation broken links

AlejandroLabourdette:


docs/docs/features/interface_call_forwarder.mdx
```diff
@@ -209,6 +209,6 @@ starknet invoke \
 
 ## Running tests
 
-For more detailed and implemented steps, you can look at the [interface_forwarder.test.ts](../../tests/interface_call_forwarder/interface_forwarder.test.ts) file.
+For more detailed and implemented steps, you can look at the [interface_forwarder.test.ts](https://github.com/NethermindEth/warp/blob/develop/tests/interface_call_forwarder/interface_forwarder.test.ts) file.
```
piwonskp, line 212: Is there any version variable in docs that could be used instead of develop? Generally we shouldn't point to develop branch in docs as it changes over the time

AlejandroLabourdette: This documentation is from a feature added in November 16th so we don't have it already in main branch. Perhaps would be a good idea to create a ticket now to change the links once we release a new Warp version.

piwonskp (approved):

rodrodros (approved):

PR #810, opened 2022-12-29
Remove event mangling

swapnilraj:
Export decoding functions for warp event encoding/decoding

### DIFF
--- a/src/cairoUtilFuncGen/event.ts
+++ b/src/cairoUtilFuncGen/event.ts
@@ -113,7 +113,9 @@ export class EventFunction extends StringIndexedFuncGen {
 
     const topic: string = warpEventCanonicalSignaturehash(
       node.name,
-      node.vParameters.vParameters.map((param) => param.typeString),
+      node.vParameters.vParameters.map((param) =>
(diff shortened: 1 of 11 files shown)


src/cairoUtilFuncGen/event.ts
```diff
@@ -113,12 +111,10 @@ export class EventFunction extends StringIndexedFuncGen {
...
 
-    const topic: BigInt =
-      BigInt(
-        `0x${keccak('keccak256')
-          .update(node.canonicalSignature(ABIEncoderVersion.V2))
-          .digest('hex')}`,
-      ) & BigInt(MASK_250);
+    const topic: string = warpEventCanonicalSignaturehash(
+      node.name,
+      node.vParameters.vParameters.map((param) => param.typeString),
```
rjnrohit: It may contain `uint` , should be `uint256` instead

swapnilraj: The typestrings are canonicalised by solc

swapnilraj: It doesn't work for structs though; fixed it

AlejandroLabourdette (approved):

PR #813, opened 2023-01-02
Ast update 11.0.2

piwonskp:


src/passes/variableDeclarationExpressionSplitter.ts
```diff
@@ -82,9 +82,12 @@ export class VariableDeclarationExpressionSplitter extends ASTMapper {
       'Expected variables to be initialised when running variable declaration expression splitter (did you run variable declaration initialiser?)',
     );
 
-    const initialValueType = safeGetNodeType(initialValue, ast.compilerVersion);
+    const initialValueType = safeGetNodeType(initialValue, ast.inference);
 
-    if (!(initialValueType instanceof TupleType)) {
+    if (
+      !(initialValueType instanceof TupleType) ||
+      (node.raw !== undefined && node.raw.assignments.length === 1)
```
piwonskp: `TypeNameRemover` removes nulls from assignments thus the usage of `raw`. Would be nice to find a way to do that without using `raw`

src/passes/implicitConversionToExplicit.ts
```diff
@@ -626,13 +626,17 @@ function handleConcatArgs(node: FunctionCall, ast: AST) {
...
-    const type = safeGetNodeType(arg, ast.compilerVersion);
+    const type = safeGetNodeType(arg, ast.inference);
     if (type instanceof StringLiteralType) {
       insertConversionIfNecessary(arg, new BytesType(), arg, ast);
     } else if (type instanceof IntLiteralType) {
       assert(arg instanceof Literal);
       const signed = BigInt(arg.value) < 0;
       insertConversionIfNecessary(arg, new IntType(256, signed), arg, ast);
+    } else if (type instanceof TupleType) {
+      arg.typeString = `uint8[${type.elements.length}] memory`;
```
piwonskp: Not causing any test to fail. Hardcoding memory doesn't look like a good idea though

src/cairoUtilFuncGen/memory/memoryStruct.ts
```diff
@@ -35,7 +35,11 @@ export class MemoryStructGen extends StringIndexedFuncGen {
       name,
       structDef.vMembers.map((decl) => {
         assert(decl.vType !== undefined);
-        const type = typeNameToSpecializedTypeNode(decl.vType, DataLocation.Memory);
+        const type = typeNameToSpecializedTypeNode(
+          this.ast.inference,
+          decl.vType,
+          DataLocation.Memory,
+        );
```
rodrodros: nitpick: use `this.ast.inference` as a last argument, since we usually pass "constant" parameters at the end of the paremeter list. 

By "constant parameter" I mean that the value taken by that parameter will remain the same during the whole code execution, in this case, the same instance of `InferType`

src/passes/bytesConverter.ts
```diff
@@ -45,21 +45,22 @@ export class BytesConverter extends ASTMapper {
...
-    const typeNode = safeGetNodeType(node, ast.compilerVersion);
-    if (!(typeNode instanceof IntLiteralType)) {
-      node.typeString = generateExpressionTypeString(replaceBytesType(typeNode));
-    }
     this.commonVisit(node, ast);
+    const typeNode = safeGetNodeType(node, ast.inference);
+    if (typeNode instanceof IntLiteralType || typeNode instanceof StringLiteralType) {
+      return;
+    }
+    node.typeString = generateExpressionTypeString(replaceBytesType(typeNode));
```
rodrodros: Why change the instruction order?

src/passes/conditionalSplitter/utils/splitter.ts
```diff
@@ -65,7 +65,11 @@ export function splitTupleAssignment(
         typeNode = generalizeType(lhsElementType)[0];
         location = DataLocation.Memory;
       } else {
-        [typeNode, location] = generalizeType(rhsElementType);
+        let lhsLocation: DataLocation | undefined;
+        [typeNode, lhsLocation] = generalizeType(lhsElementType);
```
rodrodros: Why use the lhs type instead of the rhs `typeNode`?

src/passes/conditionalSplitter/utils/splitter.ts
```diff
@@ -65,7 +65,11 @@ export function splitTupleAssignment(
...
+        [, location] = generalizeType(rhsElementType);
+        // tupleAssignments.sol
+        location = location || lhsLocation;
```
rodrodros: If location is undefined is set later as `Default`. Why use `lhsLocation` instead?

src/passes/dropInterfaceFunctionBodies.ts
```diff
@@ -0,0 +1,19 @@
+import { Block, ContractDefinition, ContractKind, FunctionDefinition } from 'solc-typed-ast';
+import { AST } from '../ast/ast';
+import { ASTMapper } from '../ast/mapper';
+
+export class DropInterfaceFunctionBodies extends ASTMapper {
```
rodrodros: Why is this pass needed?

src/passes/enumConverter.ts
```diff
@@ -58,7 +59,11 @@ export class EnumConverter extends ASTMapper {
       (node.vExpression instanceof MemberAccess &&
         node.vExpression.vReferencedDeclaration instanceof EnumDefinition)
     ) {
-      node.vExpression.typeString = generateExpressionTypeString(replaceEnumType(tNode));
+      node.vExpression.typeString = generateExpressionTypeStringForASTNode(
+        ast.inference,
```
rodrodros: Same nit: ast.inference to the last position in parameter list

src/passes/implicitConversionToExplicit.ts
```diff
@@ -269,22 +269,23 @@ export class ImplicitConversionToExplicit extends ASTMapper {
   visitTupleExpression(node: TupleExpression, ast: AST): void {
     if (!node.isInlineArray) return this.visitExpression(node, ast);
 
-    const type = generalizeType(safeGetNodeType(node, ast.compilerVersion))[0];
+    const type = generalizeType(safeGetNodeType(node, ast.inference))[0];
 
     assert(
-      type instanceof ArrayType,
-      `Expected inline array ${printNode(node)} to be array type, got ${printTypeNode(type)}`,
+      type instanceof TupleType || type instanceof ArrayType,
```
rodrodros: What changed that tuple types needs to be included here?

src/passes/userDefinedTypesConverter.ts
```diff
@@ -39,30 +42,31 @@ export class UserDefinedTypesConverter extends ASTMapper {
 
   visitVariableDeclaration(node: VariableDeclaration, ast: AST): void {
     this.commonVisit(node, ast);
-    const replacementNode = replaceUserDefinedType(safeGetNodeType(node, ast.compilerVersion));
+    const replacementNode = replaceUserDefinedType(
+      ast.inference,
```
rodrodros: same nit: ast.inference to last

src/passes/userDefinedTypesConverter.ts
```diff
@@ -105,33 +109,32 @@ export class UserDefinedTypesConverter extends ASTMapper {
   }
 }
 
-function replaceUserDefinedType(type: TypeNode): TypeNode {
+function replaceUserDefinedType(inference: InferType, type: TypeNode): TypeNode {
+  const callSelf = partial(replaceUserDefinedType, inference);
```
rodrodros: To be compliant with previous nitpicks, I would suggest to change this line to:
```ts
const callSelf = (tp) => replaceUserDefinedType(tp, inference)
```

src/utils/nodeTypeProcessing.ts
```diff
@@ -281,29 +285,39 @@ export function isStorageSpecificType(
...
+): TypeNode {
   getContainingSourceUnit(node);
-  return getNodeType(node, version);
+  // if (node instanceof VariableDeclaration) {
+  //   return inference.variableDeclarationToTypeNode(node);
+  // }
+  // if (node instanceof TypeName) {
+  //   return inference.typeNameToTypeNode(node)
+  // }
+  // return inference.typeOf(node);
```
rodrodros, line 307: This logic?

src/utils/nodeTypeProcessing.ts
```diff
@@ -348,20 +362,23 @@ export function getPackedByteSize(type: TypeNode, version: string): number | big
...
   if (type instanceof TupleType) {
     return type.elements.reduce(sumMemberSize, 0n);
   }
 
   if (type instanceof UserDefinedType && type.definition instanceof StructDefinition) {
-    assert(version !== undefined, 'Struct byte size calculation requires compiler version');
+    assert(
+      inference.version !== undefined,
+      'Struct byte size calculation requires compiler version',
+    );
```
rodrodros: The old assert instruction was a leftover, is safe to remove it (no need to update it)

src/utils/typeStrings/ast/inaccessible_dynamic_type.ts
```diff
@@ -0,0 +1,11 @@
+import { TypeNode } from 'solc-typed-ast';
```
rodrodros, line 1: I wonder why `inaccesible_dynamic_type` and `module_type` are needed

src/utils/typeStrings/typeString_grammar.pegjs
```diff
@@ -0,0 +1,482 @@
+Start =
```
rodrodros, line 1: Why are some scripts related to string parser in Warp now? Weren't they defined in solc-typed-ast?

rodrodros (commented):
_

src/passes/dropInterfaceFunctionBodies.ts
piwonskp: Good catch. I've accidentally commited that

src/utils/nodeTypeProcessing.ts
piwonskp: It's the logic to run inferType instead of typestring parser. I left that so it would be easy to just uncomment that and run tests to check how the new implementation works with InferType as the transition between the two will take some time

src/utils/typeStrings/typeString_grammar.pegjs
piwonskp: Because they were removed from `solc-typed-ast` in favor of InferType class. The InferType class however seems to be so different from the typestring parser that it requires a lot of effort to port warp from the old type inference system to the new one. Thus this partial port and the plan to adjust warp to the usage of InferType gradually

src/utils/typeStrings/ast/inaccessible_dynamic_type.ts
piwonskp: The idea is to modify the typestring parser as little as possible to minimize the maintenance of it. As I said below it's a temporary solution, eventually the whole typeStrings directory should be removed

src/passes/implicitConversionToExplicit.ts
piwonskp: Good catch. It seems after the later changes it became obsolete and now it is a dead code. Need to analyze it a little bit more and run tests though. There is more of tuple related logic that I've added to the class.

src/passes/bytesConverter.ts
piwonskp: The condition without negations is more readable. Thus the change. However literal types have no children anyway so I suppose I can just put the commonVisit at the end to preserve the order

src/passes/conditionalSplitter/utils/splitter.ts
piwonskp: Removed as typestring generation is now fixed

src/passes/bytesConverter.ts
```diff
@@ -45,21 +45,22 @@ export class BytesConverter extends ASTMapper {
   }
 
   visitExpression(node: Expression, ast: AST): void {
-    const typeNode = safeGetNodeType(node, ast.compilerVersion);
-    if (!(typeNode instanceof IntLiteralType)) {
-      node.typeString = generateExpressionTypeString(replaceBytesType(typeNode));
+    const typeNode = safeGetNodeType(node, ast.inference);
+    if (typeNode instanceof IntLiteralType || typeNode instanceof StringLiteralType) {
+      return;
```
cicr99, line 49: If the condition is true, this `return` causes the children of `node` not to be recursively visited. In the previous version, `commonVisit` was executed either way. I guess in this case, since the node is a literal, it doesn't matter; but I think I like the old design better.

src/passes/implicitConversionToExplicit.ts
```diff
@@ -374,7 +372,9 @@ export function insertConversionIfNecessary(
   } else if (currentType instanceof ImportRefType) {
     return;
   } else if (currentType instanceof IntLiteralType) {
-    insertConversion(expression, generalisedTargetType, context, ast);
+    if (!(generalisedTargetType instanceof IntLiteralType)) {
```
cicr99: Why is this `if` needed?

src/passes/bytesConverter.ts
piwonskp: Could you explain what is the purpose of `this.commonVisit(node, ast);` in methods like that?

src/passes/implicitConversionToExplicit.ts
piwonskp: Good catch! It seems it's not needed anymore

src/passes/bytesConverter.ts
piwonskp: Honestly for me its the clearest solution. It explicitly states that literals are not affected by the method. It just filters them out immediately. Please also take a look at the [conversation](https://github.com/NethermindEth/warp/pull/813#discussion_r1064589050) where we've discussed other approaches.

piwonskp: I'd even think of filtering them out by `node instanceof Literal` instead of using typenodes. WYT?

cicr99: Yes, I understand that, I'm just saying in terms of design I don't like stopping the execution at some node without visiting its children, when it should traverse the entire AST, unless of course, the logic of what is being done breaks if you do so, but that is not the case here. The other thing is what you mentioned in the last comment, if there are other types of literals, why are we only checking those?

Anyway, I checked `generateExpressionTypeString` and `replaceBytesType` methods and I think none of them modifies the typestring of these nodes. That said, my preferred solution would be to remove the condition completely and execute `node.typeString = generateExpressionTypeString(replaceBytesType(typeNode));` on all nodes. It should work the same way.

If you still think that the best solution is to use the `return`, then I do like what you said about filtering them out using `node instanceof Literal` but we need to check first that none of the typestrings a literal node might have are modified by `generateExpressionTypeString` and `replaceBytesType` methods.

piwonskp: > The other thing is what you mentioned in the last comment, if there are other types of literals, why are we only checking those?

That is a good question. These nodes existed before and they weren't checked. ~It probably worked because - as you said - it doesn't matter if we filter them out or not.~ I've just ran tests.  `StringLiteralType` has to be filtered out. Not sure about IntLiteralType yet though.

 Anyway, since typestrings logic is going to be replaced soon anyway I don't think there is any sense in splitting hairs

cicr99 (approved):

src/cairoUtilFuncGen/event.ts
```diff
@@ -112,12 +120,9 @@ export class EventFunction extends StringIndexedFuncGen {
...
 
-    const topic: string = warpEventCanonicalSignaturehash(
-      node.name,
-      node.vParameters.vParameters.map((param) =>
-        param.canonicalSignatureType(ABIEncoderVersion.V2),
-      ),
-    );
+    // TODO: Replace signature hash with solc-typed-ast's version
+    // const topic: string = this.ast.inference.signatureHash(node, ABIEncoderVersion.V2);
+    const topic: string = signatureHash(this.ast.inference.signature(node, ABIEncoderVersion.V2));
```
rodrodros: TODO?

src/utils/getTypeString.ts
```diff
@@ -185,14 +188,54 @@ function instanceOfNonRecursivePP(type: TypeNode): boolean {
     type instanceof ImportRefType ||
     type instanceof IntLiteralType ||
     type instanceof IntType ||
-    type instanceof ModuleType ||
```
rodrodros, line 188: Why is this type not needed any more?

src/utils/nodeTypeProcessing.ts
```diff
@@ -39,18 +37,19 @@ import { printNode, printTypeNode } from './astPrinter';
...
 */
 export function getParameterTypes(functionCall: FunctionCall, ast: AST): TypeNode[] {
-  const functionType = safeGetNodeType(functionCall.vExpression, ast.compilerVersion);
+  const functionType = safeGetNodeType(functionCall.vExpression, ast.inference);
   switch (functionCall.kind) {
     case FunctionCallKind.FunctionCall:
       assert(
-        functionType instanceof FunctionType,
+        functionType instanceof FunctionLikeType,
         `Expected ${printNode(functionCall.vExpression)} to be FunctionType, got ${printTypeNode(
```
rodrodros: Update assert error to be `FunctionLikeType`

src/utils/nodeTypeProcessing.ts
```diff
@@ -73,7 +72,9 @@ export function getParameterTypes(functionCall: FunctionCall, ast: AST): TypeNod
       );
       const structDef = functionType.type.to.definition;
       assert(structDef instanceof StructDefinition);
-      return structDef.vMembers.map(variableDeclarationToTypeNode);
+      return structDef.vMembers.map(
+        ast.inference.variableDeclarationToTypeNode.bind(ast.inference),
+      );
```
rodrodros: Why does  the function needs to be bound to `ast.inference`

src/cairoUtilFuncGen/event.ts
piwonskp: I've added todo and a task to https://github.com/NethermindEth/warp/issues/818 for a later discussion as simple replacement as the commented one causes tests to fail. I don't know how many changes does it require or even if we should replace our hash with the native one thus I postponed the implementation and left a trace that there's a method like this. If you've got a straightforward answer then I'd be happy to implement it, but if that's more complicated then I'd see a separate PR for that

src/utils/getTypeString.ts
piwonskp: I think it was never needed. It is the only occurrence of the type in the codebase as far as I remember. It was moved to the test utils in solc-typed-ast so it won't be available once we remove the typestring parser anyway

src/utils/nodeTypeProcessing.ts
piwonskp: I've changed that to use map's second argument instead

src/cairoUtilFuncGen/event.ts
piwonskp: It seems that #857 addresses the topic so imo it should be done there

piwonskp: Ok, after consulting with @RohitRanjangit we cannot use solc-typed-ast's signatureHash for that so I'll just remove the TODO

rodrodros (approved):

PR #819, opened 2023-01-10
Cli test using starknet devnet

rjnrohit:
Add tests for warp cli. 

Test following  warp cli commands
```
1. declare
2. deploy
3. status
4. deploy_account
5. call
6. invoke
7. new_account
```
using a ERC20 based contract. The test can be found at [here](src/cli/tests/cli.test.ts).


.github/workflows/cli.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.9 # Cairo was tested with this version
+
+      - name: Install dependencies
+        run: |
+          yarn
+          make compile
+          pip3 install cairo-lang==0.10.3
```
swapnilraj: Isn't the cairo version supported by warp 10.1? Can you update warp's cairo lang dependency as well?

src/cli/tests/cli.test.ts
```diff
@@ -0,0 +1,381 @@
...
+
+describe('Manage starknet account', function () {
+  this.timeout(TIME_LIMIT);
+  it('should create a starknet account', async () => {
+    if (fs.existsSync(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'))) {
+      fs.unlinkSync(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'));
+    }
+
+    await sh(
+      `${warpVenvPrefix} starknet new_account --wallet ${starknet_wallet} --feeder_gateway_url ${gateway_url} --gateway_url ${gateway_url} --network ${network} --account_dir ${account_dir}`,
```
swapnilraj: we should make a new cli command in warp for this `new_account`

src/cli/tests/cli.test.ts
```diff
@@ -0,0 +1,381 @@
...
+    expect(starknet_open_zeppelin_accounts[network]['__default__']['deployed']).to.be.false;
+  });
+
+  it('mint wei to the generated account', async () => {
+    const starknet_open_zeppelin_accounts = JSON.parse(
+      fs.readFileSync(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'), 'utf-8'),
+    );
+    const address = starknet_open_zeppelin_accounts[network]['__default__']['address'];
+
+    const { stdout } = await sh(
```
swapnilraj: let's make mint a function similar to what you've done with starknet compile

src/cli/tests/cli.test.ts
```diff
@@ -0,0 +1,381 @@
...
+      if (err) {
+        reject(err);
+      } else {
+        resolve({ stdout, stderr });
+      }
+    });
+  });
+}
+
+function starknetCompile(
```
swapnilraj: replace with warp compile

src/cli/cli.ts
```diff
@@ -1,10 +1,10 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import { Command } from 'commander';
-import { replaceSuffix, isValidSolFile, outputResult } from './io';
-import { compileSolFiles } from './solCompile';
-import { handleTranspilationError, transform, transpile } from './transpiler';
-import { analyseSol } from './utils/analyseSol';
+import { replaceSuffix, isValidSolFile, outputResult } from '../io';
```
swapnilraj: Why do we need to make a new folder for cli?

rjnrohit: I wanted a separate github action for a cli

rjnrohit: Placing the file "cli.test.ts" in the "tests/" directory would cause the "build" job to run it, , that I don't want

.github/workflows/cli.yml
rjnrohit: `cairo-format` is buggy in `0.10.3`,It produces invalid cairo code with no `;` at the end for some lines of code.  we can do the dependency update  in a separate PR.

swapnilraj: Okay, can you make a task for that?

rjnrohit: sure, https://www.notion.so/nethermind/Cairo-dependency-update-0-10-3-2c9c261579054a3fafd8306d8a43f36c

src/cli/tests/cli.test.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  it('should deploy the starknet account', async () => {
+    const { stdout, stderr } = await sh(
+      `${warpBin} deploy_account --wallet ${starknet_wallet} --feeder_gateway_url ${gateway_url} --gateway_url ${gateway_url} --network ${network} --account_dir ${account_dir}`,
+    );
+
+    expect(stderr).to.be.empty;
+    expect(stdout).to.not.be.empty;
+
+    const contract_address = stdout.split('\n')[3].split(':')[1].trim();
```
swapnilraj: I think if make a function that takes the txHash and contractAddress and returns the full string and we can do a exact string match that will be a lot more readable than string splitting on new lines and colons.

swapnilraj: This applies to other string splitting matchers as well

RohitRanjangit: we are extracting txHash and contractAddress from the output itself

RohitRanjangit: Creating a function may not be the most efficient solution, as the output may vary based on the fee applied and the context of the message.

rjnrohit: Creating a function may not be the most efficient solution, as the output may vary based on the fee applied and the context of the message. Each time we would need to analyze the entire string and then compare the analyzed string to the expected format.

As sample output string looks as : 
```
Running starknet compile with cairoPath /Users/rohit/nmd/warp
Sending the transaction with max_fee: 0.001247 ETH (1246520000000000 WEI).
Invoke transaction for contract deployment was sent.
Contract address: 0x05a252845cba4af66f37d1c759027049a8592b8a7ad29db567d16e96fee24ea7
Transaction hash: 0x15e02734241afc4cd37578ff160c313fb41562446ad40ea92992b4c6e36fe84

```

swapnilraj: Is the `max_fee` not static in all the test cases?

swapnilraj: In that case, instead of string splitting, please convert it to a regex and use matchAll

rjnrohit: I'm not using max_fee, but anyway I'm adding some changes to improve readability and output string validation

.github/workflows/cli.yml
```diff
@@ -0,0 +1,52 @@
+name: CLI testing
```
piwonskp: Let's keep the name of the file consistent with the name of workflow

.github/workflows/cli.yml
```diff
@@ -0,0 +1,52 @@
+name: CLI
+
+on:
+  push:
+    branches:
+      - main
+      - develop
+      - warp-ts
```
rodrodros: Not strictly related to this PR, but `warp-ts` has been stale for months now. Should we still have it?

src/cli.ts
```diff
@@ -255,9 +276,21 @@ program
   )
   .option('--use_cairo_abi', 'Use the cairo abi instead of solidity for the inputs.', false)
   .option('--network <network>', 'StarkNet network URL.', process.env.STARKNET_NETWORK)
+  .option('--gateway_url <gateway_url>', 'StarkNet gateway URL.', process.env.STARKNET_GATEWAY_URL)
+  .option(
+    '--feeder_gateway_url <feeder_gateway_url>',
+    'StarkNet feeder gateway URL.',
+    process.env.STARKNET_FEEDER_GATEWAY_URL,
+  )
```
rodrodros, line 284: Why do we need the feeder gateway for deploy transactions?

tests/cli/cli.test.ts
```diff
@@ -0,0 +1,477 @@
...
+    );
+    const address = starknet_open_zeppelin_accounts[network]['__default__']['address'];
+
+    const { stdout } = await mintEthToAccount(address);
+
+    const response = JSON.parse(stdout);
+
+    expect(response).to.not.be.undefined;
+    expect(response['new_balance']).to.not.be.undefined;
+    expect(response['new_balance']).to.equal(1000000000000000000);
```
rodrodros: Isn't the first 
```ts
    expect(response['new_balance']).to.not.be.undefined;
```
redundant?

rodrodros (commented):
I saw that we have the same options repeated across several commands (like `feeder_gateway_url`). Is there an easy way to avoid that repetition?

.github/workflows/cli.yml
swapnilraj: done

src/cli.ts
swapnilraj: For all transactions which change the feeder_gateway is required so that we can point those calls to the devnet.

tests/cli/cli.test.ts
swapnilraj: yup removed

src/cli.ts
rjnrohit: https://shard-labs.github.io/starknet-devnet/docs/guide/Interaction

tests/cli/cli.test.ts
```diff
@@ -0,0 +1,398 @@
...
+      `${warpBin} call ${contractCairoFile} --function totalSupply --address ${contract_address} --network ${network} --gateway_url ${gatewayURL} --feeder_gateway_url ${gatewayURL} --wallet ${wallet} --account_dir ${accountDir}`,
+    );
+    expect(stderr).to.be.empty;
+    expect(stdout.split('\n')[1].trim()).to.be.equal('500');
+  });
+
+  it('get Total Supply call using cairo ABI', async () => {
+    expect(contract_address).to.not.be.undefined;
+    const { stdout, stderr } = await sh(
+      `${warpBin} call ${contractCairoFile} --use_cairo_abi --function totalSupply_18160ddd --address ${contract_address} --network ${network} --gateway_url ${gatewayURL} --feeder_gateway_url ${gatewayURL} --wallet ${wallet} --account_dir ${accountDir}`,
```
swapnilraj: Remove --use_cairo_abi; that is only meant for testing.
Using this will not test the encoding of the Solidity inputs

swapnilraj (changes requested):

tests/cli/cli.test.ts
rjnrohit: done!

swapnilraj (approved):

.github/workflows/build.yml
```diff
@@ -1,11 +1,10 @@
-name: Warp CI
+name: Behaviour tests
```
piwonskp: The workflow runs `yarn test` not `yarn test:behaviour` what is the difference between them? I thought yarn test runs all tests and `yarn test:whatever` only specific

.github/workflows/cli.yml
```diff
@@ -0,0 +1,54 @@
+name: CLI
```
piwonskp: Let's call that `CLI tests`

.github/workflows/build.yml
swapnilraj: I think test just runs behaviour, because it doesn't invoke semantic tests we have a different script for them.
And the cli tests are explicitly excluded from them because @RohitRanjangit wanted them to be different.
@RohitRanjangit thoughts, let's resolve this quick so that we can merge

rjnrohit: As of now , the name should be `Warp CI` in my opinion. I agree with @piwonskp that `yarn test` does a bunch of things beside running behv tests.

swapnilraj (approved):

cicr99 (approved):

Issue #825, opened 2023-01-15
All commands should have short descriptions


JorikSchellekens:
```
Usage: warp [options] [command]

Options:
  -h, --help                          display help for command

Commands:
  transpile [options] <files...>
  transform [options] <file>
  test [options]
  analyse [options] <file>
  status [options] <tx_hash>
  compile [options] <file>
  gen_interface [options] <file>
  deploy [options] <file>
  deploy_account [options]
  invoke [options] <file>
  call [options] <file>
  install [options]
  declare [options] <cairo_contract>  Command to declare Cairo contract on a StarkNet Network.
  version
```

swapnilraj:
closed in #827

PR #827, opened 2023-01-15
Add descriptions to all commands

JorikSchellekens:
Addresses #825


src/cli.ts
```diff
@@ -61,24 +61,31 @@ export const program = new Command();
...
-  .option('--no-result')
-  .option('--no-stubs')
-  .option('--no-strict')
+  .option(
+    '-d, --debug-info',
+    'Include debug information in the compiled bytecode produced by --compile-cario',
+    false,
+  )
+  .option('--print-trees', 'Debug: Print all the intermediate ASTs')
+  .option('--no-stubs', 'Debug: Hide the stubs in the intermedidate ASTs when using print-trees')
```
JorikSchellekens:
```suggestion
  .option('--no-stubs', 'Debug: Hide the stubs in the intermedidate ASTs when using --print-trees')
```

Nit, can you do this for the rest too?

src/cli.ts
```diff
@@ -61,24 +61,31 @@ export const program = new Command();
...
+  .option('--order <passOrder>', 'Use a custom set of transpilation passes')
   .option('-o, --output-dir <path>', 'Output directory for transpiled Cairo files.', 'warp_output')
-  .option('-d, --debug-info', 'Include debug information.', false)
-  .option('--print-trees')
-  .option('--no-result')
-  .option('--no-stubs')
-  .option('--no-strict')
+  .option(
+    '-d, --debug-info',
+    'Include debug information in the compiled bytecode produced by --compile-cario',
```
piwonskp: cairo*

src/cli.ts
```diff
@@ -135,19 +142,24 @@ export function runTranspile(files: string[], options: CliOptions) {
...
+  .option(
+    '--highlight <ids...>',
+    'Debug: highlight selected ids in the AST printed by --print-trees',
+  )
+  .option('--order <passOrder>', 'Use a custom set of transpilation passes')
+  .option('-o, --output-dir <path>', 'Output directory for transformed Solidity files')
+  .option('--print-trees', 'Debug: Print all the intermediate ASTs')
+  .option('--no-stubs', 'Debug: Hide the stubs in the intermedidate ASTs when using --print-trees')
+  .option('--no-strict', 'Debug: Allow silent failure of AST consistency checks')
+  .option('--until <pass>', 'Pass to transform to')
```
piwonskp: Imo "Stop processing at specified pass" would be more straightforward

src/cli.ts
```diff
@@ -200,7 +214,8 @@ export interface IOptionalNetwork {
 
 program
   .command('status <tx_hash>')
-  .option('--network <network>', 'Starknet network URL.', process.env.STARKNET_NETWORK)
+  .description('Get the satus of a transaction')
```
piwonskp: status*

src/cli.ts
```diff
@@ -226,14 +242,20 @@ export interface SolcInterfaceGenOptions {
 
 program
   .command('gen_interface <file>')
+  .description(
+    'Use native Cario conracts in your Soldity by creating a Solidity interface and a Cairo translation contract for the target Cairo contract',
```
piwonskp: contracts*

piwonskp: Cairo*

src/cli.ts
```diff
@@ -296,44 +320,46 @@ export type ICallOrInvokeProps = ICallOrInvokeProps_ &
 
 program
   .command('invoke <file>')
-  .requiredOption('--address <address>', 'Address of contract to invoke.')
-  .requiredOption('--function <function>', 'Function to invoke.')
+  .description('Invoke a function on a warped contract using the Solidity abi')
```
piwonskp, line 371: As there is an option  `--use-cairo-abi` it should be `Uses Solidity abi by default`

swapnilraj: I don't think we should be advertising the --use-cairo-abi to the user; its really for debugging purposes.
If somehow we can remove it from it a production build that would be ideal IMO

swapnilraj:
Hey @piwonskp thanks for your thorough review, really appreciate it.
For the future can we think of a way to run grammar check on ts files that produce strings that the user can see.
Let's make a task for it and mark it good first issue + help wanted

piwonskp (approved):

src/cli.ts
```diff
@@ -61,24 +61,31 @@ export const program = new Command();
...
+    false,
+  )
+  .option('--print-trees', 'Debug: Print all the intermediate ASTs')
+  .option('--no-stubs', 'Debug: Hide the stubs in the intermedidate ASTs when using --print-trees')
+  .option('--no-strict', 'Debug: Allow silent failure of AST consistency checks')
   .option('--until <pass>', 'Stops transpilation after the specified pass')
-  .option('--no-warnings')
-  .option('--include-paths <paths...>')
-  .option('--base-path <path>')
+  .option('--no-warnings', 'Suppress printed warnings')
```
rodrodros: These warnings refers to the warnings given by the solidity compiler and not the transpiler. I would update it to something like "Suppress solc warnings". Or change the code to suppress Warp warnings as well ...

src/cli.ts
```diff
@@ -135,19 +142,24 @@ export function runTranspile(files: string[], options: CliOptions) {
...
+  .description(
+    'Debug tool which applies any set of passes to the AST and writes out the transformed Solidity',
+  )
   .option('--no-compile-errors')
-  .option('--check-trees')
-  .option('--highlight <ids...>')
-  .option('--order <passOrder>')
-  .option('-o, --output-dir <path>')
-  .option('--print-trees')
-  .option('--no-result')
```
cicr99, line 145: What about this option?

swapnilraj: On the right is the new ones with the added descriptions

cicr99: Yes, I just meant that `--no-result` option was removed, no description was provided for it either, so I wanted to know if this was something that was intended

JorikSchellekens: I think this was a mistake in https://github.com/NethermindEth/warp/pull/827/commits/1e94ea68bcf66036e2907ed313817f3323da2b98

JorikSchellekens: Correction, this option didn't do anything anymore. It has been removed in https://github.com/NethermindEth/warp/commit/1e94ea68bcf66036e2907ed313817f3323da2b98

cicr99 (approved):

GregVardz (approved):

PR #853, opened 2023-01-17
Fix deploy with no constructor

### LINKED ISSUE #828: `warp deploy_account` doesn't seem to work anymore
```
warp deploy_account
Error: AssertionError:
starknet deploy failed
```

### LINKED ISSUE #829: Deploy with no constructor args doesn't work
```

pragma solidity ^0.8.14;

contract Mapping {
    // Mapping from address to uint
    mapping(address => uint) public myMap;

    function get(address _addr) public view returns (uint) {
        // Mapping always returns a value.
        // If the value was never set, it will return the default value.
        return myMap[_addr];
    }

    function set(address _addr, uint _i) public {
        // Update the value at this address
        myMap[_addr] = _i;
    }

    function remove(address _addr) public {
        // Reset the value to the default value.
        delete myMap[_addr];
    }
}

contract NestedMapping {
    // Nested mapping (mapping from address to another mapping)
    mapping(address => mapping(uint => bool)) public nested;

    function get(address _addr1, uint _i) public view returns (bool) {
        // You can get values from a nested mapping
        // even when it is not initialized
        return nested[_addr1][_i];
    }

    function set(address _addr1, uint _i, bool _boo) public {
        nested[_addr1][_i] = _boo;
    }

    function remove(address _addr1, uint _i) public {
        delete nested[_addr1][_i];
    }
}
```

### LINKED ISSUE #851: Cairo dependency upgrade
Update Warp's cairo depedency
Will require updating the CI files, warp_venv.sh and requirements.txt

swapnilraj:
Closes #829
Closes #854
Closes #828
Closes #851


piwonskp:
The "Closes" should be a comma-separated list

JorikSchellekens:
I don't think #854 is the correct issue

swapnilraj:
> The "Closes" should be a comma-separated list

thanks for correcting, Github I expected better

src/io.ts
```diff
@@ -89,7 +89,7 @@ export function outputResult(
     );
     fs.outputFileSync(fullCodeOutPath, code);
 
-    if (options.formatCairo || options.dev) {
+    if (options.formatCairo /* || options.dev : Cairo-format is disabled, as it has a bug */) {
```
JorikSchellekens: Should this not be disabled completely in the case that the tool itself has a bug?

piwonskp:
> thanks for correcting, Github I expected better

Actually github handles newline separated as well as comma separated. I've just checked that on another repo. It just doesn't work on warp repo / this PR. I'm gonna make some more tests

piwonskp:
Oh I see, you've made a PR to a side branch. Closing keywords work only for PRs to the default branch

rjnrohit (approved):

rjnrohit:
Don't you want to merge this PR to `develop` branch?

swapnilraj:
> Don't you want to merge this PR to `develop` branch?

No my PR uses stuff from cli_test so we can merge `cli_test` first and then Github will atuomatically change the target

.github/workflows/build.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         run: |
           yarn lint
           black starknet-testnet/ --check --verbose --diff --color
-          find warplib/ -iname *.cairo -exec cairo-format -c {} +
+        # find warplib/ -iname *.cairo -exec cairo-format -c {} +
```
piwonskp, line 43: Why would we keep that? (or why is it not needed anymore?)

piwonskp: Oh I see there's a comment below explaining that

tests/cli/utils.ts
```diff
@@ -1,3 +1,28 @@
+import { sh } from '../util';
+import * as path from 'path';
+
+export const wallet = 'starkware.starknet.wallets.open_zeppelin.OpenZeppelinAccount';
+export const gatewayURL = 'http://127.0.0.1:5050';
+export const network = 'alpha-goerli';
+export const starkNetAccountDir = path.resolve(__dirname, '.');
+export const warpBin = path.resolve(__dirname, '..', '..', 'bin', 'warp');
```
piwonskp: let's keep consts uppercase snake case

tests/cli/noArgsDeploy.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    const contract_address = extractFromStdout(stdout, CONTRACT_ADDRESS_REGEX);
+    const txHash = extractFromStdout(stdout, TX_HASH_REGEX);
+
+    expect(txFeeEth).to.not.be.undefined;
+    expect(txFeeWei).to.not.be.undefined;
+    expect(contract_address).to.not.be.undefined;
+    expect(txHash).to.not.be.undefined;
+
+    const res = await sh(
+      `${warpBin} status ${txHash} --network ${network} --gateway_url ${gatewayURL} --feeder_gateway_url ${gatewayURL}`,
```
piwonskp: Imo a function `runWarpWithStarknetDefaults` or a string `STARKNET_DEFAULT_OPTIONS` would improve the readability of shell commands

tests/cli/noArgsDeploy.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    );
+
+    expect(res.stderr).to.be.empty;
+    expect(res.stdout).to.not.be.empty;
```
piwonskp, line 101: Why don't we use error codes? They're less sensitive. Example: tests fail when ran with the debugger as it prints some information to the stderr

docker.env
```diff
@@ -2,6 +2,5 @@ STARKNET_NETWORK=alpha-goerli
 STARKNET_WALLET=starkware.starknet.wallets.open_zeppelin.OpenZeppelinAccount
 STARKNET_ACCOUNT_DIR=/usr/src/warp/.starknet_accounts_devnet
 
-# Doesn't work (Cairo 0.10.0, 0.10.1)
-# STARKNET_GATEWAY_URL=http://devnet:5050
-# STARKNET_FEEDER_GATEWAY_URL=http://devnet:5050
+STARKNET_GATEWAY_URL=http://devnet:5050
+STARKNET_FEEDER_GATEWAY_URL=http://devnet:5050
```
piwonskp, line 6: Are you sure it is fixed? I've created [the issue](https://github.com/starkware-libs/cairo-lang/issues/118) in cairo repository and it is not resolved

swapnilraj: Warp now has the feeder gateway url and the gateway url options and it picks them up from the environment and passes them explicitly to the starknet cli

tests/cli/noArgsDeploy.test.ts
swapnilraj: Good point, changed from stderr detection to exit code

tests/cli/noArgsDeploy.test.ts
swapnilraj: abstracted the network options and the account options so its a lot cleaner now

tests/cli/noArgsDeploy.test.ts
swapnilraj: exec doesn't give an exit code if the process didn't fail and I realised it doesn't make sense to check the stderr or exit code because if the process fails then sh will fail and the test will fail altogether

JorikSchellekens (approved):

src/io.ts
swapnilraj: Do you mean remove the flag completely? I think its okay to suspend its actions for now because Cairo 1.0 compiler will have its own flag and in the upgrade we can enable it again

swapnilraj: yes fixed

PR #857, opened 2023-01-18
Remove 251 bit (now it's 256) keccak hashing: Events

rjnrohit:
Previously, we were only taking 251 bits of  events' topic.

### DIFF
--- a/src/cairoUtilFuncGen/event.ts
+++ b/src/cairoUtilFuncGen/event.ts
@@ -191,7 +191,6 @@ export class EventFunction extends StringIndexedFuncGen {
     return [
       `   let (mem_encode: felt) = ${abiFunc}(${argName});`,
       `   let (keccak_hash256: Uint256) = warp_keccak(mem_encode);`,
-      // `   let (keccak_hash256: Uint256) = felt_to_uint256(keccak_hash);`,
       `   let (${arrayName}_len: felt) = fixed_bytes256_to_felt_dynamic_array_spl(${arrayName}_len, ${arrayName}, 0, keccak_hash256);`,
     ].join('\n');
   }
--- a/src/cairoUtilFuncGen/event.ts
+++ b/src/cairoUtilFuncGen/event.ts
@@ -18,11 +18,10 @@ import {
   safeGetNodeType,
   TypeConversionContext,
   typeNameFromTypeNode,
-  warpEventCanonicalSignaturehash256,
+  warpEvenSignatureHash256FromString,
 } from '../export';
 import { StringIndexedFuncGen } from './base';
 import { ABIEncoderVersion } from 'solc-typed-ast/dist/types/abi';
-import createKeccakHash from 'keccak';
 
 export const MASK_250 = BigInt('0x3ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff');
 export const BYTES_IN_FELT_PACKING = 31;
@@ -30,14 +29,6 @@ const BIG_ENDIAN = 1; // 0 for little endian, used for packing of bytes (31 byte
 
 const IMPLICITS =
   '{syscall_ptr: felt*, bitwise_ptr : BitwiseBuiltin*, range_check_ptr : felt, warp_memory : DictAccess*, keccak_ptr: felt*}';
-
-function signatureHash(funcSignature: string): string {
-  const funcSignatureHash = createKeccakHash('keccak256').update(funcSignature).digest('hex');
-
-  return `0x${(
-    BigInt(`0x${createKeccakHash('keccak256').update(funcSignatureHash).digest('hex')}`) & MASK_250
-  ).toString(16)}`;
-}
 
 /**
  * Generates a cairo function that emits an event through a cairo syscall.
@@ -121,11 +112,8 @@ export class EventFunction extends StringIndexedFuncGen {
 
     const cairoParams = params.map((p) => `${p.name} : ${p.type}`).join(', ');
 
-    const topic: { low: string; high: string } = warpEventCanonicalSignaturehash256(
-      node.name,
-      node.vParameters.vParameters.map((param) =>
-        param.canonicalSignatureType(ABIEncoderVersion.V2),
-      ),
+    const topic: { low: string; high: string } = warpEvenSignatureHash256FromString(
+      this.ast.inference.signature(node, ABIEncoderVersion.V2),
     );
 
     const code = [
--- a/src/cairoUtilFuncGen/event.ts
+++ b/src/cairoUtilFuncGen/event.ts
@@ -18,7 +18,7 @@ import {
   safeGetNodeType,
   TypeConversionContext,
   typeNameFromTypeNode,
-  warpEvenSignatureHash256FromString,
+  warpEventSignatureHash256FromString,
   EMIT_PREFIX,
 } from '../export';
 import { StringIndexedFuncGen } from './base';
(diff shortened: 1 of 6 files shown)


src/utils/event.ts
```diff
@@ -61,3 +61,27 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
...
+    if (typeof arg === 'string') return arg;
+    return `(${arg.map(getArgStringRepresentation).join(',')})`;
+  };
+
+  const funcSignature = `${eventName}(${argTypes.map(getArgStringRepresentation).join(',')})`;
+  const funcSignatureHash = createKeccakHash('keccak256').update(funcSignature).digest('hex');
+
+  const hash = BigInt(`0x${createKeccakHash('keccak256').update(funcSignatureHash).digest('hex')}`);
+
+  const hash_low_128 = hash & 0xffffffffffffffffffffffffffffffffn;
```
swapnilraj: can you use camel case because that's the code standard for our codebase

swapnilraj: There is utility function to split into low and high please use that

tests/behaviour/expectations/behaviour.ts
```diff
@@ -2224,7 +2228,15 @@ export const expectations = flatten(
                 [
                   {
                     data: ['69'],
-                    keys: [warpEventCanonicalSignaturehash('uintEvent', ['uint256'])],
+                    keys: [
+                      `0x${(
+                        BigInt(warpEventCanonicalSignaturehash256('uintEvent', ['uint256']).low) +
```
swapnilraj: write a utility function for this

src/utils/event.ts
```diff
@@ -61,3 +62,26 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
...
+): { low: string; high: string } {
+  const getArgStringRepresentation = (arg: argType): string => {
+    if (typeof arg === 'string') return arg;
+    return `(${arg.map(getArgStringRepresentation).join(',')})`;
+  };
+
+  const funcSignature = `${eventName}(${argTypes.map(getArgStringRepresentation).join(',')})`;
+  const funcSignatureHash = createKeccakHash('keccak256').update(funcSignature).digest('hex');
+
+  const hash = BigInt(`0x${createKeccakHash('keccak256').update(funcSignatureHash).digest('hex')}`);
```
swapnilraj: what is this hash, we want to be as close to ethereum's encoding.
just the hash of the function signature is enough

rjnrohit: It went unnoticed in the upper function also

tests/behaviour/expectations/utils.ts
```diff
@@ -88,3 +88,7 @@ export function toCairoInt256(val: number | bigint): [string, string] {
 }
 
 export const toCairoInt8 = (val: number | bigint) => BigInt.asUintN(8, BigInt(val)).toString();
+
+export function cairoUint256toBigIntHex(val: { low: string; high: string }): string {
```
piwonskp: Let's change the name to `cairoUint256toHex`. Bigint in the name suggests that the returned value is of BigInt type which is not true

src/utils/event.ts
```diff
@@ -57,7 +60,24 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
   const funcSignature = `${eventName}(${argTypes.map(getArgStringRepresentation).join(',')})`;
   const funcSignatureHash = createKeccakHash('keccak256').update(funcSignature).digest('hex');
 
-  return `0x${(
-    BigInt(`0x${createKeccakHash('keccak256').update(funcSignatureHash).digest('hex')}`) & MASK_250
-  ).toString(16)}`;
+  const splitHash: bigint[] = toUintOrFelt(BigInt(`0x${funcSignatureHash}`), 256);
```
piwonskp: You can use pattern matching, so you don't have to retrieve low and high by index

src/utils/event.ts
```diff
@@ -57,7 +60,24 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
...
+  };
+}
+
+// NOTE: argTypes must not contain `uint` , it should be `uint256` instead
+export function warpEvenSignatureHash256FromString(functionSignature: string): {
+  low: string;
+  high: string;
+} {
+  const funcSignatureHash = createKeccakHash('keccak256').update(functionSignature).digest('hex');
+  const splitHash: bigint[] = toUintOrFelt(BigInt(`0x${funcSignatureHash}`), 256);
```
piwonskp: Same as above

src/cairoUtilFuncGen/event.ts
```diff
@@ -188,13 +185,13 @@ export class EventFunction extends StringIndexedFuncGen {
...
-    this.requireImport('warplib.keccak', 'warp_keccak_felt');
+    this.requireImport('warplib.keccak', 'warp_keccak');
     this.requireImport('warplib.dynamic_arrays_util', 'fixed_bytes256_to_felt_dynamic_array_spl');
 
     return [
       `   let (mem_encode: felt) = ${abiFunc}(${argName});`,
-      `   let (keccak_hash: felt) = warp_keccak_felt(mem_encode);`,
-      `   let (keccak_hash256: Uint256) = felt_to_uint256(keccak_hash);`,
+      `   let (keccak_hash256: Uint256) = warp_keccak(mem_encode);`,
+      // `   let (keccak_hash256: Uint256) = felt_to_uint256(keccak_hash);`,
```
piwonskp: Why is it commented?

src/utils/event.ts
rjnrohit: > There is utility function to split into low and high please use that: Swapnil

I wanted to reuse the pre-defined function, so I used this function

src/cairoUtilFuncGen/event.ts
rjnrohit: This comment is redundant, lemme remove it

src/utils/event.ts
piwonskp: By pattern matching I meant the syntax `const [low, high] = foo()` so it would be more readable

src/utils/event.ts
```diff
@@ -57,7 +60,24 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
...
-  return `0x${(
-    BigInt(`0x${createKeccakHash('keccak256').update(funcSignatureHash).digest('hex')}`) & MASK_250
-  ).toString(16)}`;
+  const [low, high] = toUintOrFelt(BigInt(`0x${funcSignatureHash}`), 256);
+
+  return {
+    low: `0x${low.toString(16)}`,
+    high: `0x${high.toString(16)}`,
+  };
+}
```
piwonskp, line 62: Couldn't the function call `warpEvenSignatureHash256FromString`? seems that the logic since `const funcSignatureHash`  is the same

src/utils/event.ts
```diff
@@ -57,7 +60,24 @@ export function warpEventCanonicalSignaturehash(eventName: string, argTypes: arg
...
+
+// NOTE: argTypes must not contain `uint` , it should be `uint256` instead
+export function warpEvenSignatureHash256FromString(functionSignature: string): {
```
piwonskp: Typo. missing "t" in event

piwonskp (approved):

cicr99 (approved):
lgtm

src/utils/event.ts
```diff
@@ -48,16 +48,29 @@ export function decodeEventLog(eventsLog: EventItem[]): EventItem[] {
 export type argType = string | argType[];
 
 // NOTE: argTypes must not contain `uint` , it should be `uint256` instead
-export function warpEventCanonicalSignaturehash(eventName: string, argTypes: argType[]): string {
+export function warpEventCanonicalSignaturehash256(
```
swapnilraj, line 51: We need the event decode and event encode functions as well

swapnilraj (dismissed):

src/utils/event.ts
RohitRanjangit: will add in a separate PR

rjnrohit: will add in a separate PR

PR #868, opened 2023-01-23
Add typename to declaration in expressionSplitter

piwonskp:


src/passes/expressionSplitter.ts
```diff
@@ -35,6 +40,16 @@ function* expressionGenerator(prefix: string): Generator<string, string, unknown
   }
 }
 
+function getRootIdentifier(node: Identifier | IndexAccess): Identifier | IndexAccess {
```
AlejandroLabourdette: Wouldn't be better to have this function after the class definition to improve readability and keep the style of other passes?

piwonskp: Actually in this file functions are both below and above the class. So we will have to fix it in other passes as well probably. Creating a good first issue like that would be a good idea

AlejandroLabourdette (approved):

cicr99 (approved):
LGTM

PR #869, opened 2023-01-23
New Cairo Gen

### LINKED ISSUE #843: Cairo Generated Function Pruner
# Cairo Generated Function Pruner

In Warp the function pruner pass removes all internal functions definitions which are not called  by any external function (be it directly or indirectly), nonetheless if any of the pruned functions it’s the only one to call a generated function, the generated function will still remain, even if it has no use. This produces unnecessary code and can be the cause of some bugs like the generated function implicits not being imported.

As a solution you should initially make the FunctionPrunner pass work at Source Unit level and not Contract level. But as you will notice that eliminating generated function definitions doesn’t stop them from being generated. The reason is they work differently and once they are created the generated code is stored and printed later, without checking if the function definition still exists.

Here are two possible solutions:

1. An ugly patchy one where you check for the generated functions definition which were deleted and erase accordingly. But this can get ugly really fast with edge cases, like generated functions which call other generated functions, and it would be overly complicated to cover all cases
2. The other way needs extensive modification, but I think the pros outweighs the cons. It’s making all generated functions a special type of `CairoFunctionDefinition` like `CairoGeneratedFunctionDefinition` , where they have properties like code, and a list of other generated functions they call. The cons are that all function generators should be modified to work like this and the cairo writer will need some modifications. The pros are you don’t need to do anything else, the already existing FunctionPrunner pass will handle which function to prune and not and also getting more internal api consistency, like if I delete a function definiton I expect for it not to be generated (which is not what’s currently happening).

### LINKED ISSUE #855: Cairo util generated functions code is outside the nodes that define them.
Cairo functions generated under `src/cairoUtilFuncsGenerated` are just a stub, the code from those functions is in `generatedFunctions` map from `StringIndexedFuncGen` class. The Idea is to substitute those stubs for a function definition that have inside the Cairo code from those functions.

rodrodros:
Big update to `cairoUtilFuncGen` that make generated functions part of the AST and their code is stored inside this nodes

Fix #855 
Fix #843 

Before this PR:

Cairo generated functions raw string were stored in mappings outside of the solc-typed ast. Dummy functions definitions where created with empty bodies and function calls to this function definitions. When removing unnecessary function definitions the raw strings remained because there are outside of the AST. 

After this PR:

1. Raw strings functions are part of the AST, so when deleting one, generated raw code is deleted accordingly.

2. Imports are treated as function definitions as well, so they get pruned when no one is using it.

To accomplish (1) and (2) there are now two new Cairo-AST nodes: `CairoGeneratedFunctionDefintion` and `CairoImportFunctionDefinition`.

Also on this PR:

All raw string cairo functions signatures gets parsed instead of being written by hand.

Refactorization in all util function generations (some more than others) to make them all have all the same external interface.

Warplib function signatures (name and implicits) are parsed from the generated code.


src/ast/ast.ts
```diff
@@ -55,7 +57,6 @@ export class AST {
 
   context: ASTContext;
   // node requiring cairo import -> file to import from -> symbols to import
```
rjnrohit: can remove this comment!

src/ast/ast.ts
```diff
@@ -355,9 +325,9 @@ export class AST {
     replaceNode(oldNode, newNode, parent);
     this.context.unregister(oldNode);
     this.setContextRecursive(newNode);
-    if (copyImports) {
-      this.copyRegisteredImports(oldNode, newNode);
-    }
+    // if (copyImports) {
+    //   this.copyRegisteredImports(oldNode, newNode);
+    // }
```
piwonskp: Why is it commented?

rodrodros: Leftover

src/ast/cairoNodes/cairoBlockFunctionDefinition.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  FunctionKind,
+  FunctionStateMutability,
+  FunctionVisibility,
+  ParameterList,
+} from 'solc-typed-ast';
+import { FunctionStubKind } from './cairoFunctionDefinition';
+import { CairoRawStringFunctionDefinition } from './cairoRawStringFunctionDefinition';
+
+export class CairoFunctionStub extends CairoRawStringFunctionDefinition {
```
piwonskp: The name of the file do not match the only class within it

src/cairoUtilFuncGen/abi/abiDecode.ts
```diff
@@ -62,47 +64,65 @@ export class AbiDecode extends StringIndexedFuncGenWithAuxiliar {
...
       ),
-      ['bitwise_ptr', 'range_check_ptr', 'warp_memory'],
       this.ast,
-      sourceUnit ?? this.sourceUnit,
+      this.sourceUnit,
     );
 
-    return createCallToFunction(functionStub, [expressions[0]], this.ast);
+    this.generatedFunctionsDef.set(key, funcDef);
+    return funcDef;
```
piwonskp, line 96: Usually when you've got a `getOrCreate` its good to have a `get` and `create` methods separately too

rodrodros: Yes, I agree. Can we postpone that? It will mean changing all the function generators :cold_sweat:

src/ast/cairoNodes/cairoGeneratedFunctionDefinition.ts
```diff
@@ -0,0 +1,53 @@
...
+  FunctionStateMutability,
+  FunctionVisibility,
+  ParameterList,
+} from 'solc-typed-ast';
+import { FunctionStubKind } from './cairoFunctionDefinition';
+import { CairoRawStringFunctionDefinition } from './cairoRawStringFunctionDefinition';
+
+export class CairoGeneratedFunctionDefinition extends CairoRawStringFunctionDefinition {
+  /**
+   * List of functions defintions called by the generated function
```
cicr99: Typo _definitions_

src/ast/visitor.ts
```diff
@@ -169,6 +172,9 @@ export abstract class ASTVisitor<T> {
   visitCairoFunctionDefinition(node: CairoFunctionDefinition, ast: AST): T {
     return this.visitFunctionDefinition(node, ast);
   }
+  visitCairoGeneratedFunctionDefinition(node: CairoGeneratedFunctionDefinition, ast: AST): T {
```
cicr99, line 175: Is there no need to add a CairoImportFunctionDefinition visit function as well?

src/cairoUtilFuncGen/abi/abiDecode.ts
piwonskp: Sure

src/cairoUtilFuncGen/base.ts
```diff
@@ -38,51 +42,42 @@ export abstract class CairoUtilFuncGenBase {
...
-  protected checkForImport(type: TypeNode): void {
-    if (type instanceof IntType && type.nBits === 256) {
-      this.requireImport('starkware.cairo.common.uint256', 'Uint256');
-    }
-  }
+  // TODO: Erase this
+  // protected checkForImport(type: TypeNode): void {
+  //   if (type instanceof IntType && type.nBits === 256) {
+  //     this.requireImport('starkware.cairo.common.uint256', 'Uint256');
+  //   }
```
piwonskp: Leftovers?

src/cairoUtilFuncGen/base.ts
```diff
@@ -117,6 +112,7 @@ export function delegateBasedOnType<T>(
   valueFunc: (type: TypeNode) => T,
 ): T {
   if (type instanceof PointerType) {
+    assert(false);
```
piwonskp: Either remove the assertion or the whole if it's a dead code

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'deploy':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'emit_event':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'get_caller_address':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'get_contract_address':
+      return createFuncImport('syscall_ptr');
+    default:
+      assert(false, `Import ${name} from ${path} is not defined.`);
```
rjnrohit: remove redundant `assert`

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+  outputs?: ParameterInfo[],
+  options?: { acceptsRawDarray?: boolean; acceptsUnpackedStructArray?: boolean },
+) {
+  const sourceUnit = getContainingSourceUnit(nodeInSourceUnit);
+
+  const hasInputs = inputs !== undefined && inputs.length > 0;
+  const hasOutputs = outputs !== undefined && outputs.length > 0;
+
+  const existingImport = findExistingImport(name, sourceUnit);
+  // PATCH PATCH PATCH
```
rjnrohit: nitpick : comment

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (existingImport !== undefined) {
```
rjnrohit, line 38: If there is an existing import then we can directly return it, as there would be no conflicting import with same name.

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+    //if (
+    //  existingImport.vParameters.vParameters.length > 0 ||
+    //  existingImport.vReturnParameters.vParameters.length > 0
+    //)
+    //  return existingImport;
+    if (!hasInputs || !hasOutputs) return existingImport;
+
+    // sourceUnit.removeChild(existingImport);
+    return createImportFuncFuncDefinition(
```
rjnrohit: what if there is an existing struct definition

src/ast/cairoNodes/cairoGeneratedFunctionDefinition.ts
rjnrohit: nitpick: definitions

rjnrohit (commented):
first batch of reviews from my side!!

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -1,10 +1,11 @@
 import assert from 'assert';
-import { ASTWriter, ContractKind, SourceUnit, SrcDesc } from 'solc-typed-ast';
+import { ASTWriter, ContractKind, FunctionDefinition, SourceUnit, SrcDesc } from 'solc-typed-ast';
+import { CairoImportFunctionDefinition, FunctionStubKind } from '../../ast/cairoNodes';
+import { CairoGeneratedFunctionDefinition } from '../../ast/cairoNodes/cairoGeneratedFunctionDefinition';
```
cicr99: Can we make these two imports into one?

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -41,21 +42,60 @@ export class SourceUnitWriter extends CairoASTNodeWriter {
       writer.write(v),
     );
 
-    const functions = node.vFunctions.map((v) => writer.write(v));
+    const [importFunctions, generatedFunctions, functions] =
+      node.vFunctions.length === 0
+        ? [[], [], []]
```
cicr99: Why is this condition needed? I think that if `node.vFunctions` is empty, then it should return three empty arrays, as you are providing that as the initial value.

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -41,21 +42,60 @@ export class SourceUnitWriter extends CairoASTNodeWriter {
...
+    const writtenImportFuncs =
+      importFunctions.length === 0
+        ? ''
+        : getGroupedImports(
+            importFunctions
+              .sort((funcA, funcB) =>
+                `${funcA.path}.${funcA.name}`.localeCompare(`${funcB.path}.${funcB.name}`),
+              )
+              .filter((func, index, importFuncs) => func.name !== importFuncs[index - 1]?.name),
+          ).reduce((writtenImports, importFunc) => `${writtenImports}\n${importFunc}`);
```
cicr99: It's the same thing here, but you'll need to provide an initial value for the reduce method. It would look like this:
```suggestion
    const writtenImportFuncs = getGroupedImports(
          importFunctions
            .sort((funcA, funcB) =>
              `${funcA.path}.${funcA.name}`.localeCompare(`${funcB.path}.${funcB.name}`),
            )
            .filter((func, index, importFuncs) => func.name !== importFuncs[index - 1]?.name),
        ).reduce((writtenImports, importFunc) => `${writtenImports}\n${importFunc}`, '');
```

src/cairoWriter/writers/cairoImportFunctionDefinitionWriter.ts
```diff
@@ -0,0 +1,9 @@
+import { ASTWriter, SrcDesc } from 'solc-typed-ast';
+import { CairoImportFunctionDefinition } from '../../ast/cairoNodes';
+import { CairoASTNodeWriter } from '../base';
+
+export class CairoImportFunctionDefinitionWriter extends CairoASTNodeWriter {
+  writeInner(node: CairoImportFunctionDefinition, _writer: ASTWriter): SrcDesc {
+    return [`from ${node.path} import ${node.name}`];
```
cicr99, line 8: I think this is not used anymore. We could delete it (I mean, return an empty string) , but also there is not harm in leaving it as it is. Just wanted to make sure this is noted. Either way we should add a comment saying that it is not used.

src/utils/cairoParsing.ts
```diff
@@ -0,0 +1,93 @@
...
+    rawFunction.match(/func (?<name>\w+)\s*/);
+
+  assert(
+    funcSignature !== null && funcSignature.groups !== undefined,
+    `Invalid parsing of raw string function:\n${rawFunction}`,
+  );
+
+  const name = funcSignature.groups.name;
+  //try {
+  // console.log('parsing implicits of', name);
```
cicr99: remove commented try-catch

src/passes/annotateImplicits.ts
```diff
@@ -140,6 +139,8 @@ class ImplicitCollector extends ASTVisitor<Set<Implicits>> {
   }
 }
 
+// TODO: There is code repetition here with utils/cairoParsing.ts
+// Should be updated before merging
```
cicr99: This can be left to do in a different PR when addressing Cairo block nodes, right?

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+      throw new TranspileFailedError(`Import ${name} from ${path} is not defined.`);
+  }
+}
+
+function findExistingImport(name: string, node: SourceUnit) {
+  const found = node.vFunctions.filter(
+    (n) => n instanceof CairoImportFunctionDefinition && n.name === name,
+  );
+
+  // assert(found.length < 2, `More than 1 import functions where found with name: ${name}.`);
```
cicr99: Why is this assert commented?

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  if (found[0] !== undefined) {
+    assert(found[0] instanceof CairoImportFunctionDefinition);
+    return found[0];
+  }
+  return undefined;
```
cicr99: Why do you use that condition instead of asking for the length? The filtering should return only `CairoImportFunctionDefinition` nodes, so instead of making the assertion you could add it  in the filter method. It could look like the following:
```suggestion
function findExistingImport(name: string, node: SourceUnit) {
  const found = node.vFunctions.filter(
    (n) : n is CairoImportFunctionDefinition => n instanceof CairoImportFunctionDefinition && n.name === name,
  );

  // assert(found.length < 2, `More than 1 import functions where found with name: ${name}.`);

  return found.length !== 0 ? found[0] : undefined;
}
```

src/passes/cairoUtilImporter.ts
```diff
@@ -20,28 +33,85 @@ export class CairoUtilImporter extends ASTMapper {
...
+        type.definition.walkChildren((child) => this.commonVisit(child, ast));
+        this.dummySourceUnit =
+          this.dummySourceUnit === currentSourceUnit ? undefined : this.dummySourceUnit;
+
+        // type.definition.vMembers.forEach((decl) => {
+        //   const declType = safeGetNodeType(decl, ast.inference);
+        //   if (declType instanceof IntType && declType.nBits > 251) {
+        //     createImportFuncDefinition('starkware.cairo.common.uint256', 'Uint256', node, ast);
+        //   }
+        // });
```
piwonskp: Commented code

src/utils/cairoParsing.ts
piwonskp: leftovers

src/utils/cairoParsing.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Given several Cairo function represented in plain text extracts information from it
+ *  @param rawFunctions Multiple cairo functions in a single text
+ *  @returns A list of each function information
+ */
+export function parseMultipleRawCairoFunctions(rawFunctions: string): RawCairoFunctionInfo[] {
+  const functions = rawFunctions.matchAll(/func (\w+)\s*[{]?.*?[}]?\s*[(].*?[)]/gis);
+
+  return [...functions].map((func) => getRawCairoFunctionInfo(func[0]));
+}
```
piwonskp, line 19: What if there is a comment with `{`, `}`, `(`, or `)` within cairo stub? Should we consider such an edge case? If we want to support any code compatible with cairo syntax we have to use a parser instead of regex

src/cairoUtilFuncGen/base.ts
rodrodros: Using assert so I could get a trace when debugging. Leaving it as it was.

src/utils/importFuncGenerator.ts
rodrodros: Yes that would be the ideal, but sometimes we have imports which we call several times but the meaning we give to their result may be different. For example, the result of `wm_read_id` can be any memory type, and we need the compiler to have that info, so is required to "recreate" the node (at least for now)

src/utils/importFuncGenerator.ts
rodrodros: In case of a struct it will never be recreated because they are always returned in case of being created previously

src/ast/cairoNodes/cairoGeneratedFunctionDefinition.ts
rodrodros: Do you mean the comment? Change it to:
```
List of function definitions....
```

src/utils/cairoParsing.ts
rodrodros: If the `{}()` symbols are inside the function signature, then it will crash. If they are in the body definition, then they will be ignored (unless the comment includes a proper func signature like `func name{...}(...)`) So yes i definitely agree to the use of a parser because an adversarial user could break the regex easily. I would propose making a task for it once the Cairo1 transition is complete. (I think a parser, and cairo-ast in typescript can be helpful to the transpiler)

piwonskp: Sure, we should definitely make an issue for that

src/utils/functionGeneration.ts
```diff
@@ -72,10 +77,12 @@ interface CairoFunctionStubOptions {
   acceptsUnpackedStructArray?: boolean;
 }
 
+export type ParameterInfo = [string, TypeName] | [string, TypeName, DataLocation];
+
 export function createCairoFunctionStub(
```
cicr99: Is this function being used at all? If not, can we remove it?

src/utils/cairoParsing.ts
rodrodros: Here #949

src/passes/cairoUtilImporter.ts
```diff
@@ -20,28 +33,78 @@ export class CairoUtilImporter extends ASTMapper {
...
+        'Uint256',
+        this.dummySourceUnit ?? node,
+        ast,
+      );
+    }
+
+    //  Patch to struct inlining
+    if (type instanceof UserDefinedType && type.definition instanceof StructDefinition) {
+      const currentSourceUnit = node.getClosestParentByType(SourceUnit);
+      assert(currentSourceUnit !== undefined);
```
cicr99: Use instead `getContainingSourceUnit` method from `src/utils`

src/passes/cairoUtilImporter.ts
```diff
@@ -11,7 +21,10 @@ import { isExternallyVisible, primitiveTypeToCairo } from '../utils/utils';
   the warplib maths functions as they are added to the code, but for determining if
   Uint256 needs to be imported, it's easier to do it here
 */
+
 export class CairoUtilImporter extends ASTMapper {
+  private dummySourceUnit: SourceUnit | undefined;
```
cicr99, line 25: I didn't get very well what's the use of this variable in this pass, could you explain?

src/utils/implicits.ts
```diff
@@ -30,18 +32,20 @@ export const implicitTypes: { [key in Implicits]: string } = {
   syscall_ptr: 'felt*',
   warp_memory: 'DictAccess*',
   keccak_ptr: 'felt*',
+  dict_ptr: 'DictAccess*',
 };
 
+// TODO: look at this again by the end
```
cicr99: Do you need to do anything else here?

tests/testing.ts
```diff
@@ -55,7 +55,6 @@ const expectedResults = new Map<string, ResultType>(
     ['exampleContracts/boolOpSideEffects.sol', 'Success'],
     ['exampleContracts/bytesXAccess.sol', 'Success'],
     ['exampleContracts/c2c.sol', 'Success'],
-    // Uses conditionals explicitly
```
cicr99, line 58: Why where these comments removed?

src/ast/visitor.ts
rodrodros: `CairoImportFunctionDefintion` is used as a placeholder for external (outside of the sourceUnit) functions. Functions which don't use anything from the current source unit. So they get the default visitor, and since they have an empty body, nothing besides the signature gets read.

src/cairoWriter/writers/sourceUnitWriter.ts
rodrodros: You are right, I forgot to update it

src/utils/functionGeneration.ts
rodrodros: Yes, we don't need it. I was leaving it there in case we might, but best to delete now I think

src/passes/cairoUtilImporter.ts
rodrodros: It's related to free struct inlining. Free structs get __written__ as part of a contract even if they are not a part of the contract definition (ast-wise). 

The problem of a free struct not being a part of the contract definition means that any import associated with those struct does not get asociated to the current contract/source unit producing compilation errors after __writing__ the cairo file.

This patch checks for all "external" struct defintions and adds the proper imports to the current source unit.

It is not the cleanest solution, but it was the quickest given the circumstances.

src/utils/implicits.ts
rodrodros: Nay

tests/testing.ts
rodrodros: Initially comments where there to explain why certain compilation tests were failing. After the errors worked out, usually the comments  got deleted. I could restore these if you want.

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,137 @@
...
+    case STARKWARE_CAIRO_COMMON_UINT256 + 'uint256_sub':
+      return createFuncImport('range_check_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'deploy':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'emit_event':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'get_caller_address':
+      return createFuncImport('syscall_ptr');
+    case STARKWARE_STARKNET_COMMON_SYSCALLS + 'get_contract_address':
+      return createFuncImport('syscall_ptr');
```
piwonskp, line 118: An `if` clause would be more readable here than switch

cicr99: Honestly, I think using switch case here is better. Having to write the if condition all the time would make it more difficult to read

piwonskp: I mean `if` could group cases yielding the same result. And there are a lot of those. Alternatively a good option would be object syntax. It's just more concise. Imo it could also be divided into two blocks: struct imports and func imports to further improve readability

src/ast/cairoNodes/cairoGeneratedFunctionDefinition.ts
rjnrohit: yes

src/utils/importFuncGenerator.ts
rodrodros: I would agree in theory, but the actual code gets bigger, and  I see no improvements in readability. If you still think grouping through `if` is  better I will change it.

piwonskp: It's not that I insist on if. Let's try to find the best solution :) There are multiple solutions apart from if, we could use object syntax as I mentioned before, we could also try to deplete the `case` to one liners. What I wonder is if it was the best choice to concatenate path and name. Wouldn't it be simpler to write something like that?
```
if (path === STARKWARE_CAIRO_COMMON_UINT256) {
 assert(UINT256_FUNCTIONS.includes(name), importError(path, name));
 return createFuncImport('range_check_ptr');
}

Comment (piwonskp) [reply src/utils/importFuncGenerator.ts:118]: Not sure how it works internally but imo it wouldn't hurt to have either `createFuncImport('range_check_ptr')` as a module constant or a function `rangeCheckPtr = () => createFuncImport('range_check_ptr')`

File: src/cairoUtilFuncGen/abi/abiDecode.ts [hunk 29]
```diff
@@ -124,17 +144,16 @@ export class AbiDecode extends StringIndexedFuncGenWithAuxiliar {
       `}`,
     ].join('\n');
 
-    const cairoFunc = { name: funcName, code: code };
-    this.generatedFunctions.set(key, cairoFunc);
-    return cairoFunc.name;
+    const cairoFunc = { name: funcName, code: code, functionsCalled: functionsCalled };
+    return cairoFunc;
```
piwonskp: The assignment doesn't seem to provide any value here

piwonskp: The name of the variable does not give any more context so let's just return the value without assignment

src/cairoUtilFuncGen/abi/abiDecode.ts
```diff
@@ -390,44 +431,69 @@ export class AbiDecode extends StringIndexedFuncGenWithAuxiliar {
...
-      return [
-        `// Decoding member ${member.name}`,
-        `${decodingCode}`,
-        `${getMemLocCode}`,
-        `${writeMemLocCode}`,
-      ].join('\n');
-    });
+    const decodingInfo: [string, CairoFunctionDefinition[]][] = definition.vMembers.map(
+      (member, index) => {
+        const type = generalizeType(safeGetNodeType(member, this.ast.inference))[0];
```
piwonskp: Let's use destructuring assignment instead of accessing by index. `const [type] = `

src/cairoUtilFuncGen/abi/abiDecode.ts
```diff
@@ -390,44 +431,69 @@ export class AbiDecode extends StringIndexedFuncGenWithAuxiliar {
...
+      },
+    );
+
+    const [instructions, functionsCalled] = decodingInfo.reduce(
+      ([instructions, functionsCalled], [currentInstruction, currentFuncs]) => [
+        [...instructions, currentInstruction],
+        [...functionsCalled, ...currentFuncs],
+      ],
+      [new Array<string>(), new Array<CairoFunctionDefinition>()],
+    );
```
piwonskp: It could be replaced by:
```
instructions = decodinginfo.map(a => a[0]);
functionsCalled = decodinginfo.map(a => a[1]);
```
Right?

src/cairoUtilFuncGen/abi/abiEncode.ts
```diff
@@ -42,23 +43,29 @@ export class AbiEncode extends AbiBase {
...
+          'bytes_index',
+          'bytes_offset',
+          '0',
+          `param${index}`,
         );
-        return [params, encodings];
+
+        encodings.push(paramEncoding);
+
+        return [params, encodings, functionsCalled.concat(paramFunctionsCalled)];
```
piwonskp, line 62: We usually prefer destructuring over concat in the code afaik

src/cairoUtilFuncGen/abi/abiEncode.ts
```diff
@@ -438,34 +494,60 @@ export class AbiEncode extends AbiBase {
...
-        `let (elem${index}) = ${readFunc};`,
-        `${encoding}`,
-        `let mem_ptr = mem_ptr + ${elemWidth};`,
-      ].join('\n');
-    });
+    if (existing !== undefined) return existing;
+
+    const decodingInfo: [string, CairoFunctionDefinition[]][] = def.vMembers.map(
+      (member, index) => {
+        const type = generalizeType(safeGetNodeType(member, this.ast.inference))[0];
```
piwonskp, line 522: Destructuring

src/cairoUtilFuncGen/abi/abiEncode.ts
```diff
@@ -438,34 +494,60 @@ export class AbiEncode extends AbiBase {
...
+      },
+    );
+
+    const [instructions, functionsCalled] = decodingInfo.reduce(
+      ([instructions, functionsCalled], [currentInstruction, currentFuncs]) => [
+        [...instructions, currentInstruction],
+        [...functionsCalled, ...currentFuncs],
+      ],
+      [new Array<string>(), new Array<CairoFunctionDefinition>()],
+    );
```
piwonskp: Maps?

src/utils/importFuncGenerator.ts
piwonskp: I thought a while and let's leave it as it is. After reviewing some more files the import mechanism requires a more complete solution. Modifying it here wouldn't resolve the root cause.

src/cairoUtilFuncGen/abi/abiEncodePacked.ts
```diff
@@ -62,24 +75,29 @@ export class AbiEncodePacked extends AbiBase {
...
+    const cairoFuncInfo = {
+      name: funcName,
+      code: code,
+      functionsCalled: [...importedFuncs, ...functionsCalled],
+    };
 
-    const cairoFunc = { name: funcName, code: code };
-    this.generatedFunctions.set(key, cairoFunc);
-    return cairoFunc.name;
+    return cairoFuncInfo;
```
piwonskp: The assignment doesn't seem to provide any value

src/cairoUtilFuncGen/abi/abiDecode.ts
rodrodros: Yes, way more clean

src/cairoUtilFuncGen/abi/abiEncode.ts
rodrodros: I think there are other 100 lines similar to these where we just take the zero index, all cairo writer are going to get updated sooner rather than later due to the cairo1 transition, I would suggest delaying this type of change until then

src/cairoUtilFuncGen/abi/abiEncode.ts
rodrodros: Yes, style varies from writer to writer. What we try to do is follow a functional programming style, with a new variable each iteration. I think some writers don't do that at all like this one (like params and encodings with push)

rodrodros: Normalizing the style throught the whole writers can be a time consuming stuff. I would suggest delaying it, and probably a lot of it will change in the cairo1 transition

src/cairoUtilFuncGen/abi/abiEncode.ts
piwonskp: Yeah I've seen that there is a lot of the same code in other func generators (I mean using reduces instead of maps, accessing 0 index instead of destructuring etc. the issues were repeating throughout all func gens). I've stopped commenting those because of that. So what do you propose? Focus on functionality not the code style while reviewing the PR? I don't think we've got the time to change the architecture here too so I assume we should just validate the correctness of the code? Maybe point out the bigger issues only?

piwonskp: Because we're falling into the trap: fundamental changes are too big to fix in the PR and code style changes on the other hand are too numerous to fix. So we have to define what kind of issues we want to be fixed before merging the PR

rodrodros: I believe is best to focus on correctness and fix style issues as we update the generators one by one for cairo1

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -1,101 +1,114 @@
...
+      return exprType;
+      // TODO: Only string and bytes (with fixed bytes) are concatenable, why there are extra types here!!!
+      // if (
+      //   !isDynamicArray(exprType) &&
+      //   !(exprType instanceof IntType || exprType instanceof FixedBytesType)
+      // )
+      //   throw new TranspileFailedError(
+      //     `Unexpected type ${printTypeNode(exprType)} in ${printNode(expr)} to concatenate.` +
+      //       'Expected FixedBytes, IntType, ArrayType, BytesType, or StringType',
+      //   );
```
piwonskp: Commented code

src/cairoUtilFuncGen/utils/encodeToFelt.ts
```diff
@@ -258,50 +293,88 @@ export class EncodeAsFelt extends StringIndexedFuncGen {
...
+      if (isValueType(elemenT)) {
+        return cairoElementT.width === 2
+          ? [
+              `assert to_array[to_index] = ${element}.low;`,
+              `assert to_array[to_index + 1] = ${element}.high;`,
+              `let to_index = to_index + 2;`,
+            ]
+          : [`assert to_array[to_index] = ${element};`, `let to_index = to_index + 1;`];
+      }
+      return [`let (to_index) = ${auxFunc!.name}(to_index, to_array, ${element});`];
```
piwonskp: Usage of `!`

tests/testing.ts
cicr99: I thought those comments were there as a way to explain what was being tested with them either the test was supposed to fail or be successful. Anyway, it's true that they are not so helpful now that those tests are passing. Just wanted to check if there was any other reason to delete them.
You can leave it like that, no need to restore it.

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -41,21 +45,54 @@ export class SourceUnitWriter extends CairoASTNodeWriter {
...
+          ? [[funcDef, ...importFunctions], generatedFunctions, functions]
+          : funcDef instanceof CairoGeneratedFunctionDefinition
+          ? [importFunctions, [funcDef, ...generatedFunctions], functions]
+          : [importFunctions, generatedFunctions, [funcDef, ...functions]],
+      [
+        new Array<CairoImportFunctionDefinition>(),
+        new Array<CairoGeneratedFunctionDefinition>(),
+        new Array<FunctionDefinition>(),
+      ],
+    );
```
piwonskp, line 58: That's a `filter`

src/utils/functionGeneration.ts
```diff
@@ -133,6 +113,97 @@ export function createCairoFunctionStub(
   return funcDef;
 }
 
+export function createImportFuncFuncDefinition(
```
piwonskp: FuncFuncy funky one

piwonskp: It seems the problem is the node class naming. `cairoImport` node wouldn't cause the problem. We can create an issue for that if that's too much work for now

src/cairoWriter/writers/sourceUnitWriter.ts
rodrodros: Something like this is what you have in mind?
```ts
   const importFunctions = node.vFunctions.filter(
      (f) => f instanceof CairoImportFunctionDefinition,
    );
    const generatedFunctions = node.vFunctions.filter(
      (f) => f instanceof CairoGeneratedFunctionDefinition,
    );
    const functions = node.vFunctions.filter(
      (f) =>
        !(f instanceof CairoGeneratedFunctionDefinition) &&
        !(f instanceof CairoImportFunctionDefinition),
    );
```

src/utils/importFuncGenerator.ts
piwonskp: +1

src/cairoWriter/writers/sourceUnitWriter.ts
piwonskp: Yes

src/utils/importFuncGenerator.ts
```diff
@@ -0,0 +1,137 @@
...
+  nodeInSourceUnit: ASTNode,
+  ast: AST,
+  inputs?: ParameterInfo[],
+  outputs?: ParameterInfo[],
+  options?: { acceptsRawDarray?: boolean; acceptsUnpackedStructArray?: boolean },
+) {
+  const sourceUnit = getContainingSourceUnit(nodeInSourceUnit);
+
+  const hasInputs = inputs !== undefined && inputs.length > 0;
+  const hasOutputs = outputs !== undefined && outputs.length > 0;
```
piwonskp: The placement of hasIn, out is misleading, they're only used when there is an existing import

src/warplib/gatherWarplibImports.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Implicits } from '../utils/implicits';
+import { parseMultipleRawCairoFunctions } from '../utils/cairoParsing';
+
+export const warplibImportInfo = readAllCairoScripts('warplib', 'warplib').reduce(
+  (warplibMap, [pathToFile, importPath]) => {
+    const rawCairoCode = fs.readFileSync(pathToFile, { encoding: 'utf8' });
+
+    let fileMap: Map<string, Implicits[]>;
+    if (warplibMap.has(importPath)) {
+      fileMap = warplibMap.get(importPath)!;
```
piwonskp: Usage of `!`

src/warplib/gatherWarplibImports.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+    if (file.endsWith('.cairo')) {
+      return [
+        ...filesInfo,
+        [fullPathName, `${importPath}.${file.slice(0, file.length - '.cairo'.length)}`],
+      ];
+    }
+
+    return filesInfo;
+  }, new Array<[string, string]>());
```
piwonskp: That is so much easier with [glob](https://www.npmjs.com/package/glob). You could simply specify `/**/*.cairo`.

src/passes/annotateImplicits.ts
rodrodros: Surely it will change, nonetheless update it to avoid code repetition until the time comes

src/utils/importFuncGenerator.ts
rodrodros: Done

src/cairoWriter/writers/cairoImportFunctionDefinitionWriter.ts
rodrodros: :+1:

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -299,9 +249,38 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
+        ? 2
+        : CairoType.fromSol(type, this.ast, TypeConversionContext.Ref).width;
+      return [
+        [
+          `let (read_${index}) = ${readFunc.name}(${add('mem_loc', offset)}, ${uint256(
+            allocSize,
+          )});`,
+          `let (member${index})= ${memberGetterFunc.name}(read_${index});`,
+        ],
+        [this.requireImport('starkware.common.uint256', 'Uint256'), memberGetterFunc, readFunc],
```
cicr99: The path of this import, isn't `cairo` missing? As in `'starkware.cairo.common.uint256'`

src/cairoUtilFuncGen/calldata/implicitArrayConversion.ts
```diff
@@ -125,415 +87,488 @@ export class ImplicitArrayConversion extends StringIndexedFuncGen {
...
-          ].join('\n');
-        }
-      }
-    });
-    return instructions;
+    return {
+      name: funcName,
+      code: code,
+      functionsCalled: [
+        this.requireImport('starkware.cario.common.uint256', 'Uint256'),
```
cicr99: typo in `cairo` in path

piwonskp (approved):

cicr99 (approved):

PR #871, opened 2023-01-24
literalExpressionEvaluator binary operation with boolean changes

### LINKED ISSUE #838: literalExpressionEvaluator binary operations with booleans
In the literalExpressionEvaluator pass, a binary expression could be calculated at compile time, even when only one argument is a literal. In the case of a boolean literal with the `&&` operator, the following is done:

```tsx
case '&&': // false && x = false
	return notNullMember ? null : false;
```

It is also known that if `notNullMember` is `true`, the result of the operation is the other member, and so the binaryOperation expression can be replaced by the left or right expression correspondingly. This is something that could be added.

Same thing happens for `||` operator

Add tests for this behaviour

aniket2405:
Fix #838 

The result of the binaryOperation in the `literalExpressionEvaluator` pass expression can be replaced by the left or right expression if the `notNullMember` is `true`.


src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
   if (left === null || right === null) {
     // In some cases a binary expression could be calculated at
     // compile time, even when only one argument is a literal.
+
+    let rightExpression;
+    if (left === null) {
+      rightExpression = true;
+    } else {
+      rightExpression = false;
+    }
```
cicr99: You could rewrite this to only:
`const rightExpression = (left === null);`

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+            if (rightExpression) {
+              replaceNode(node, node.vRightExpression);
+            } else {
+              replaceNode(node, node.vLeftExpression);
+            }
+            return null;
+          } else {
+            return false;
+          }
+        //return notNullMember ? null : false;
```
cicr99: This comment can be removed

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
         case '||': // true || x = true
-          return notNullMember ? true : false;
+          if (notNullMember === false) {
```
cicr99: same thing here about the if condition

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+            if (rightExpression) {
+              replaceNode(node, node.vRightExpression);
+            } else {
+              replaceNode(node, node.vLeftExpression);
+            }
+            return false;
```
cicr99: This part should return `null` instead of `false`. This is probably a mistake that was left in the last [pr](https://github.com/NethermindEth/warp/pull/803) that was made related to this pass. Obviously this behavior is not being tested which is not good. If you have a simple contract like: 
```
// SPDX-License-Identifier: AGPL-3.0-only
pragma solidity ^0.8.14;

contract WARP {
  function f(bool a) public pure returns (bool) {
    return false || a;
  }
}
```
Calling function `f(true)` will return `false`, which is not correct. Please fix this and add a test case as well.

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+          } else {
+            return true;
+          }
+        // return notNullMember ? true : false;
```
cicr99: Remove comment

tests/behaviour/contracts/expressions/literalExpressionEvaluator.sol
```diff
@@ -0,0 +1,13 @@
...
+//SPDX-License-Identifier: MIT
+
+contract WARP {
+  function false_and() pure public returns (bool) {
+    return true && false;  // false 
+  }
+
+  function true_or() pure public returns (bool) {
+    return false || true;  // true 
+  }
```
cicr99: None of these functions are testing the behavior you just implemented. Notice that for your logic to be executed one of the members of the binary operation must return `null` in its evaluation, meaning that it was not possible to know its value during compilation time. In this case both members are boolean literals, so none of them is going to return `null` in that evaluation

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,15 +129,43 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+    }
+
     const notNullMember = left ?? right;
     if (notNullMember === null) {
       return null;
     } else if (typeof notNullMember === 'boolean') {
       switch (node.operator) {
         case '&&': // false && x = false
-          return notNullMember ? null : false;
+          if (notNullMember === true) {
```
cicr99: This is a bad programming practice, since you already know that `notNullMember` is a boolean value. Please change it to:
```suggestion
          if (notNullMember) {
```

cicr99 (changes requested):

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -128,16 +129,37 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+    } else if (typeof notNullMember === 'boolean') {
       switch (node.operator) {
         case '&&': // false && x = false
-          return notNullMember ? null : false;
+          if (notNullMember) {
+            if (rightExpression) {
+              replaceNode(node, node.vRightExpression);
+            } else {
+              replaceNode(node, node.vLeftExpression);
+            }
```
piwonskp: you can use ternary operator to make the code more concise

piwonskp:
Conflicts need to be resolved

swapnilraj:
delete the `package-lock.json` file its an npm artifact we use yarn for this project

tests/behaviour/contracts/expressions/literalExpressionEvaluator.sol
```diff
@@ -0,0 +1,17 @@
+pragma solidity ^0.8.10;
```
cicr99: Consider changing the name of the file to be more specific, for example "booleanOperations" or something like that, as this file doesn't check all of the cases implemented in `literalExpressionEvaluator` pass, but only the boolean related ones.

tests/behaviour/contracts/expressions/literalExpressionEvaluator.sol
```diff
@@ -0,0 +1,17 @@
...
+    return callNull() && false;  // false 
+  }
+
+  function true_or() pure public returns (bool) {
+    return callNull() || true;  // true 
+  }
+
+  function callNull() pure public returns (bool) {
+    return true && false;
+  }
```
cicr99: These functions are still not checking the behavior you implemented. Notice in the case of the `&&` operation that one of the members is `false`, so the entire expression is `false` (that's the logic that was already implemented in the pass); the behavior you added is for the case where that member is `true`, so the expression can be replaced from `callNull() && true` to just `callNull()`. Same thing about the `||` operation.

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable prettier/prettier */
```
swapnilraj: why disable prettier?

aniket2405: i did not disable prettier anytime. not sure why this was automatically generated.

tests/behaviour/expectations/behaviour.ts
```diff
@@ -2737,6 +2737,10 @@ export const expectations = flatten(
             Expect.Simple('bitwiseOr', [], ['22']),
             Expect.Simple('bitwiseXor', [], ['20']),
           ]),
+          File.Simple('booleanOperation', [
+            Expect.Simple('false_and', [], ['1']),
```
cicr99: I don't understand why the expected value here is `1`. If `callNull()` returns `false`, then `callNull() && true` should return `false` as well. If this test is passing there is probably something wrong with the code.

cicr99 (changes requested):

tests/behaviour/expectations/behaviour.ts
cicr99: The bug was because the boolean expression was being replaced by the incorrect member. Once that was fixed another errors arose due to ast inconsistencies.

aniket2405: Interesting. Why exactly was the boolean expression replaced by incorrect members?

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -23,10 +24,14 @@ import { RationalLiteral, stringToLiteralValue } from './rationalLiteral';
...
   &&, ||
-
   Hence all literal nodes are of type Literal | UnaryOperation | BinaryOperation
 */
 
+interface NodeEvaluationInfo {
+  evaluation: RationalLiteral | boolean | null;
+  node: ASTNode;
+}
+
```
rodrodros: This is just a nitpick but I've noticed through out the code that we prefer `type NodeEvaluateInfo = {...}` to `interface ...`.

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -265,9 +344,34 @@ function createLiteralFromType(typeString: string): RationalLiteral | null {
...
+function generateInfo(
+  value: RationalLiteral | boolean | null,
+  node: ASTNode,
+  ast: AST,
+): NodeEvaluationInfo {
+  if (value === null) return { evaluation: value, node: node };
+  else if (typeof value === 'boolean')
+    return { evaluation: value, node: createBoolLiteral(value, ast) };
+  else return { evaluation: value, node: createNumberLiteralNode(value, ast) };
```
rodrodros: Just for the sake of extra security, I would add a condition for this last else statement `typeof value === number` and throw an error in other case

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
cicr99: True

cicr99 (approved):

rodrodros (approved):

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -38,20 +44,12 @@ export class LiteralExpressionEvaluator extends ASTMapper {
...
-    } else {
-      const intValue = result.toInteger();
-      if (intValue === null) {
-        throw new TranspileFailedError('Attempted to make node for non-integral literal');
-      }
-
-      ast.replaceNode(node, createNumberLiteral(intValue, ast));
+    let result = createLiteralFromType(node, ast);
+    if (result.evaluation === null) {
+      result = evaluateLiteralExpression(node, ast);
```
piwonskp, line 49: Could be `let result = createLiteralFromType(node, ast) ?? evaluateLiteralExpression(node, ast)`

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -38,20 +44,12 @@ export class LiteralExpressionEvaluator extends ASTMapper {
...
-    } else if (typeof result === 'boolean') {
-      ast.replaceNode(node, createBoolLiteral(result, ast));
-    } else {
-      const intValue = result.toInteger();
-      if (intValue === null) {
-        throw new TranspileFailedError('Attempted to make node for non-integral literal');
-      }
-
-      ast.replaceNode(node, createNumberLiteral(intValue, ast));
+    let result = createLiteralFromType(node, ast);
```
piwonskp, line 47: It should be verified with infertype class - handling of types there is different. For example literal strings don't contain string value in typestring anymore afaik

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
```diff
@@ -179,83 +221,121 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+        result = generateInfo(left.evaluation !== right.evaluation, node, ast);
+        break;
       case '&&':
-        return left && right;
+        result = generateInfo(left.evaluation && right.evaluation, node, ast);
+        break;
       case '||':
-        return left || right;
+        result = generateInfo(left.evaluation || right.evaluation, node, ast);
+        break;
```
piwonskp, line 244: Consider using lodash. With lodash the switch would have been:
```
ops = {
 "==": eq,
 ...
}
result = generateInfo(ops[node.operator](left.evaluation, right.evaluation), node, ast)
''

File: src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts [hunk 18]
```diff
@@ -179,83 +221,121 @@ function evaluateBinaryLiteral(node: BinaryOperation): RationalLiteral | boolean
...
+        result = generateInfo(left.evaluation.bitwiseAnd(right.evaluation), node, ast);
+        break;
       case '|':
-        return left.bitwiseOr(right);
+        result = generateInfo(left.evaluation.bitwiseOr(right.evaluation), node, ast);
+        break;
       case '^':
-        return left.bitwiseXor(right);
+        result = generateInfo(left.evaluation.bitwiseXor(right.evaluation), node, ast);
+        break;
```
piwonskp, line 303: Imo it would be more readable:
```
opToFunction = {
 "|": left.evaluation.bitwiseOr,
  ...
}
result = generateInfo(opToFunction[node.operator](right.evaluation), node, ast)
```

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
piwonskp: I've made an [issue](https://github.com/NethermindEth/warp/issues/971) for that

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
cicr99: I don't think you can do that in this case, `result` is of type `NodeEvaluationInfo`, so even if the evaluation is `null`, `result` isn't

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
cicr99: Thanks, I had just used the implementation we had in this pass for this method, and made the necessary changes to comply to the new return type of the function; but I didn't notice this had to be modified for the solc-typed-ast update. Good catch.

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
piwonskp: Oh, you're right :)

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
piwonskp: It would probbably require `"|": left.evaluation.bitwiseOr.bind(left.evaluation)` or something though

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
piwonskp: We've got higher priorities right now, let's merge is as it is

src/passes/literalExpressionEvaluator/literalExpressionEvaluator.ts
piwonskp: We've got higher priorities right now, let's merge is as it is

piwonskp (approved):

Issue #863, opened 2023-01-20
Initialization of static arrays in storage


cicr99:
The following contract:
```
// SPDX-License-Identifier: AGPL-3.0-only
pragma solidity ^0.8.14;

contract WARP {
  uint256[100000000] d;
}
```
fails with error: `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`
This only happens when the size of the array is big enough, for smaller size values it works fine.

The issue probably has to do with line https://github.com/NethermindEth/warp/blob/develop/src/passes/storageAllocator.ts#L60 in `StorageAllocator` pass, since it tries to create a new typeNode for each element of the array

LuisLara-UH:
This issue lies on line https://github.com/NethermindEth/warp/blob/develop/src/utils/defaultValueNodes.ts#L147 in `VariableDeclarationInitializer` pass. When it substitutes the static array for a tuple, it creates an ExpressionList containing an expression for each element of the array, which overflows the Javascript heap.

I think the solution for this could be to increase the maximum size of Node heap. But even with this, with larger arrays it is going to overflow the heap too. So, the solution also should have some throwing error for larger arrays.

@cicr99 could you please assign this task to me?

cicr99:
The issue you are talking about is different from this one and it's being worked on already in [#821](https://github.com/NethermindEth/warp/issues/821). That one refers to the initialization of static arrays in memory and this one has to do with storage.

The thing with this contract is that `d` is a state variable so we handle it first on the `StorageAllocator` pass. I looked into the bug a bit further and the transpilation breaks at this [line](https://github.com/NethermindEth/warp/blob/242439ab73029cee4de63ac3898f8bee0cdb4fff/src/utils/cairoTypeSystem.ts#L74):
```
return new CairoTuple(Array(narrowedLength).fill(elementType));
```
This is called from the `StorageAllocator` pass in the line I had mentioned in the previous comment. Maybe the solution for this is to create a different type of node rather than using `CairoTuple`. I think @rodrigo-pino had an idea about it, so maybe he can comment some more on the matter

rodrodros:
The problem is that we treat solidity static arrays as cairo tuples. But static arrays are a subset of tuples, because though they can have `n` elements, each element must be of a single type `T`. The solution to this bug would be taking advantage of this characteristic of static arrays, where there is no need to initialize the whole `n` elements with type `T`.

[...]

PR #873, opened 2023-01-26
Fix storage static arrays bug

LuisLara-UH:
This PR is to fix the bug of Initialization of large-size static arrays, described on #863. 

@rodrigo-pino suggested me 3 possible approaches to this issue:
1. Update the `CairoTuple` type so that it groups the tuple items by types.
2. Create a new type related to `CairoTuple` for arrays that receives the type and the size.
3. Same as 2, but without the new type being related to `CairoTuple`.

For the approach (1) I found that for tuples having `n` elements and more than 1 type (`t >1 && t<=n`), you need to save the positions of the tuple corresponding to each single type. I didn't see any clean way to do this without having the list of types that overflow the Node heap.
While implementing approach (2) I found the new type didn't have to reuse any of the implementations from `CairoTuple`, so I choose approach (3) and created a new type `CairoStaticArray`.

### DIFF
--- a/src/utils/cairoTypeSystem.ts
+++ b/src/utils/cairoTypeSystem.ts
@@ -256,10 +256,10 @@ export class CairoStaticArray extends CairoType {
     return `[Tuple]${`(${this.type.fullStringRepresentation})`.repeat(this.size)}`;
   }
   toString(): string {
-    return `(${this.type.toString()}` + `${`,` + this.type.toString()}`.repeat(this.size - 1) + `)`;
+    return `(${this.type.toString()}` + `${`, ` + this.type.toString()}`.repeat(this.size - 1) + `)`;
   }
   get typeName(): string {
-    return `${this.type.typeName}` + `${`,` + this.type.typeName}`.repeat(this.size - 1);
+    return `${this.type.typeName}` + `${`x` + this.type.typeName}`.repeat(this.size - 1);
   }
   get width(): number {
     return this.type.width * this.size;
--- a/src/utils/cairoTypeSystem.ts
+++ b/src/utils/cairoTypeSystem.ts
@@ -71,7 +71,7 @@ export abstract class CairoType {
           tp.size,
           `Arrays of very large size (${tp.size.toString()}) are not supported`,
         );
-        return new CairoTuple(Array(narrowedLength).fill(elementType));
+        return new CairoStaticArray(elementType, narrowedLength);
       }
     } else if (tp instanceof BoolType) {
       return new CairoFelt();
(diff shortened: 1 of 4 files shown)


tests/behaviour/contracts/largeSizeArray/large_size_array.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity ^0.8.0;
+
+contract WARP {
+    uint256[100000000] public data;
```
rodrodros: Warp crashed with one billion, this is 100 millions (if I didn't miss any zeroes). It would be cool to have the biggest possible number here. The closest to 2**256, the best.

tests/behaviour/expectations/behaviour.ts
```diff
@@ -3435,6 +3435,12 @@ export const expectations = flatten(
             new File('derived', 'Derived', [], [Expect.Simple('f', [], ['36', '0', '24', '0'])]),
           ]),
         ]),
+        new Dir('largeSizeArray', [
+          File.Simple('large_size_array', [
+            Expect.Simple('getLength', [], ['100000000', '0']),
+            Expect.Simple('setValue', ['10000000', '0', '456', '0'], ['456', '0']),
```
rodrodros: Use utility function `toCairoUint256` to represent the Uint256 inputs. Located in `tests/behaviour/expectations/utils.ts`

src/utils/cairoTypeSystem.ts
```diff
@@ -248,6 +248,33 @@ export class CairoTuple extends CairoType {
...
+  }
+  get width(): number {
+    return this.type.width * this.size;
+  }
+  serialiseMembers(name: string): string[] {
+    const serializedMembers: string[] = [];
+    for (let index = 0; index < this.size; ++index) {
+      serializedMembers.push(...this.type.serialiseMembers(`${name}[${index}]`));
+    }
+    return serializedMembers;
```
rodrodros: nitpick: Use mapRange function.
Something like:
```ts
return mapRange(this.size, (n) => this.type.serializeMembers(`${name}[${n}]`);
```

AlejandroLabourdette (approved):

src/utils/cairoTypeSystem.ts
```diff
@@ -248,6 +248,29 @@ export class CairoTuple extends CairoType {
   }
 }
 
+export class CairoStaticArray extends CairoType {
+  constructor(public type: CairoType, public size: number) {
+    super();
+  }
+  get fullStringRepresentation(): string {
+    return `[Tuple]${`(${this.type.fullStringRepresentation})`.repeat(this.size)}`;
```
rodrodros: I think the keyword `Tuple` in `fullStringRepresentation` should be changed to something unique different than `CairoTuple`. Perhaps something like `StaticArray` or `SameTypeTuple`.

Do we still need `CairoTuple`?

rodrodros (approved):

src/utils/cairoTypeSystem.ts
rodrodros: Also make the string readable, a uint8[2] will be smth like "[Tuple]feltfelt" which may hurt readability. Since we know there is only one type, for the previous example you could print instead "[Tuple][2][felt]" meaning a static array of size  two where each element is a felt

cicr99 (approved):
LGTM

rodrodros (approved):

PR #880, opened 2023-02-02
Disable function overloading by default

### LINKED ISSUE #858: Stop warp mangling if no overloading
Add a new flag `--no-mangle` which makes sure that the function names are not mangled if the functions are not overloaded.
If the flag is passed and function names are overloaded then it should throw an error and ask the user to retry without the flag.
Make sure that this code is tested.

JorikSchellekens:
I've opted to disable the mangling of function names by default on all contracts. The user will have to explicitly turn the feature on. I've added a check in unsupported features which rejects overloaded functions and warns the user that doing so will change the abi.

fixes https://github.com/NethermindEth/warp/issues/858

### DIFF
--- a/src/ast/ast.ts
+++ b/src/ast/ast.ts
@@ -58,6 +58,7 @@ export class AST {
   imports: Map<ASTNode, Map<string, Set<string>>> = new Map();
   public inference: InferType;
 
+  public mangleFunctionNames = false;
   readonly tempId = -1;
 
   constructor(
@@ -387,6 +388,10 @@ export class AST {
       }
     });
     return node.id;
+  }
+
+  setMangleFunctionNames(mangleFunctionNames: boolean): void {
+    this.mangleFunctionNames = mangleFunctionNames;
   }
 }
 
--- a/src/ast/ast.ts
+++ b/src/ast/ast.ts
@@ -58,7 +58,6 @@ export class AST {
   imports: Map<ASTNode, Map<string, Set<string>>> = new Map();
   public inference: InferType;
 
-  public mangleFunctionNames = false;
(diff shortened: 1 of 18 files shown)


swapnilraj:
We will also need to modify the paths to call/invoke functions because they are working under the assumption that the function names are mangled.

JorikSchellekens:
This implementation isn't good enough. It needs the following improvements:

- [x] It should mangle internal functions by default since these won't change the abi
- [x] It should annotate the Cairo file that external functions have been mangled and then call and invoke should parse this

JorikSchellekens:
I've decided that Swapnil's suggestion to always overload only in the cases where it's required is actually a much cleaner way to do. I won't question you again @swapnilraj. This resolves both of the prior suggestions.

JorikSchellekens:
Next steps:

- [ ] check that the removal of name mangling doesn't introduce any clashes in the constructor from the variable instantiation code inserted into the constructor. 
- [ ] alter the import statements to take an optional third parameter to replace the imports with so that they don't clash with user variables `register("a.b.c", "d", "potato") -> import a.b.c.d as potato`.
- [ ] replace all imports with an appropriately mangled version
  - [ ] replace references to imports in the generated cairo !!! This is the most likely step to introduce several bugs by omission.
  - [ ] replace references to import in the CairoWriter (if any)
 - [ ] update the cli to mangle functions according to whether the contract's abi contains overloaded functions
 - [ ] update the test to rename functions according to whether the contract's abi contains overloaded functions
 - [ ] update hardhat-warp to do the same

JorikSchellekens:
Going to close this. It can be picked up later

PR #882, opened 2023-02-02
event APIs

rjnrohit:
Add event APIs `(encodeWarpEvent, decodeWarpEvent)` in WarpInterface


src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -93,6 +97,33 @@ export class RejectUnsupportedFeatures extends ASTMapper {
...
+      if (param.indexed) {
+        const paramType = generalizeType(safeGetNodeType(param, ast.inference))[0];
+        if (isValueType(paramType)) {
+          this.commonVisit(node, ast);
+          return;
+        }
+        if (
+          paramType instanceof ArrayType ||
+          paramType instanceof MappingType ||
+          (paramType instanceof PointerType && isReferenceType(paramType.to))
```
rodrodros, line 112: Why no using `isReferenceType` directly?

rjnrohit: string/bytes types are supported

src/transcode/interface.ts
```diff
@@ -0,0 +1,21 @@
...
+export class WarpInterface {
+  ethersInterface: Interface;
+  decodeEventLog(eventFragment: EventFragment | string, data: string, topics: string[]): Result {
+    return this.ethersInterface.decodeEventLog(eventFragment, data, topics);
+  }
+  encodeEventLog(
+    eventFragment: EventFragment | string,
+    values: any[],
+  ): { data: string; topics: Array<string> } {
+    return this.ethersInterface.encodeEventLog(eventFragment, values);
```
swapnilraj: This is a neat way, I think we can get an even better experience by:
- making WarpInterface inherit from Interface 
- putting the logic of `encodeWarpEvent` and `decodeWarpEvent` in `encodeEventLog` and `decodeEventLog`

By doing libraries programmed against ethers.js API can easily use our replacement without changing any API.
Nicely done

tests/behaviour/contracts/events/indexed.sol
```diff
@@ -25,9 +25,9 @@ contract WARP{
 
     function arrayString() public {
         string[] memory a = new string[](3);
-        a[0] = "a";
-        a[1] = "b";
-        a[2] = "c";
+        a[0] = "roh";
```
swapnilraj, line 28: lol

src/utils/event.ts
```diff
@@ -1,48 +1,81 @@
...
+export type EventItem = { data: string[]; keys: string[]; order?: number };
 
-export function decodeEventLog(eventsLog: EventItem[]): EventItem[] {
-  // flat number to hex string with rjust 62
-  const flatNumberToHexString = (num: number | bigint | string): string => {
+export function decodeWarpEvent(fragment: EventFragment | string, warpEvent: EventItem): Result {
+  warpEvent = join248bitChunks(warpEvent); // reverse 248 bit packing
+
+  // Remove leading 0x from each element and join them
+  const data = `0x${warpEvent.data.map((x) => x.slice(2)).join('')}`;
```
swapnilraj: We will need to do the same for keys right?

src/utils/event.ts
```diff
@@ -1,48 +1,81 @@
...
+  const result = new WarpInterface(fragment).decodeEventLog(fragment, data, warpEvent.keys);
+  return result;
+}
+
+export function encodeWarpEvent(fragment: EventFragment, values: argType[], order = 0): EventItem {
+  const { data, topics }: { data: string; topics: string[] } = new WarpInterface(
+    fragment,
+  ).encodeEventLog(fragment, values);
+
+  const topicFlatHex = '0x' + topics.map((x) => x.slice(2).padStart(64, '0')).join('');
```
swapnilraj: why add the `0x` if you're going to slice it off in the next line anyway

src/utils/event.ts
```diff
@@ -1,48 +1,81 @@
...
+  const topicItems248: string[] = splitInto248BitChunks(topicFlatHex.slice(2));
+  const dataItems248: string[] = splitInto248BitChunks(data.slice(2));
+
+  return { order, keys: topicItems248, data: dataItems248 };
+}
+
+export function splitInto248BitChunks(data: string): string[] {
+  if (data.startsWith('0x')) data = data.slice(2);
```
swapnilraj, line 7: we remove the `0x` before calling this function
so we can remove this if

src/transcode/interface.ts
rjnrohit: But the problem is we can't override the extended class function if types/numbers of arguments are different.

src/utils/event.ts
rjnrohit: No, `decodeEventLog` expects data as a string while topics as an array of strings

src/transcode/interface.ts
swapnilraj: decodeWarpEvent has the same signature and changing the signature of encodeWarpEvent shouldn't be too hard

rjnrohit: decodeWarpEvent has signature `(fragment: EventFragment | string, warpEvent: EventItem)` while decodeEventLog has signature `(eventFragment: EventFragment | string, data: BytesLike, topics?: ReadonlyArray<string>)`

rjnrohit: Problem is `data array` we receive is of form `string[]` while decodeEventLog expects data as `BytesLikes` , hence,  we can't completely put all logic of `warpDecodeEvent` inside `decodeEventLog` keeping the type of `data` same, unless if may have you different opinion that I'm not able figure it out

rjnrohit: what do you think about the following re-structuring: does it solve your purpose?
```ts
export class WarpInterface extends Interface {
  decodeWarpEvent(fragment: EventFragment | string, warpEvent: EventItem): Result {
    warpEvent = join248bitChunks(warpEvent); // reverse 248 bit packing

    // Remove leading 0x from each element and join them
    const data = `0x${warpEvent.data.map((x) => x.slice(2)).join('')}`;

    const result = super.decodeEventLog(fragment, data, warpEvent.keys);
    return result;
  }

  encodeWarpEvent(fragment: EventFragment, values: argType[], order = 0): EventItem {
    const { data, topics }: { data: string; topics: string[] } = super.encodeEventLog(
      fragment,
      values,
    );

    const topicFlatHex = '0x' + topics.map((x) => x.slice(2).padStart(64, '0')).join('');
    const topicItems248: string[] = splitInto248BitChunks(topicFlatHex);
    const dataItems248: string[] = splitInto248BitChunks(data);

    return { order, keys: topicItems248, data: dataItems248 };
  }
}

```

swapnilraj: Yea this looks good

swapnilraj (approved):

src/transcode/interface.ts
```diff
@@ -0,0 +1,28 @@
+import { EventFragment, Interface, Result } from 'ethers/lib/utils';
+import { argType, EventItem, join248bitChunks, splitInto248BitChunks } from '../utils/event';
+
+export class WarpInterface extends Interface {
+  decodeWarpEvent(fragment: EventFragment | string, warpEvent: EventItem): Result {
```
swapnilraj, line 5: There are no tests using `decodeWarpEvent`, which is what the hardhat integration and other tools will make most use of.
Can you add a few tests for it?

swapnilraj (changes requested):

src/transcode/interface.ts
swapnilraj: tests similar to test/encode.test.ts would be great

rjnrohit: on it, thanks

src/transcode/interface.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  encodeWarpEvent(fragment: EventFragment, values: argType[], order = 0): EventItem {
+    const { data, topics }: { data: string; topics: string[] } = super.encodeEventLog(
+      fragment,
+      values,
+    );
+
+    const topicFlatHex = '0x' + topics.map((x) => x.slice(2).padStart(64, '0')).join('');
+    const topicItems248: string[] = splitInto248BitChunks(topicFlatHex);
```
piwonskp: Imo would be more readable if you did `0x${topicFlatHex}` here instead of addition above

src/passes/rejectUnsupportedFeatures.ts
```diff
@@ -93,6 +97,33 @@ export class RejectUnsupportedFeatures extends ASTMapper {
     this.addUnsupported('User defined Errors are not supported', node);
   }
 
+  visitEventDefinition(node: EventDefinition, ast: AST): void {
+    node.vParameters.vParameters.forEach((param) => {
+      if (param.indexed) {
```
piwonskp: `filter` could be used instead of an if clause

piwonskp (approved):

swapnilraj (approved):

PR #893, opened 2023-02-03
Add utility to encode strings for tests

cicr99:
There was already a part of the code for the semantic tests that did the encoding of strings, so I just extracted the logic into a method and updated the behaviour tests.

I did this to check some semantic tests that were failing and realize there were other tests that were commented on the `semantic_whitelist` that were not failing anymore. So I updated that as well

### DIFF
--- a/tests/behaviour/expectations/utils.ts
+++ b/tests/behaviour/expectations/utils.ts
@@ -92,3 +92,9 @@ export const toCairoInt8 = (val: number | bigint) => BigInt.asUintN(8, BigInt(va
 export function cairoUint256toHex(val: { low: string; high: string }): string {
   return `0x${(BigInt(val.low) + (BigInt(val.high) << 128n)).toString(16)}`;
 }
+
+export function encodeString(value: string): string[] {
+  const valueEncoded: number[] = Buffer.from(value).toJSON().data;
+  const byteString: string[] = valueEncoded.map((val) => val.toString());
+  return [byteString.length.toString()].concat(byteString);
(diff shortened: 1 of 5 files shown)


rjnrohit (approved):

swapnilraj (approved):

tests/behaviour/expectations/utils.ts
```diff
@@ -92,3 +92,9 @@ export const toCairoInt8 = (val: number | bigint) => BigInt.asUintN(8, BigInt(va
 export function cairoUint256toHex(val: { low: string; high: string }): string {
   return `0x${(BigInt(val.low) + (BigInt(val.high) << 128n)).toString(16)}`;
 }
+
+export function encodeString(value: string): string[] {
+  const valueEncoded: number[] = Buffer.from(value).toJSON().data;
```
piwonskp: You can use entries or values instead of toJSON

tests/behaviour/expectations/utils.ts
```diff
@@ -92,3 +92,9 @@ export const toCairoInt8 = (val: number | bigint) => BigInt.asUintN(8, BigInt(va
...
+  const byteString: string[] = valueEncoded.map((val) => val.toString());
+  return [byteString.length.toString()].concat(byteString);
```
piwonskp: `[byteString.length, ...valueEncoded].map(a => a.toString())`

piwonskp (approved):
LGTM

PR #897, opened 2023-02-06
Print transformed result to stdout if no output dir is given 

### LINKED ISSUE #898: `warp transform` does not output results

rjnrohit:
closes #898 
`bin/warp transform` was either printing nothing or fails for custom provided output dir

### DIFF
--- a/src/io.ts
+++ b/src/io.ts
@@ -58,6 +58,7 @@ export function replaceSuffix(filePath: string, suffix: string): string {
 }
 
 export function outputResult(
+  contractName: string,
   outputPath: string,
   code: string,
   options: OutputOptions & TranspilationOptions,
@@ -75,7 +76,6 @@ export function outputResult(
     const fullCodeOutPath = path.join(options.outputDir, outputPath);
     const abiOutPath = fullCodeOutPath.slice(0, -'.cairo'.length).concat('_sol_abi.json');
 
-    const contractName = path.basename(outputPath).slice(0, -'.cairo'.length);
     const solFilePath = path.dirname(outputPath);
 
     outputFileSync(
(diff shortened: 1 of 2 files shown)


src/io.ts
```diff
@@ -82,12 +83,16 @@ export function outputResult(
       abiOutPath,
       JSON.stringify(ast.solidityABI.contracts[solFilePath][contractName]['abi'], null, 2),
     );
-    outputFileSync(fullCodeOutPath, code);
-
+    outputFileSync(
+      _transform ? replaceSuffix(fullCodeOutPath, '_warp.cairo') : fullCodeOutPath,
```
piwonskp: Can we just push fullCodeOutPath generation outside the function? So we won't need the condition and an extra argument

rjnrohit: No, we can't get true `contractName` at line 79, if we move the logic outside the function and without an extra argument.

piwonskp: But contractName is not used to construct fullCodeOutPath. Only OutputPath and options.outputDir are

rjnrohit: Yeah, but still we have to add an extra argument which enables different file output paths between `transpile` and `transform` commands.

piwonskp: I've pushed changes. It works on my side, please verify

rjnrohit: did you provide the optional output dir argument in `bin/warp transform` command?

rjnrohit: Because I'm getting the following error
Screenshot 2023-02-08 at 4 41 21 PM

piwonskp: I didn't test it for the `-o` option. Now it should be fixed

piwonskp (approved):

swapnilraj (approved):

PR #900, opened 2023-02-06
bytes32(string) fix

### LINKED ISSUE #892: `bytes32` in `BuiltinHandler` pass
The following contract warps to a invalid cairo contract:
```solidity
contract ClientReceipt {
    function Deposit(address _from, bytes32 _id, uint _value) public {
	}
    constructor() {
        Deposit(msg.sender, bytes32("abc"), 7);
    }
}
```

rjnrohit:
closes #892

### DIFF
--- a/src/passes/builtinHandler/explicitConversionToFunc.ts
+++ b/src/passes/builtinHandler/explicitConversionToFunc.ts
@@ -213,6 +213,8 @@ function literalToFixedBytes(arg: Expression, typeTo: FixedBytesType): Expressio
     }
   }
 
+  arg.value = BigInt('0x' + arg.hexValue).toString();
   arg.typeString = typeTo.pp();
+  if (arg.kind === LiteralKind.String) arg.kind = LiteralKind.HexString;
   return arg;
 }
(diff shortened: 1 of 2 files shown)


swapnilraj:
the tests pass so it looks good but could explain the bug and how your change fixes it?

rjnrohit:
> the tests pass so it looks good but could explain the bug and how your change fixes it?

This fixes the bug because previously `arg.kind` remains as a `string` all the way upto cairo writer, And in case of strings to fixedBytes conversion, the cairoWriter writes the expression as a single `felt` value using node.value despite node having typestring `uint256` . Since, all FixedBytes conversions are later converted into number/hexNumbers, the, arg.kind for string conversion shouldn't be stay as `string`
https://github.com/NethermindEth/warp/blob/a158b06d2ddb6a1e647894b3e21c43462c0ceb0b/src/cairoWriter/writers/literalWriter.ts#L8
```js
switch (node.kind) {
      case LiteralKind.Number:
        switch (primitiveTypeToCairo(node.typeString)) {
          case 'Uint256': {
            const [high, low] = divmod(BigInt(node.value), BigInt(Math.pow(2, 128)));
            return [`Uint256(low=${low}, high=${high})`];
          }
          case 'felt':
            return [node.value];
          default:
            throw new TranspileFailedError('Attempted to write unexpected cairo type');
        }
      case LiteralKind.Bool:
        return [node.value === 'true' ? '1' : '0'];
      case LiteralKind.String:
      case LiteralKind.UnicodeString: {
        if (
          node.value.length === node.hexValue.length / 2 &&
          node.value.length < 32 &&
          node.value.split('').every((v) => v.charCodeAt(0) < 127)
        ) {
          return [`'${node.value}'`];
        }
        return [`0x${node.hexValue}`];
      }
```

piwonskp (approved):

swapnilraj:
> 

Nice catch and good explanation! thanks

swapnilraj (approved):

PR #902, opened 2023-02-07
Add support for user-defined erros

### LINKED ISSUE #830: Support UserDefinedErrors
This is a little tricky since StarkNet doesn't support anything like this. Maybe we can replace this with a simple revert and emit a compiler warning stating that we've done so?

swapnilraj:
closes #830


JorikSchellekens:
@swapnilraj should we close this?

Issue #849, opened 2023-01-16
Handle literals


piwonskp:
Type of literal was previously determined by typestring. That means any type could be assigned to Literal, eg. literal could have `uint8` type. InferType class infers the type based on node properties and structure thus [have a specific set of type nodes that may be returned given literal node as an argument](https://github.com/ConsenSys/solc-typed-ast/blob/c44a58acf183a0c359dddf30255fe951b8e4656c/src/types/infer.ts#L1085). 

To conform to infertype class we could:
* Keep using typestring parser only for literals
* Convert literals to constants
* Make subclasses for literals (Uint8Literal etc.)
* Have a WarpLiteral class with a `kind` attribute

rodrodros:
I personally don't like the first option because it would mean the existence of thousands of line of code only to support the old way. @piwonskp do you have any prefered one? Do you have some insight on what would be the pros and cons of each one?

piwonskp:
Honestly I don't have a strong opinion on that. I agree that typestring parser is not an option long term. I'd see that rather as a temporary solution to fix other problems. Now if you ask me about my preferred solution. Philosophically speaking the best option would be to subclass solc-typed-ast's Literal. Why? Because the library do not provide the features we need. So an extension to its capabilities seems to be rational. Using constants for me does not seem like a real solution since we loose control over the output - we no longer can create plain literals in the output cairo code. So I consider using constants as a workaround which is a little bit hackish. It limits our capabilities. However on the other hand I don't think we care whether we produce literals or constants. So there might be more practical reasons, eg. simplification of code due to converting literals once to constants and not having to deal with both types of nodes afterwards. So the real question is whether converting to constants would simplify our codebase? Or maybe there is a reason why we did not do that before?

PR #903, opened 2023-02-07
InferType-changes

AlejandroLabourdette:
This PR is to solve all conflicts from the use of the new InferType class to get node's types. Infer Type and the current way to compute types using e parser over typeString node's property are not compatible and they can't exist together so this branch can only be merged once it's completed the transition.

**Remaining errors**
Before the transition to Cairo 1, there were 7 behavior tests failing, some of them were caused by hard modifications to the AST that leave it in an inconsistent state that the former parser approach couldn't identify. Those AST modifications have a meaning if we have in mind the big picture and the last state of the transpiration, but in an immediate step it seams like an error. An example of those is the simulation of an `or` operation with a substraction of booleans (assuming false is 0 and true 1). A subtraction of two booleans have no semantic meaning.

These errors do not exist's anymore with the transition to `Cairo 1` so was decided to go for the transition but with the new version we also lost all testing over the transpiled files. The final state of warp pass some of the behavior test but not all of them and until don't exist a test pool strong like the one for `Cairo 0` it's not advised to merge this changes. The implementation of new features can be stuck by errors related to Infer Type if we merge these changes and could be hard to trace them back to this as a cause.

**Known errors**
- typeOfLiterals returns a specific and narrower set of types, the logic over some of the assumption on the expected types must be changed. Handling literals is an open discussion #849   
- Can not Abi-encode mapping types. Warp create new function definitions along the transpiling process, It's common for inferType to compute the type of those func computing their signature. This is not always possible for the new functions.
- `safeGetNodeType` with inferType return a specialized type, some cases were implemented thinking in a generalized type


JorikSchellekens:
This pr is waiting for behaviour tests to be working on the Cairo 1.0 branch.

PR #905, opened 2023-02-08
improving static array initialization

### LINKED ISSUE #821: Improve initialisation of static arrays in memory
To initialise empty static arrays in memory we generate a function to write 0 or the default value for the type (think structs) at the right offsets in memory. For example

```
function test() external {
   uint256[100] d;
}
```

Generates a function with 100 parameters all taking a value of 0 and then it writes 0 at each index sequentially, for these cases we only need to pass one parameter and instead of writing 0 sequentially to each location we should generate a recursive function that loops through the indexes of the array to do this.

esdras-santos:
fixes #821


src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -69,6 +69,7 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
   tupleGen(node: TupleExpression): FunctionCall {
     const elements = node.vOriginalComponents.filter(notNull);
     assert(elements.length === node.vOriginalComponents.length);
+    let typeString = node.raw.typeDescriptions.typeString;
```
cicr99: If you run `yarn lint --fix` on the root directory of the project, it should fix most of the linting errors and point out others that you need to fix yourself. CI checks linting of the code so it has to be successful

cicr99 (changes requested):
There are some semantic and behaviour tests failing that need to be fixed. If you go to the details of the checks, you could see which ones. You can take a look at the readme [here](https://github.com/NethermindEth/warp#testing-for-contributors-stethoscope) to see how to run tests locally.
To run only behaviour tests you use `yarn test:behaviour`. You could also add a filter if you want to test a particular set of contracts, for example `FILTER=abiDecode yarn test:behaviour` will run all the behaviour tests that include `abiDecode` in their path.
It's similar if you want to run semantic tests: `FILTER=abiDecode yarn test:semantic`.

src/cairoUtilFuncGen/memory/arrayLiteral.ts
esdras-santos: Thank you very much @cicr99 , I'll follow your instructions and improve the logic in the code

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -108,41 +157,148 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
...
   }
 
-  private getOrCreate(type: TypeNode, size: number, dynamic: boolean): GeneratedFunctionInfo {
+  private getOrCreate(
+    type: TypeNode,
+    size: number,
+    dynamic: boolean,
+    isUserDefined: boolean,
+    isMatrix: boolean,
+    is3dMatrix: boolean,
```
AlejandroLabourdette, line 166: Hi Echo, I think this approach is not considering arrays of 3 or more dimentions(eg: `uint[a][b][c][d]` being `a`,`b`,`c` and `d` constants) at least is not working for me. Should it be working?

AlejandroLabourdette (commented):
Also, all commits on the PR are squashed into one single commit when PR is merged into `develop`. You will have the chance to change that squash commit message when you merge but by default it is the PR name. Feel free to name the commits inside the PR the way you like, perhaps a name that bring you more specific information of what they contain

src/cairoUtilFuncGen/memory/arrayLiteral.ts
esdras-santos: okay @AlejandroLabourdette , i'm a little busy with this issue https://github.com/NethermindEth/warp/issues/987 , when i finish I will update this PR to the Cairo-1.0 syntax and fix the problem that you pointed out. Thank you

piwonskp:
Hey @esdras-santos, we appreciate your contribution. We know how much effort you have put into the PR, but we have talked about it internally and we think you would spend a lot of time on refactoring the PR before it could be merged. Since we expect soon to drop cairo 0 support and make cairo 1 a default branch we think that it would be more productive and beneficial to warp if you worked on cairo 1 issues instead of addressing comments on the PR. Otherwise we might end up in a state when the PR is still in review, but the default branch is cairo 1 or it is merged just before the transition to cairo 1.  Given the circumstances and the fact that merging the PR will cause merge conflicts with cairo 1 branch we  think it's best to close the PR and for you to focus on cairo 1 issues

PR #908, opened 2023-02-09
Error message when starknet-class-hash command do not work as expected

AlejandroLabourdette:


src/utils/utils.ts
```diff
@@ -551,7 +551,15 @@ export function getSourceFromLocations(
 export function runStarkNetClassHash(filePath: string): string {
   const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', '..', 'warp_venv', 'bin')}:$PATH`;
 
-  const classHash = execSync(`${warpVenvPrefix} starknet-class-hash ${filePath}`).toString().trim();
+  let classHash: string;
+  try {
+    classHash = execSync(`${warpVenvPrefix} starknet-class-hash ${filePath}`).toString().trim();
```
swapnilraj: you can return from inside the return from inside the try block no need to make let declaration outside

src/utils/utils.ts
```diff
@@ -551,7 +551,15 @@ export function getSourceFromLocations(
...
+  } catch (error) {
+    throw new TranspileFailedError(
+      `starknet-class-hash command do not work as expected.\n Please make sure you are using the required version of starknet by executing 'pip install -r requirements.txt'`,
```
swapnilraj: `pip install -r requirement` is for us but for the user it would be to run `warp install`

swapnilraj: look at the warp install command to see how its used

swapnilraj:
can you also wrap other places where we invoke starknet cli

AlejandroLabourdette:
> can you also wrap other places where we invoke starknet cli

Other places are already wrapped, [eg](https://github.com/NethermindEth/warp/blob/fbc4acfccd6360c6689aaa9719761a198d392a70/src/starknetCli.ts#L313).
I'm not sure that those cases errors are always because of an old starknet version, so I didn't include the recommendation of `warp install` in the message

src/utils/utils.ts
AlejandroLabourdette: Wouldn't be better that exist a difference between the 2 errors thrown in that function? Otherwise the error from the conditional will be caught by the `catch`

swapnilraj: The error message from the condition on line 555 is not helpful.
We should extract the reason from the error message, this should be done in the catch.

Also there are two cases of the execSync failing, one where the command fails because the starknet-class-hash wasn't found in that we should recommend the user to use `warp install` the other case is where the `starknet-class-hash` command returns an error. This should be made visible to the user rather than just leaving it as `starknet-class-hash` failed

swapnilraj (changes requested):

swapnilraj (approved):

src/utils/utils.ts
```diff
@@ -541,11 +541,21 @@ export function getSourceFromLocations(
 export function runStarkNetClassHash(filePath: string): string {
   const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', '..', 'warp_venv', 'bin')}:$PATH`;
 
-  const classHash = execSync(`${warpVenvPrefix} starknet-class-hash ${filePath}`).toString().trim();
-  if (classHash === undefined) {
-    throw new Error(`starknet-class-hash failed`);
+  try {
+    return execSync(`${warpVenvPrefix} starknet-class-hash ${filePath}`).toString().trim();
+  } catch (e: any) {
```
piwonskp: If you remove explicit `any` I think there are no linting errors

src/utils/utils.ts
```diff
@@ -541,11 +541,21 @@ export function getSourceFromLocations(
...
+    const error = getPersonalizedStarknetError('starknet-class-hash', e.stderr.toString());
+    throw new TranspileFailedError(error);
   }
-  return classHash;
+}
+
+export function getPersonalizedStarknetError(starknetCommand: string, err: string) {
```
piwonskp: Personalized is a wrong word I think. Maybe `addInstallationHintIfMissingCmd`?

src/starknetCli.ts
```diff
@@ -344,15 +347,16 @@ export function runStarknetNewAccount(options: StarkNetNewAccountOptions) {
...
         encoding: 'utf8',
       },
     );
-  } catch {
-    logError('Starknet new account creation failed');
+    console.log(output);
+  } catch (e: any) {
+    const error = getPersonalizedStarknetError('starknet new_account', e.stderr.toString());
+    logError(error);
   }
```
piwonskp: We might want to build a wrapper around execSync, the pattern is repeating

piwonskp: I mean, setting encoding, logging, adding installation hints

src/utils/utils.ts
AlejandroLabourdette: by default is `unknown`, so you can't get any member access. When I tried to restrict to a more specific class it say that it's only two possible value: `any` and `undefined`

piwonskp: So maybe use `Error` interface instead

AlejandroLabourdette: It won't work,
> Catch clause variable type annotation must be 'any' or 'unknown' if specified.ts(1196)

https://stackoverflow.com/questions/69021040/why-catch-clause-variable-type-annotation-must-be-any

https://stackoverflow.com/questions/57484453/how-to-get-err-stderr-from-execsync

The closest approach imo will be to create an Error interface with the properties we want to make sure they exists (stderr in this case). Then inside the catch block ask if `e` is instance of that error interface, if not then is an error that we don't expect there as a call stack limit exceeded or something like that. I thought that would be improbable, so I decided to keep it simple, but I will include those cases right now

src/starknetCli.ts
AlejandroLabourdette: there are places with no catch block like this
https://github.com/NethermindEth/warp/blob/94e1912f09760147ba2a986355a43d69b54df646/src/solCompile.ts#L148
So we should be returning the `stdout` and `stderr` to apply some logic there if one is provided and the other one not. Also the options change, take this as an example
https://github.com/NethermindEth/warp/blob/94e1912f09760147ba2a986355a43d69b54df646/src/autoRunSemanticTests.ts#L190
so we should then specify which options and their values to pass.
I think that if we create a wrapper we need to specialize it so far that would be kind of the same effort as doing it directly

src/utils/utils.ts
piwonskp: I see, honestly I'm not that against `any`. If other solutions will prove to be cumbersome let's use eslint disable line just to silence the linter

src/starknetCli.ts
piwonskp: We don't have to specialize the wrapper. We can simply use them where it's handy and stick with execSync whenever we need more customization :)

AlejandroLabourdette: ok

src/starknetCli.ts
```diff
@@ -193,23 +180,18 @@ export function runStarknetDeployAccount(options: IDeployAccountProps) {
...
-  } catch {
-    logError('starknet deploy failed');
-  }
+  execSyncAndLog(
+    `${warpVenvPrefix} ${command} --wallet ${options.wallet} --network ${
+      options.network
+    } ${account} ${options.gateway_url ? `--gateway_url ${options.gateway_url}` : ''} ${
+      options.feeder_gateway_url ? `--feeder_gateway_url ${options.feeder_gateway_url}` : ''
+    } ${options.account_dir ? `--account_dir ${options.account_dir}` : ''} ${
+      options.max_fee ? `--max_fee ${options.max_fee}` : ''
```
piwonskp: There is a lot of it so let's make a function 
```optionalArg (name, value) => value ? `${name} ${value}`  : ''```

File: src/starknetCli.ts [hunk 7]
```diff
@@ -94,17 +90,12 @@ export function runStarknetStatus(tx_hash: string, option: IOptionalNetwork & IG
...
-      {
-        stdio: 'inherit',
-      },
-    );
-  } catch {
-    logError('starknet tx_status failed');
-  }
+  execSyncAndLog(
+    `${warpVenvPrefix} starknet tx_status --hash ${tx_hash} --network ${option.network} ${gatewayUrlOption} ${feederGatewayUrlOption}`.trim(),
+    command,
```
piwonskp, line 104: `command` is used just once. Please either use it in the line above or insert plain string in the current line

src/starknetCli.ts
AlejandroLabourdette: nice

piwonskp (approved):

PR #909, opened 2023-02-09
Include cairo-lang version in the version command output

JorikSchellekens:


piwonskp (approved):

src/cli.ts
```diff
@@ -535,5 +535,13 @@ program
   .action(() => {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const pjson = require('../package.json');
+    const pythonRequirements = fs
```
rjnrohit: why aren't we using `starknet --vesion`? instead we're looking for an entry in `requirements.txt`

JorikSchellekens: Good point. Will update

cicr99 (approved):

PR #914, opened 2023-02-10
Create WarpInferType

AlejandroLabourdette:
This PR is to extend the function `typeOf` from `InferType` to work also with warp defined Nodes


piwonskp:
It is already implemented in `safeGetNodeType`

AlejandroLabourdette:
> It is already implemented in `safeGetNodeType`

Oh, ok, I didn't see that you already add it recently. I'll remove this PR then

PR #916, opened 2023-02-13
Add external calls validation

### LINKED ISSUE #904: CRITICAL: Contract-to-contract calling must check bounds of return variables
We don't validate whether a Uint256 returned by an untrusted callee contract is within bounds.

piwonskp:
Fixes #904


rjnrohit:
can you add a behaviour test where this check should throw an error? https://github.com/NethermindEth/warp/blob/84b742aa6d7c8a43124ec7182fd9d9281039e5b0/tests/behaviour/behaviour.test.ts#L230-L234

rjnrohit (approved):

tests/behaviour/contracts/cross_contract_calls/invalid_output.sol
```diff
@@ -0,0 +1,28 @@
...
+
+interface I{
+    function invalid() external returns (uint8);
+}
+
+
+contract A{
+    /// warp-cairo
+    /// DECORATOR(external)
+    /// func invalid_bde9e40e() -> (number : felt) {
```
swapnilraj: we have a macro for current func, this will make it possible to not edit this test when we remove name mangling

swapnilraj: smart way to get around the solc type system!

rodrodros (approved):

PR #923, opened 2023-02-13
Generated function pruner

LuisLara-UH:
Added functionality in `FunctionPruner` to remove generated functions #843 . The only generated functions being removed here are those which where accessible by the contract's removed functions, and are unreachable after removing those contract's functions. Here I made 2 main changes:
1. Created the variable `reachableFunctionsFromRemovedFunctions` which has the functions reachable from the removed contract's functions. In the beginning I thought on removing all unreachable generated functions, but this was not possible because there are calls to those functions generated after `FunctionPruner` pass, so on this pass there are unreachable generated functions which are later reachable.
2. Changed the identifier of the functions in the dfs, from `id` to `name`. This is because in the ast I found there are some repeated `FunctionDefinition` nodes, with different ids but same function name. This was causing some functions being removed because the ids didn't match.

### DIFF
--- a/src/cairoUtilFuncGen/base.ts
+++ b/src/cairoUtilFuncGen/base.ts
@@ -71,6 +71,13 @@ export abstract class CairoUtilFuncGenBase {
 export class StringIndexedFuncGen extends CairoUtilFuncGenBase {
   protected generatedFunctions: Map<string, CairoFunction> = new Map();
 
+  removeGeneratedFunction(fname: string) {
+    const functionsToRemove = [...this.generatedFunctions.keys()].filter(
+      (k) => this.generatedFunctions.get(k)?.name === fname,
+    );
+    functionsToRemove.forEach((fun) => this.generatedFunctions.delete(fun));
+  }
+
   getGeneratedCode(): string {
     return [...this.generatedFunctions.values()].map((func) => func.code).join('\n\n');
   }
--- a/src/cairoUtilFuncGen/base.ts
+++ b/src/cairoUtilFuncGen/base.ts
@@ -72,10 +72,9 @@ export class StringIndexedFuncGen extends CairoUtilFuncGenBase {
   protected generatedFunctions: Map<string, CairoFunction> = new Map();
 
   removeGeneratedFunction(fname: string) {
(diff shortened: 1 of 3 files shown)


src/cairoUtilFuncGen/base.ts
```diff
@@ -71,6 +71,13 @@ export abstract class CairoUtilFuncGenBase {
 export class StringIndexedFuncGen extends CairoUtilFuncGenBase {
   protected generatedFunctions: Map<string, CairoFunction> = new Map();
 
+  removeGeneratedFunction(fname: string) {
+    const functionsToRemove = [...this.generatedFunctions.keys()].filter(
+      (k) => this.generatedFunctions.get(k)?.name === fname,
+    );
+    functionsToRemove.forEach((fun) => this.generatedFunctions.delete(fun));
```
piwonskp: `.map(this.generatedFunctions.delete, this)` would remove the unnecessary wrapping function

src/cairoUtilFuncGen/index.ts
```diff
@@ -248,6 +248,11 @@ export class CairoUtilFuncGen {
       })
       .join('\n\n');
   }
+  removeGeneratedFunction(name: string) {
+    this.getAllChildren()
+      .filter((c) => c instanceof StringIndexedFuncGen)
+      .forEach((c) => (c as StringIndexedFuncGen).removeGeneratedFunction(name));
```
piwonskp, line 254: Wouldn't `(c: StringIndexedFuncGen)` work without an explicit cast? :thinking:

LuisLara-UH: `(c: StringIndexedFuncGen)` throws invalid cast error. It seems I cannot use `CairoUtilFuncGenBase` object as a `StringIndexedFuncGen` object without the explicit cast.

rodrodros:
@LuisLara-UH  PR #869 solves this issue#905

LuisLara-UH:
> @LuisLara-UH PR #869 solves this issue#905

Ok, if this is already solved I think we could close this PR. Also, I think we should link PR #869 to this issue #843.

cicr99:
I agree. Thanks anyway @LuisLara-UH for your work.

PR #948, opened 2023-02-20
Make most of the I/O operations asynchronously

Hyodar:
Hi everyone! Well, starting off, I'd like to say I don't really expect you to necessarily be merging this considering the amount of changes here, the current priorities and how something like this requires quite some time from a number of pairs of eyes to make sure there's no problems.

Still, I think this is a quite relevant improvement that could be made when you have the time, so I'm adding this PR to bring the discussion and for reference if this is ever implemented.

The rationale is that if you consider before most I/O operations were done synchronously, now this idle CPU time doesn't really exist that much anymore - magic of Node.js - as while we're waiting for the system to mess with the filesystem we can go on with other stuff. Also, the compilation step, which was probably the most significant bottleneck and done file-by-file, is now done in parallel, taking advantage of multi-threading without really being multi-threaded.

In terms of results, the execution time of `yarn test:examples` in my machine is ~30% of what it was before :).

Curiously, the behaviour test execution time didn't change, probably because the starknet devnet is the bottleneck there.


rjnrohit:
> 

we don't use starknet-devnet as a our execution platform yet, but I was afraid that the execution time is going to increase on starknet-devnet, since you have shaped `warp` to perform asynchronous task quickly, is it possible to integrate starknet-devnet.

I think the bottleneck for behaviour tests on github actions is the logging we do. 

Huge thanks to you @Hyodar 

We'll review this PR as soon as possible and we'll make warp better!!

exampleContracts/exampleContracts.test.ts
```diff
@@ -4,7 +4,10 @@ import { runTests } from '../tests/testing';
 const TIME_LIMIT = 2 * 60 * 60 * 1000;
 
 // Transpiling the solidity files using the `bin/warp transpile` CLI command.
-describe('Transpiling example contracts:', function () {
+describe('Example contracts transpilation', function () {
```
rjnrohit, line 7: we are now getting rid of word `transpilation`, we can use `compilation` instead

exampleContracts/exampleContracts.test.ts
```diff
@@ -4,7 +4,10 @@ import { runTests } from '../tests/testing';
...
   this.timeout(TIME_LIMIT);
-  runTests(false, false);
+
+  it('should output in expected transpilation results', async function () {
```
rjnrohit, line 10: `compilation`

src/passes/tupleFixes/index.ts
```diff
@@ -13,6 +13,7 @@ export class TupleFixes extends ASTMapper {
   static map(ast: AST): AST {
     TupleFlattener.map(ast);
     TupleFiller.map(ast);
+
```
rjnrohit: nitpick: extra line

src/transcode/utils.ts
```diff
@@ -146,8 +146,8 @@ function validateParam(param: unknown) {
   throw new CLIError('Input invalid');
 }
 
-export function parseSolAbi(filePath: string): [] {
-  const abiString = readFileSync(filePath, 'utf-8');
+export async function parseSolAbi(filePath: string): Promise<any[]> {
```
rjnrohit, line 149: `Promise<any[]>` -> `Promise<SolABIFunction[]>`

src/utils/errors.ts
```diff
@@ -27,8 +27,8 @@ function getSourceCode(node: ASTNode | undefined): string {
   const sourceUnit = node.getClosestParentByType(SourceUnit);
   if (sourceUnit === undefined) return '';
   const filePath = sourceUnit.absolutePath;
-  if (fs.existsSync(filePath)) {
-    const content = fs.readFileSync(filePath, { encoding: 'utf-8' });
+  if (existsSync(filePath)) {
```
rjnrohit, line 31: can't we make it asynchronous also?

src/utils/errors.ts
```diff
@@ -50,24 +50,27 @@ export function getErrorMessage(
...
+    };
+  });
+
+  let errorId = 0;
+
+  return [
+    error(initialMessage),
+    sourceErrors.flatMap((source) => [
+      `File: ${source.path}:`,
+      ...source.errors.map((err) => `${error(`${++errorId}. ${err.message}`)}:\n\n${err.code}`),
```
rjnrohit, line 74: can avoid `errorId` variable , we can use the optional index arguments from `flatMap` fn

src/io.ts
```diff
@@ -43,46 +52,60 @@ export function findAllFiles(targetPath: string, recurse: boolean): string[] {
...
-      if (!targetInformation.isDirectory()) {
+    try {
+      await fs.access(options.outputDir);
+
+      if (!(await fs.lstat(options.outputDir)).isDirectory()) {
         throw new TranspileFailedError(
           `Cannot output to ${options.outputDir}. Output-dir must be a directory`,
         );
       }
+    } catch {
```
rjnrohit: can add a warning here

starknet-testnet/server.py
```diff
@@ -36,9 +36,12 @@ async def pingpost():
 @app.route("/deploy", methods=["POST"])
 async def deploy():
     data = request.get_json()
+
```
rjnrohit: nitpick: extra line

tests/cli/cli.test.ts
```diff
@@ -37,19 +37,22 @@ describe('Manage starknet account', function () {
   this.timeout(TIME_LIMIT);
 
   it('should create a starknet account', async () => {
-    if (fs.existsSync(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'))) {
-      fs.unlinkSync(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'));
+    try {
+      await fs.access(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'));
+      await fs.unlink(path.resolve(__dirname, 'starknet_open_zeppelin_accounts.json'));
+    } catch {
+      // file does not exist
```
rjnrohit: warning

rjnrohit (changes requested):
Everything looks very good to me

src/utils/errors.ts
Hyodar: Initially, I also made it asynchronous, but that made it necessary to make a number of other functions and some class methods also async. Since this, IIRC, is only used at error throwing, in my opinion the benefit is not significant considering all the additional trouble, that's why I kept it like this. Let me know what you think!

tests/cli/cli.test.ts
Hyodar: I don't think it's necessary in this case, wdyt? This is almost equivalent to before, since the idea is `fs.access` throws if it can't find the directory. Still - I do think an `exists` helper would be nice here, since an empty catch does not look that good and the current way may hide an error from `unlink` (we could still check the error, but a helper would be better IMO).

src/io.ts
Hyodar: I think the same comment I made in a similar case applies here, let me know what you think.

src/utils/errors.ts
Hyodar: Actually, not sure something similar can be avoided in this case since we're doing a flat map including all errors. The index we'll be getting in `flatMap` is the index from the source array, and each `source` can contain multiple errors.

exampleContracts/exampleContracts.test.ts
Hyodar: I can change it here, no problem, but maybe it would be better to do this in another PR so as to keep it specific, wdyt?

src/utils/errors.ts
rjnrohit: Ohh, sorry I didn't see line 73 arrow function

src/transcode/utils.ts
Hyodar: Good! The more specific the types the better.

Hyodar: It would also be good to improve typing in `compileSolFilesAndExtractContracts` and the `YulTransformer` constructor.

rjnrohit: sure, we will take care of that, if you have some  spare time then you can also do it, rn we're (warp team) busy on cairo1 related changes for warp

Hyodar: Maybe we should make another PR also fixing those? This PR is already big and quite broad.

starknet-testnet/server.py
Hyodar: Done in 965f9b5.

src/passes/tupleFixes/index.ts
Hyodar: Done in 965f9b5.

src/transcode/utils.ts
rjnrohit: Yeah, I meant a different PR

exampleContracts/exampleContracts.test.ts
rjnrohit: okay, yeah make sense

rjnrohit:
Hey can you pls resolve the merge conflict, we can take this in !!

Hyodar:
Sure, will do! I'm a bit busy today and probably tomorrow, but I'll see what I can do. Also, I want to make one more change before merging to we guarantee we are cleaning up any child processes if our process exits.

rodrodros:
@Hyodar  I reviewed and it looked good to me. Thank you for this big PR. If you can fix the conflicts, I think we are ready to merge

PR #952, opened 2023-02-22
updates on warplib for erc20

### LINKED ISSUE #919: Update warplib functions code
ERC20 uses:
```
from warplib.maths.external_input_check_address import warp_external_input_check_address
from warplib.maths.external_input_check_ints import warp_external_input_check_int256
from warplib.maths.add import warp_add256
from warplib.maths.ge import warp_ge256
from warplib.maths.sub_unsafe import warp_sub_unsafe256
```
Most of these functions can be removed from our warplib folder and use the ones provided by cairo1.0 instead.
The preferred solution is to remove these functions and modify the passes where they were called to call the ones from cairo instead. 
A temporarily solution is to use these functions as a wrapper of the cairo 1.0 ones.

AlejandroLabourdette:
update functions used from warplib in ERC20 contract
Fix #919


warplib/maths/ge.cairo
```diff
@@ -1,11 +1,18 @@
-from starkware.cairo.common.math_cmp import is_le_felt
-from starkware.cairo.common.uint256 import Uint256, uint256_lt
 
-func warp_ge{range_check_ptr}(lhs: felt, rhs: felt) -> (res: felt) {
-    return (is_le_felt(rhs, lhs),);
+fn warp_ge(lhs: felt, rhs: felt) -> felt {
+    if lhs >= rhs {
```
rjnrohit, line 3: we can use builtin defined function here!

warplib/maths/ge.cairo
```diff
@@ -1,11 +1,18 @@
...
+    else {
+        return 0;
+    }
 }
 
-func warp_ge256{range_check_ptr}(op1: Uint256, op2: Uint256) -> (result: felt) {
-    let (lt: felt) = uint256_lt(op1, op2);
-    return (1 - lt,);
+fn warp_ge256(op1: u256, op2: u256) -> felt {
+    if op1 >= op2 {
```
rjnrohit, line 12: same , builtin function can be used

rjnrohit (changes requested):

rjnrohit (approved):

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
```diff
@@ -5,31 +5,19 @@ const INDENT = ' '.repeat(4);
...
-          `func warp_external_input_check_int256{range_check_ptr}(x : Uint256){`,
-          `${INDENT}alloc_locals;`,
-          `${INDENT}let inRangeHigh : felt = is_le_felt(x.high, ${mask(128)});`,
-          `${INDENT}let inRangeLow : felt = is_le_felt(x.low, ${mask(128)});`,
-          `${INDENT}with_attr error_message("Error: value out-of-bounds. Values passed to high and low members of Uint256 must be less than 2**128."){`,
-          `${INDENT.repeat(2)}assert 1 = (inRangeHigh * inRangeLow);`,
-          `${INDENT}}`,
-          `${INDENT}return();`,
+          `fn warp_external_input_check_int256(x : u256) {`,
+          `${INDENT}return ();`,
```
rodrodros: This function has no use, no need to generate it

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
```diff
@@ -5,31 +5,19 @@ const INDENT = ' '.repeat(4);
...
         return [
-          `func warp_external_input_check_int${int_width}{range_check_ptr}(x : felt){`,
-          `${INDENT}let inRange : felt = is_le_felt(x, ${mask(int_width)});`,
-          `${INDENT}with_attr error_message("Error: value out-of-bounds. Value must be less than 2**${int_width}."){`,
-          `${INDENT.repeat(2)}assert 1 = inRange;`,
-          `${INDENT}}`,
-          `${INDENT}return ();`,
+          `fn warp_external_input_check_int${int_width}(x : felt){`,
+          `${INDENT}let max: felt = ${mask(int_width)};`,
+          `${INDENT}assert( x <= max, 'Error: value out-of-bounds.');`,
```
rodrodros: I am no longer sure we need to do external input check for anything. I think their added automatically by the cairo1

src/warplib/implementations/maths/add.ts
```diff
@@ -11,26 +11,22 @@ import {
...
   const functions = forAllWidths((width) => {
     if (width === 256) {
       return [
-        `func warp_add256{range_check_ptr}(lhs : Uint256, rhs : Uint256) -> (res : Uint256){`,
-        `    let (res : Uint256, carry : felt) = uint256_add(lhs, rhs);`,
-        `    assert carry = 0;`,
-        `    return (res,);`,
+        // Add is safe by default with u256
+        `fn warp_add256(lhs : u256, rhs : u256) -> u256 {`,
+        `    return lhs + rhs;`,
```
rodrodros: Since addition is already defined for u256, I don't think we need a functon for it.

src/warplib/implementations/maths/add.ts
```diff
@@ -11,26 +11,22 @@ import {
...
       return [
-        `func warp_add${width}{range_check_ptr}(lhs : felt, rhs : felt) -> (res : felt){`,
+        `fn warp_add${width}(lhs : felt, rhs : felt) -> felt{`,
         `    let res = lhs + rhs;`,
-        `    let inRange : felt = is_le_felt(res, ${mask(width)});`,
-        `    assert inRange = 1;`,
-        `    return (res,);`,
+        `    let max: felt = ${mask(width)};`,
+        `    assert (res <= max, 'Value out of bounds');`,
+        `    return res;`,
```
rodrodros: Same with this, I don't think we are going to be using felts anytime into the future, `uintX` will be treated as `uX` in the new Cairo1

src/warplib/implementations/maths/add.ts
```diff
@@ -51,9 +47,8 @@ export function add_unsafe(): WarplibFunctionInfo {
     functions: forAllWidths((width) => {
       if (width === 256) {
         return [
-          `func warp_add_unsafe256{range_check_ptr}(lhs : Uint256, rhs : Uint256) -> (res : Uint256){`,
-          `    let (res : Uint256, _) = uint256_add(lhs, rhs);`,
-          `    return (res,);`,
+          `fn warp_add_unsafe256(lhs : u256, rhs : u256) -> u256 {`,
+          `    return lhs + rhs;`,
```
rodrodros: This is not unsafe

src/warplib/implementations/maths/add.ts
```diff
@@ -11,26 +11,22 @@ import {
 
 export function add(): WarplibFunctionInfo {
   const fileName = 'add';
-  const imports = [
-    'from starkware.cairo.common.math_cmp import is_le_felt',
-    'from starkware.cairo.common.uint256 import Uint256, uint256_add',
-  ];
+  const imports: string[] = [];
```
rodrodros: We can delete this variable if it isn't doing anything

src/warplib/implementations/maths/sub.ts
```diff
@@ -16,39 +16,19 @@ import {
...
-          `        return (Uint256(low_safe, high),);`,
-          `    }else{`,
-          '        //the borrow was used',
-          `        let (high) = bitwise_and(${bound(128)} + lhs.high - rhs.high - 1, ${mask(
-            128,
-          )});`,
-          `        return (Uint256(low_safe, high),);`,
-          `    }`,
+          `fn warp_sub_unsafe256(lhs : u256, rhs : u256) -> u256 {`,
+          '    return lhs - rhs;',
```
rodrodros: This is not unsafe

src/warplib/implementations/maths/sub.ts
```diff
@@ -16,39 +16,19 @@ import {
...
       } else {
         return [
-          `func warp_sub_unsafe${width}{bitwise_ptr : BitwiseBuiltin*}(lhs : felt, rhs : felt) -> (`,
-          `        res : felt){`,
-          `    let res : felt = ${bound(width)} + lhs - rhs;`,
-          `    let (res) = bitwise_and(res, ${mask(width)});`,
-          `    return (res,);`,
+          // TODO: Check if default sub of felt thrown error on underflow
+          `fn warp_sub_unsafe${width}(lhs : felt, rhs : felt) -> felt {`,
+          `    return lhs - rhs;`,
```
rodrodros: About the `todo` felts I do think they have no safety included in their opperations, but investigation should follow. I think you should test `warp_sub_unsafe` against this new implementation ( ineach of the cairos: 1 and 0.10) because I am not sure both functions have the same behaviour.

warplib/maths/ge.cairo
```diff
@@ -1,11 +1,18 @@
...
+        return 1;
+    }
+    else {
+        return 0;
+    }
```
rodrodros, line 17: I don't see a use for these functions, why keep functionalizing it?

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
AlejandroLabourdette: It's just to keep compatibility with all the code that use this function. At the time this PR was created the Cairo func gen was not merged yet so exist a lot of Cairo code written by it that could not be updated to avoid future merge conflicts

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
AlejandroLabourdette: At this point, except for u256, all other int are stored in a felt, so we should keep a track to not exceed that maximum capacity from the bits limit. Wouldn't you agree that we should keep this input check until we map directly uintN from Solidity to uN from Cairo1?

AlejandroLabourdette: Same as first comment response, compatibility issues.

src/warplib/implementations/maths/add.ts
AlejandroLabourdette: Same as first comment response, compatibility issues.

src/warplib/implementations/maths/add.ts
AlejandroLabourdette: Do we already start working on that or are rushing to have erc20 transpilable?

warplib/maths/ge.cairo
AlejandroLabourdette: One of the imports from erc20 use it `from warplib.maths.ge import warp_ge256`. Check #919

src/warplib/implementations/maths/add.ts
AlejandroLabourdette: its necessary to be able to return that way, otherwise would be necessary create a variable for the returned element and both ways are kind of the same

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
rodrodros: I think the best solution is just not to generate it and then go to the parts of the project which requires it and erase them

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
rodrodros: Ok

src/warplib/implementations/maths/add.ts
rodrodros: Same as the first answer, could you try remove the functionalization altoghether. Nothing should break

src/warplib/implementations/maths/add.ts
rodrodros: Ok, let's leave it for now

src/warplib/implementations/maths/add.ts
rodrodros: I think that it is a declaration with no use, and provides no value until the last line of code. I think it's better to just return empty list as part of the import:
```ts
return {fileName, imports: [], functions};
```

src/warplib/implementations/external_input_checks/externalInputChecksInts.ts
AlejandroLabourdette: Ok

warplib/maths/ge.cairo
rodrodros: I would say for this last case, to avoid functionalization altogether as well, but since you'll probably would need to modify the binary writers as well, then I guess  we can postpone it

warplib/maths/external_input_check_address.cairo
```diff
@@ -1,12 +1,5 @@
...
-func warp_external_input_check_address{range_check_ptr}(x: felt) {
-    let inRange: felt = is_le_felt(
-        x, 0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
-    );
-    with_attr error_message("Error: value out-of-bounds. Value must be less than 2**251.") {
-        assert 1 = inRange;
-    }
-    return ();
+fn warp_external_input_check_address(x: felt) {
+    let max: felt = 0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
```
rodrodros: address max value is 2**251 - 256 as per [this](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/definitions/constants.py#L23).  You should update the hex to match that value. It would be nice as well to add a comment with the new hex value

AlejandroLabourdette: perfect

warplib/maths/external_input_check_address.cairo
```diff
@@ -1,12 +1,6 @@
...
-        x, 0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
-    );
-    with_attr error_message("Error: value out-of-bounds. Value must be less than 2**251.") {
-        assert 1 = inRange;
-    }
-    return ();
+fn warp_external_input_check_address(x: felt) {
+    // The L2 address upper bound is 2**256 - 256, in max is stored its hex representation
+    let max: felt = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00;
+    assert( x <= max, 'Error: value out-of-bounds');
```
rodrodros: Nit: I would put the error message to include the "address" word somewhere, to make the message more helpful

Also in the comment I would not use generic L2 but StarkNet.

Finally in the comment, upper bound is 2**251 -  256. You had a typo there

rodrodros (approved):

PR #953, opened 2023-02-23
update return parameters in func definitions cairo writer

### LINKED ISSUE #950: Writer for return parameters in function definitions
Currently, return parameters are handled in `ParameterListWriter`, but the syntax for them has changed. Now return parameters are not named and you either return a tuple (in which case, each type has to be specified) or a simple type. Since this is so different from the way input parameters are handled, the suggestion is to create another writer just for return parameters.

AlejandroLabourdette:
fix #950

### DIFF
--- a/src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
+++ b/src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
@@ -122,7 +122,13 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
 
   private getReturns(node: CairoFunctionDefinition, writer: ASTWriter): string {
     if (node.kind === FunctionKind.Constructor) return '';
-    return `-> (${writer.write(node.vReturnParameters)})`;
+    const returnStr = writer.write(node.vReturnParameters);
+    if (returnStr.includes(',')) {
+      // If a ',' is included is because are returned more than 1 element so it will be a tuple
+      return `-> (${returnStr})`;
+    }
+    // Cairo1 does not need to always return a tuple as former versions
+    return returnStr !== '' ? `-> ${returnStr}` : '';
   }
 
   private getImplicits(node: CairoFunctionDefinition): string {
--- a/src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
+++ b/src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
@@ -122,13 +122,13 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
 
   private getReturns(node: CairoFunctionDefinition, writer: ASTWriter): string {
     if (node.kind === FunctionKind.Constructor) return '';
+
     const returnStr = writer.write(node.vReturnParameters);
-    if (returnStr.includes(',')) {
-      // If a ',' is included is because are returned more than 1 element so it will be a tuple
-      return `-> (${returnStr})`;
-    }
+    const paramLen = node.vReturnParameters.vParameters.length;
     // Cairo1 does not need to always return a tuple as former versions
-    return returnStr !== '' ? `-> ${returnStr}` : '';
+    if (paramLen > 1) return `-> (${returnStr})`;
+    else if (paramLen === 1) return `-> ${returnStr}`;
+    else return ''; // No return specified so nothing to print
   }
 
   private getImplicits(node: CairoFunctionDefinition): string {
--- a/src/cairoWriter/writers/parameterListWriter.ts
+++ b/src/cairoWriter/writers/parameterListWriter.ts
@@ -1,13 +1,12 @@
 import {
   ASTWriter,
-  ContractDefinition,
   DataLocation,
   FunctionDefinition,
   InferType,
   ParameterList,
   SrcDesc,
 } from 'solc-typed-ast';
-import { CairoDynArray, CairoType, TypeConversionContext } from '../../utils/cairoTypeSystem';
+import { CairoType, TypeConversionContext } from '../../utils/cairoTypeSystem';
 import { safeGetNodeType } from '../../utils/nodeTypeProcessing';
 import { isExternallyVisible, isReturnParamList } from '../../utils/utils';
 import { CairoASTNodeWriter } from '../base';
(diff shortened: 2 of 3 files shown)


AlejandroLabourdette:
Solve #950

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
```diff
@@ -122,7 +122,13 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
 
   private getReturns(node: CairoFunctionDefinition, writer: ASTWriter): string {
     if (node.kind === FunctionKind.Constructor) return '';
-    return `-> (${writer.write(node.vReturnParameters)})`;
+    const returnStr = writer.write(node.vReturnParameters);
+    if (returnStr.includes(',')) {
```
rjnrohit: would be better if we can make use `node.vReturnParameters` length function

src/cairoWriter/writers/parameterListWriter.ts
```diff
@@ -30,14 +30,32 @@ export class ParameterListWriter extends CairoASTNodeWriter {
         this.ast,
         varTypeConversionContext,
       );
-      if (tp instanceof CairoDynArray && node.parent instanceof FunctionDefinition) {
-        return isExternallyVisible(node.parent) ||
-          node.getClosestParentByType(ContractDefinition)?.name.includes('@interface')
-          ? `${value.name}_len : ${tp.vLen.toString()}, ${value.name} : ${tp.vPtr.toString()}`
-          : `${value.name} : ${tp.toString()}`;
+      if (isReturnParamList(node)) {
```
rjnrohit: not necessarily need to define a new function for this, we'd just insert the check statement inside the `if`

rjnrohit (changes requested):

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
AlejandroLabourdette: the thing is that if the parameter is a dynamic array it will get split into 2: the len and the array, check that behavior (or any other that could be added in the future) on the function definition writer could lead to a big condition so its best to delegate that to the param list writer. The ideal solution imo would be that the param list writer returned a list of the string representation of each param, but its assumed the current returned structure in other logics, so this is the closest to that

src/cairoWriter/writers/parameterListWriter.ts
AlejandroLabourdette: It is not clear what the condition do per se, so I thought could be a good idea to wrap it in a more meaningful representation

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
rjnrohit: we're not sure though, that dynamic array returns would be split into two parameters , this examples : https://github.com/starkware-libs/cairo/blob/main/examples/fib_array.cairo tells us that  an array parameter can be returned directly

AlejandroLabourdette: Ye, but that is beyond the context of this PR, perhaps when we address the representation of arrays we could solve it. @rodrigo-pino is the way of representing arrays already updated in the changes that you were working on? If so, I see no problem in change it to use the len of `vReturnParameters`

rodrodros: From what I've seen from Cairo1, I think there will be no need to split dynamic arrays between it's elements and it's length, so I would suggest counting all arguments as one, no matter the type.

src/cairoWriter/writers/parameterListWriter.ts
```diff
@@ -30,13 +30,24 @@ export class ParameterListWriter extends CairoASTNodeWriter {
...
+        // In Cairo1 return parameters are not named
+        if (tp instanceof CairoDynArray && node.parent instanceof FunctionDefinition) {
```
rodrodros: I would like to add that the condition `node.parent instanceof FunctionDefintion` is used liked 3 or 4 times, I would suggest refactoring to use it at least once if possible

src/utils/utils.ts
```diff
@@ -325,6 +326,10 @@ export function isExternallyVisible(node: FunctionDefinition): boolean {
   );
 }
 
+export function isReturnParamList(node: ParameterList) {
+  return node.parent instanceof FunctionDefinition && node.parent.vReturnParameters === node;
+}
```
rodrodros: If this function is only used by the `parameterListWriter`  then I guess we could store it directly there.

Also if the refactoring (suggested by the comment above) is possible I would suggest removing the function altogether, and try to improve the readiblity of the original condition using variables, something like
```ts
if (node.parent instanceof FunctionDefinition) {
    const funcDef = node.parent;
    if  (funcDef.vReturnParameters === node) { ... }
   ....
}

```

AlejandroLabourdette: It's only used for now in there, but given that Cairo1 introduce that return parameters are not named I think is a condition that will be needed in other places

src/cairoWriter/writers/parameterListWriter.ts
AlejandroLabourdette: it's used 3 times, but gets called 2. Refactor the code to use only 1 imo will make the code less readable. But I will do a minor refactor of the condition isReturnParamList inside the return instruction that I didn't do before because I think oversaturate the instruction.

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
AlejandroLabourdette: Yes, the thing is that I was trying to not write thing with the writters that our logic not support yet, that's why I was asking if we already do that process of dynamic arrays. But ok, I will change it now anyway.

src/utils/utils.ts
rodrodros: Why would we need it in other places that we already didn't?

rodrodros:
Please lint this PR before merging

src/utils/utils.ts
AlejandroLabourdette: Because so far made no diference if the param list was from args or from return, perhaps it only affects the writing, but is still a general propose function.

rodrodros: Alright, scanned all instances where `vReturnParameters` was used, and right now the only instance where a paremeter list asserts itself as returnParameter is in the writer. Also we never visit a `parameterList` node in any pass (and so it never checks for its parent and so neither checks for its location inside the parent).

So I think it's safe to say the scope of your function is the `ParameterListWriter`

AlejandroLabourdette: ok

PR #954, opened 2023-02-23
Dependency injection of infer type in Cairo Writer

AlejandroLabourdette:
Some of the Cairo Writers create a new instance of InferType every time their functions are called, this pr change that to work with a reference passed when Writers were created.


rjnrohit (approved):
lgtm

rodrodros (approved):

PR #955, opened 2023-02-23
compilation tests for cairo1 example contracts

rjnrohit:


tests/testing.ts
```diff
@@ -275,6 +283,74 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+      execSync("curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh", {
+        stdio: 'inherit',
+      });
+      execSync('rustup override set stable && rustup update', { stdio: 'inherit' });
+    } catch (e) {
+      throw new Error(
+        'Failed to install rust, try manual installation of rust from https://www.rust-lang.org/tools/install',
+      );
+    }
+  }
```
rodrodros: I think it is not the best approach to install rust for the users automatically. There multiple ways of installing a package, and we shouldn't decide for the user

tests/testing.ts
```diff
@@ -275,6 +283,74 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+  try {
+    console.log('Cloning cairo1 repo...');
+    if (fs.existsSync(CAIRO1_DIR)) {
+      console.warn('!! cairo directory already exists, overwriting it...');
+      execSync(`rm -rf ${CAIRO1_DIR}`, { stdio: 'inherit' });
+    }
```
rodrodros: I don't think we should overwrite, instead I would suggest throwing an exception with why the process was stopped and how to solve it

tests/testing.ts
```diff
@@ -275,6 +283,74 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+    execSync('git clone https://github.com/starkware-libs/cairo', { stdio: 'inherit' });
+  } catch (e) {
+    throw new Error('Failed to clone cairo repo');
```
rodrodros: I would include the error message from `e` in here

cicr99 (approved):
lgtm

rodrodros (approved):

tests/testing.ts
```diff
@@ -275,6 +283,65 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+    throw e;
+  }
+  // follow cairo1 README setup instructions
+  try {
+    console.log('Installing cairo1 dependencies...');
+    execSync(`cargo test --manifest-path ${CAIRO1_CARGO_TOML}`, { stdio: 'inherit' });
+  } catch (e) {
+    throw new Error('Failed to compile cairo1 and its dependencies');
+  }
+}
```
piwonskp: Can't we use binaries from their github? https://github.com/starkware-libs/cairo/releases

rjnrohit: Binaries are specific to linux machines  with x86_64 architecture, not sure whether it works for other machines

piwonskp: True

tests/testing.ts
```diff
@@ -275,6 +283,65 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+  } catch (e) {
+    if (e instanceof Error) {
+      throw `Failed to clone the cairo1 repo: ${e.message}`;
+    }
+    throw e;
+  }
+  // follow cairo1 README setup instructions
+  try {
+    console.log('Installing cairo1 dependencies...');
+    execSync(`cargo test --manifest-path ${CAIRO1_CARGO_TOML}`, { stdio: 'inherit' });
```
piwonskp: We should checkout to a fixed version of repo first

piwonskp: Let's use the newest release

tests/testing.ts
```diff
@@ -252,8 +257,11 @@ const expectedResults = new Map<string, ResultType>(
 export function runTests(force: boolean, onlyResults: boolean, unsafe = false, exact = false) {
   const results = new Map<string, ResultType>();
   if (force) {
+    if (!cairo1setup()) doCairo1Setup();
     postTestCleanup();
-  } else if (!preTestChecks()) return;
+  } else {
+    if (!cairo1setup() || !preTestChecks()) return;
+  }
```
piwonskp, line 258: Let's check for the presence of binaries first

src/starknetCli.ts
```diff
@@ -27,6 +27,44 @@ interface CompileResult {
...
+  const cairoPathRoot = filePath.slice(0, -'.cairo'.length);
+  const sierraResultPath = `${cairoPathRoot}.sierra`;
+  const casmResultPath = `${cairoPathRoot}.casm`;
+
+  try {
+    console.log(`Running cairo1 compile with cargo`);
+    execSync(
+      `${warpVenvPrefix} cargo run --manifest-path ${cairoTomlPath} --bin cairo-compile -- ${filePath} ${sierraResultPath} --replace-ids`,
+      { stdio: 'inherit' },
+    );
```
piwonskp: We can use `CARGO_TARGET_DIR=../bin` to generate compiler binaries and run them directly afterwards. This way we'll be able to use binaries downloaded from github

tests/testing.ts
rjnrohit: cairo1 does check for presence of binaries, line -307

src/starknetCli.ts
RohitRanjangit: github binaries don't work in my system

tests/testing.ts
rjnrohit: Good point!

tests/testing.ts
piwonskp: Oh I see. As I think about that. Why don't we upload binaries to repo until they're available via package manager? It's no use to compile them over and over again. Even for mac ARM users one person can compile it so others won't have to

rjnrohit: but we need to upload binaries for all available different architectures and machines, which is pretty tough task , also we have to detect the system architecture during runtime and download required binaries related to specific architecture.

src/starknetCli.ts
rjnrohit: Github binaries are only specific to some machines

tests/testing.ts
piwonskp: I think we already do that for nethersolc and other test utilities from solidity, what is the difference? Imo it's maximum 4 different binaries (arm linux, x86 linux, arm mac, x86 mac). Anyway that's not a big deal. I assume in the long term it will be distributed via package manager so we will manage without binaries in repo

src/starknetCli.ts
piwonskp: Sorry for confusion. What I meant is calling binaries directly (not through cargo) thus allowing to run binaries without cloning repo and to bypass all that stuff. Binaries could be manually put where cargo would put them

tests/testing.ts
```diff
@@ -275,6 +283,65 @@ export function runTests(force: boolean, onlyResults: boolean, unsafe = false, e
...
+    !fs.existsSync(CAIRO1_DIR) ||
+    !fs.existsSync(CAIRO1_CARGO_TOML) ||
+    !fs.existsSync(CAIRO1_COMPILE_BIN) ||
+    !fs.existsSync(SIERRA_RUN_BIN)
+  ) {
+    console.log(
+      "Please make sure you've done the cairo1 setup from here: https://github.com/starkware-libs/cairo#readme",
+    );
+    return false;
+  }
```
piwonskp: If we use binaries as explained [here](https://github.com/NethermindEth/warp/pull/955/files#r1122381161) then the check should be put before `!isRustInstalled`. If binaries exist the function would return true regardless of rust installation

src/starknetCli.ts
```diff
@@ -27,6 +36,39 @@ interface CompileResult {
...
+}
+
+export function compileCairo1(filePath: string): CompileCairo1Result {
+  assert(filePath.endsWith('.cairo'), `Attempted to compile non-cairo file ${filePath} as cairo`);
+  const cairoPathRoot = filePath.slice(0, -'.cairo'.length);
+  const sierraResultPath = `${cairoPathRoot}.sierra`;
+  const casmResultPath = `${cairoPathRoot}.casm`;
+
+  try {
+    console.log(`Running cairo1 compile with cargo`);
```
piwonskp: Remove `with cargo` not to cause a confusion please

piwonskp (approved):

piwonskp:
Changed the base to `cairo-1.0`

PR #956, opened 2023-02-23
[Cairo 1] Add string hash

piwonskp:
I left `wm_string_hash` commented as I'm not sure if it's needed anymore

### DIFF
--- a/warplib/string_hash.cairo
+++ b/warplib/string_hash.cairo
@@ -8,15 +8,10 @@
 
 use array::Array;
 use array::ArrayTrait;
-use array::array_len;
-use integer::u32_to_felt;
-use integer::u32_try_from_felt;
-use debug::print_felt;
-use option::OptionTrait;
 
 
 fn hash_chain(arr: @Array::<felt>, pos: u32) -> felt {
-    let next = pos + u32_try_from_felt(1).unwrap();
+    let next = pos + 1_u32;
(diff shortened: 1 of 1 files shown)


warplib/string_hash.cairo
```diff
@@ -1,33 +1,40 @@
...
 }
 
+// func wm_string_hash{pedersen_ptr: HashBuiltin*, range_check_ptr, warp_memory: DictAccess*}(
+//     mem_loc: felt
+// ) -> felt {
+//     let (len, ptr) = wm_to_felt_array(mem_loc);
+
+//     let hashedValue = string_hash(len, ptr);
+//     return hashedValue;
+// }
```
rodrodros, line 33: Normal `string_hash` is used when the array is a normal cairo array, but `wm_string_hash` is used when it is scattered  in a dict access where it transforms it into a normal cairo array, and then applies `string_hash`. I wouldn't mind for it to stay commented a little longer in the code

rodrodros (approved):

warplib/string_hash.cairo
piwonskp: I see, then let's leave it for now. We'll see what to do with that later

cicr99 (approved):
LGTM

piwonskp:
Replaced by #994

PR #961, opened 2023-03-01
update return parameters in func definitions cairo writer (rebased)

AlejandroLabourdette:


src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
```diff
@@ -122,7 +122,13 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
 
   private getReturns(node: CairoFunctionDefinition, writer: ASTWriter): string {
     if (node.kind === FunctionKind.Constructor) return '';
-    return `-> (${writer.write(node.vReturnParameters)})`;
+
+    const returnStr = writer.write(node.vReturnParameters);
+    const paramLen = node.vReturnParameters.vParameters.length;
+    // Cairo1 does not need to always return a tuple as former versions
+    if (paramLen > 1) return `-> (${returnStr})`;
```
piwonskp, line 129: Cairo 1 does not need to return a tuple, then why parentheses around return parameters?

src/cairoWriter/writers/parameterListWriter.ts
```diff
@@ -29,13 +28,11 @@ export class ParameterListWriter extends CairoASTNodeWriter {
...
-          node.getClosestParentByType(ContractDefinition)?.name.includes('@interface')
-          ? `${value.name}_len : ${tp.vLen.toString()}, ${value.name} : ${tp.vPtr.toString()}`
-          : `${value.name} : ${tp.toString()}`;
-      }
-      return `${value.name} : ${tp}`;
+      const isReturnParamList =
+        node.parent instanceof FunctionDefinition && node.parent.vReturnParameters === node;
+      // TODO: In the position of the type is written the typeString of the var. Needs to be checked the transformation
+      // of that typestring into de Cairo 1 syntax for that type (Eg: dynamic arrays of some variable)
+      return isReturnParamList ? `${tp}` : `${value.name} : ${tp}`;
```
piwonskp: String interpolation for tp is unnecessary

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
AlejandroLabourdette: If you want to return more than one value then the tuple is needed, the writer for params return a string of the form 't1, t2, ..., tN' that's why we wrap it with parenthesis '(t1, ...)'.
Tuple is not needed if we want to return just 1 element

piwonskp: Oh, you're right, I thought it was `paramLen < 1`. I guess it should be done on AST level though, so a tuple node containing multiple typenames I guess. It might require more work though so let's merge the PR and we'll fix it later

piwonskp (approved):

rjnrohit (approved):

PR #963, opened 2023-03-02
Use boolean operations as a part of Cairo1 compiler

AlejandroLabourdette:
This PR is on hold because bools are being represented in memory as a felt (that could be generalized to all being represented as felts) so was needed some sort of new pass that intervene in those write/read from memory or storage to do a proper conversion
Take a look at this issue: https://github.com/NethermindEth/warp/issues/1012


rodrodros:
Good work! Only missing thing is you need to update the binary expression evaluator for boolean values as well. You should wait for [this](https://github.com/NethermindEth/warp/pull/871) to get merged

AlejandroLabourdette:
> Good work! Only missing thing is you need to update the binary expression evaluator for boolean values as well. You should wait for [this](https://github.com/NethermindEth/warp/pull/871) to get merged

Ok, thanks

src/passes/references/externalReturnReceiver.ts
```diff
@@ -58,9 +58,9 @@ export class ExternalReturnReceiver extends ASTMapper {
 }
 
 function addOutputValidation(decl: VariableDeclaration, ast: AST) {
-  const validationFunctionCall = ast
-    .getUtilFuncGen(decl)
-    .boundChecks.inputCheck.gen(decl, safeGetNodeType(decl, ast.inference));
+  const type = safeGetNodeType(decl, ast.inference);
+  if (!checkableType(type)) return;
+  const validationFunctionCall = ast.getUtilFuncGen(decl).boundChecks.inputCheck.gen(decl, type);
```
piwonskp: I think it's here by an accident, it's the code from #962 and would cause merge conflicts between develop and cairo-1.0

src/warplib/generateWarplib.ts
```diff
@@ -43,15 +39,6 @@ export const warplibFunctions: WarplibFunctionInfo[] = [
   shl(),
   shr(),
   shr_signed(),
-  // ge - handwritten
-  ge_signed(),
```
cicr99: Is it save to remove operations for signed integers as well? I thought those were not supported yet.

AlejandroLabourdette:
> Good work! Only missing thing is you need to update the binary expression evaluator for boolean values as well. You should wait for [this](https://github.com/NethermindEth/warp/pull/871) to get merged

Imo these is solved since @cicr99 and @aniket2405 's PR introduce the literal binary expression evaluator changes for booleans.

rodrodros:
@AlejandroLabourdette when defining constant values such as 
```solidity
bool x = true;
```
It gets transpiled as
```
let x = 1;
```
Which is a felt. This PR should use boolean for those values as well.

AlejandroLabourdette:
> @AlejandroLabourdette when defining constant values such as
> 
> ```solidity
> bool x = true;
> ```
> 
> It gets transpiled as
> 
> ```
> let x = 1;
> ```
> 
> Which is a felt. This PR should use boolean for those values as well.

It was a minor hard condition in literal writer
```
case LiteralKind.Bool:
        return [node.value === 'true' ? '1' : '0'];
```
It's fixed now

rodrodros (approved):

cicr99 (commented):
Lgtm. I would like to get this [PR](https://github.com/NethermindEth/warp/pull/1002) to go in first though, so we can run the tests here as well

src/cairoUtilFuncGen/storage/storageRead.ts
```diff
@@ -61,13 +67,30 @@ export class StorageReadGen extends StringIndexedFuncGen {
...
+      const equivalentStoredType = new CairoFelt();
+      // Given that only 1 variable was generated, then, in `pack` is stored its name
+      const [reads, pack] = serialiseReads(equivalentStoredType, readFelt, readId);
+      assert(reads.length === 1, 'Storage read for Cairo Bools should generate only 1 variable');
+      // It's assumed here that `1` -> True and `0` -> False. With simplification purposes it´s
+      // infered that if is not `1` then it must be false.
+      funcBody = endent`
+        ${reads.map((s) => `  ${s}`).join('\n')}
+        ${pack} == 1
+      `;
```
rodrodros: Since we are going to have multiple types in Cairo 1.0, it is good to prepare this code to be more extensible for future PRs:
1. You need to change `getOrCreate` to receive the node solidity’s type instead of the its Cairo equivalent.
2. The `instanceof` comparison should be against the solidity type and not the Cairo type.
3. The equivalent in Cairo needs to be obtained by this solidity type.

rodrodros: Finally you need to replicate this code for `storageWrite`, `memoryRead` and `memoryWrite`, so there is translations back and forth for boolean types in all of their forms

AlejandroLabourdette: This Cairo Type is computed from the Solidity Type already in the caller function, so it's actually obtained from it. I think your requests resume to point 1 and points 2-3 are embedded in that change. I'm not sure though how it will improve extensibility. This pattern of receiving the Cairo Type in the `getOrCreate` function is not unique to `storageWrite` or `storageRead` Eg:
https://github.com/NethermindEth/warp/blob/abb2b5508acf3e19a51344d7fef6d6d7a3be4aa3/src/cairoUtilFuncGen/memory/memoryMemberAccess.ts#L82
I look at other funcGen scripts and they do both approach. If that's what you want to solve (have a unified structure for all `getOrCreate` functions) I think it could be a separate Issue

AlejandroLabourdette: About extend it to `memoryRead` and `memoryWrite` scripts, they were not cairo1 compatible yet
https://github.com/NethermindEth/warp/blob/abb2b5508acf3e19a51344d7fef6d6d7a3be4aa3/src/cairoUtilFuncGen/memory/memoryRead.ts#L105
https://github.com/NethermindEth/warp/blob/abb2b5508acf3e19a51344d7fef6d6d7a3be4aa3/src/cairoUtilFuncGen/memory/memoryWrite.ts#L78
If they will be required then I think is a good Idea to put this PR on hold until we address the rewriting of those scripts

rodrodros: >This Cairo Type is computed from the Solidity Type already in the caller function, so it's...

When we are dealing in the memory space, we are dealing with felts: uXs, booleans, addresses etc.. are stored as felts. When you read from those values we are no longer expecting a `felt` but a uXs, boolean or address. So the function that both read and write to the memory space should serve as translation layers to handle this logic.

You wrote a logic for bool but it is not extensible because you create a hardcoded cairo felt to generate the writings of such bool. (You could directly generate the writings for the `bool`)

Also, I am now noticing that both branches would have the same result no matter the difference 🤔 . I think there is no need to do `if is a bool` condtion. The read and write functions will work flawlessly without defining any special case for any cairo type

AlejandroLabourdette: >Also, I am now noticing that both branches would have the same result no matter the difference 🤔...

Both, `storageRead` and `storageWrite` create a distinction for the function definition generated taking a look at the type of the value that must be read/written. If the value is a bool then it will manage the conversion of that bool into a felt using the logic of True <-> 1,  False <-> 0. This would be the functions generated:
```
fn WS0_WRITE_bool(loc: felt252, value: bool) -> bool{
  if value {
    WARP_STORAGE::write(loc, 1);
  }
  else {
    WARP_STORAGE::write(loc, 0);
  }
  return value;
}

fn WS1_READ_bool(loc: felt252) -> bool{
  let read0 = WARP_STORAGE::read(loc);
  read0 == 1
}
```
Otherwise, if the value is a number then it will simply apply the logic so far, generating something like this:
```
fn WS0_READ_u256(loc: felt252) -> u256{
    let read0 = WARP_STORAGE::read(loc);
    let read1 = WARP_STORAGE::read(loc + 1);
  u256{low:u128_from_felt252(read0),high:u128_from_felt252(read1)}
}

fn WS1_WRITE_u256(loc: felt252, value: u256) -> u256{
    WARP_STORAGE::write(loc, u128_to_felt252(value.low));
    WARP_STORAGE::write(loc + 1, u128_to_felt252(value.high));
  return value;
}
```

AlejandroLabourdette:
Actually about this
>If they (memoryRead and memoryWrite) will be required then I think is a good Idea to put this PR on hold until we address the rewriting (to Cairo1) of those scripts

I think we should merge this PR, since is blocking others like InferType, and then when they are rewritten treat the case of bools just like are managed in `storageRead` and `storageWrite`

src/cairoUtilFuncGen/storage/storageWrite.ts
```diff
@@ -58,14 +63,26 @@ export class StorageWriteGen extends StringIndexedFuncGen {
...
+        if (cairoTypeToWrite instanceof CairoBool) {
+          // It will store a felt corresponding to the bool value
+          // True -> `1`   False -> `0`
+          return endent`
+            if ${name} {
+              ${write(add('loc', index), `1`)}
+            }
+            else {
+              ${write(add('loc', index), `0`)}
+            }
```
piwonskp: I'd store that in warplib as `encodeBoolAsFelt` and simply call the function here. Something like `${write(add('loc', index), encodeBoolAsFelt(name))}`

src/cairoUtilFuncGen/storage/storageRead.ts
piwonskp: The more I think of it the more I'm convinced that we should use `into()` trait for those conversions (or make a similiar one). Perhaps there is into trait that translates bool -> felt and back?

src/cairoUtilFuncGen/storage/storageWrite.ts
piwonskp: Or implement (use exisiting if there is one) `Into<Bool, felt>`

src/cairoUtilFuncGen/storage/storageRead.ts
rodrodros: yes you're right I incorrectly thought when looking at `storageRead` that both flows (when it was a bool or not) ended in the same body expression, but there is a small change.

rodrodros: Nonetheless, I still stand by the point you should not define a equivalent type Cairo felt, and instead define a serialize read for `CairoBool`.  

Also an extra refactoring of the code generated so it's clear that the reading from storage stays the same, except for the return expression. This last point will also help with any other type that needs translation, since reading from memory will still be the same.

Smth like
```ts
readExpr = reads.map(..)
returnExpr = if bool: expr.Read == 1 else expr.Read

func_body = endent`
    {readExpr}
    {returnExpr}
`
```

src/cairoUtilFuncGen/storage/storageRead.ts
rodrodros, line 1: Ok, I think there was a misunderstanding. I believe adding another reading function is no good. Think we would need to do that for every different type such as all of `uX` or `ContractAddress`

rodrodros: If you notice the function that reads from storage does two things now:
1. It reads from memory where all type are treated as a felt
2. It shape those felts accordingly to the type we are expecting to read

(1) is the same for everyone, and should stay like that. What I meant in the past review comment is that you should add in `producePackExpression` the case for `bool` but return a `readFelt` and not a new `readBool`. If we a do a new `readType` for every new type, it is not extensible.

(2) For this part, you already have the felt data from (1), you only need to convert them in the type we are expecting. This way is easy to add new types. For example if it is a u8 or a bool, then the corresponding converting function is added. We don't need to worry about complex types because all of those are returned as a single `felt`

rodrodros: Your `readMemory` function in pseudocode should look smth like this:
```ts
let read = producePackExpression(...);
let returnType = // it takes the read data in form of a felt and make it into the desired return data type

function = `
    fn read_memory(...) -> T {
        ${read}
        ${returnType} 
   }
`
```

rodrodros (changes requested):

warplib/integer.cairo
```diff
@@ -13,3 +13,14 @@ fn get_u128_try_from_felt_result(value: felt252) -> u128 {
...
+impl Felt252IntoBool of Into::<felt252, bool> {
+    fn into(self: felt252) -> bool {
+        self == 1
+    }
+}
+impl BoolIntoFelt252 of Into::<bool, felt252> {
+    fn into(self: bool) -> felt252 {
+        if self { return 1; } else { return 0; }
+    }
+}
```
piwonskp: If we use `serialize` from Serde trait we won't even have to code it https://github.com/starkware-libs/cairo/blob/c2a04b3c640a7b6cec2940ce3990fb7f39ed81e4/corelib/src/serde.cairo#L20

piwonskp:
Conflicts

warplib/integer.cairo
```diff
@@ -13,3 +16,20 @@ fn get_u128_try_from_felt_result(value: felt252) -> u128 {
...
+    resp
+}
+
+fn bool_into_felt252(val: bool) -> felt252 {
+    let mut serialization_array: Array<felt252> = ArrayImpl::<felt252>::new();
+    BoolSerde::serialize(ref serialization_array, val);
+    let resp_option = ArrayImpl::pop_front(ref serialization_array);
+    let resp = OptionTraitImpl::<felt252>::unwrap(resp_option);
+    resp
+}
```
rodrodros: Although I am in favor of serde, I think this might be an overkill?

piwonskp: It'd be fine if we generalized that to a function `fn add_to_memory(val: Serde::<T>)` which would serialize any value and insert it into the memory

piwonskp: That might require some more work though so I'd be willing to leave it as it is and refactor the code to support any type in the next PR

piwonskp: WYT @rodrigo-pino ?

piwonskp:
Conflicts

warplib/integer.cairo
piwonskp: Missing newline

rodrodros: I agree @piwonskp

rodrodros (approved):

piwonskp (approved):

PR #967, opened 2023-03-03
uintN->uN

### LINKED ISSUE #964: uint/intX -> uX : cairo1

rjnrohit:
closes #964


src/cairoWriter/writers/literalWriter.ts
```diff
@@ -1,22 +1,21 @@
...
-            const [high, low] = divmod(BigInt(node.value), BigInt(Math.pow(2, 128)));
-            return [`Uint256(low=${low}, high=${high})`];
-          }
-          case 'felt':
-            return [node.value];
-          default:
-            throw new TranspileFailedError('Attempted to write unexpected cairo type');
+        if (
+          (primitiveTypeToCairo(node.typeString) as CairoPrimitiveIntType) ||
+          primitiveTypeToCairo(node.typeString) === 'felt'
```
piwonskp: Can we use typenodes instead? We should avoid using typestrings due to solc-typed-ast update. I suppose that might be difficult though since those are cairo nodes

rjnrohit: we're defining correspondence between sol int type typestring to cairo int type typestring, the function `primitiveTypeToCairo` takes for examples `uint24` as an input and returns `u24` as a output

piwonskp: I see, I'll think of that

src/utils/cairoTypeSystem.ts
```diff
@@ -169,6 +169,24 @@ export class CairoFelt extends CairoType {
   }
 }
 
+export class CairoUint extends CairoType {
```
piwonskp, line 172: It adresses #849 so it should be done as one of the first passes. We can merge that as a quick fix for cairo 1.0 but we need a long term solution so please add your comments to #849 @rjnrohit @cicr99 @rodrigo-pino @AlejandroLabourdette

src/cairoWriter/writers/literalWriter.ts
```diff
@@ -31,16 +30,13 @@ export class LiteralWriter extends CairoASTNodeWriter {
...
-            return [`0x${node.hexValue}`];
-          default:
-            throw new TranspileFailedError('Attempted to write unexpected cairo type');
+        if (
+          (primitiveTypeToCairo(node.typeString) as CairoPrimitiveIntType) ||
+          primitiveTypeToCairo(node.typeString) === 'felt'
+        ) {
+          return [`0x${node.hexValue}`];
+        } else {
+          throw new TranspileFailedError('Attempted to write unexpected cairo type');
```
piwonskp, line 52: Let's make a function that throws  the error if the condition is met, then return:
```
validate(node);
return [`0x${node.hexValue`]
```
This way it can be reused for hexString and number

src/utils/cairoTypeSystem.ts
```diff
@@ -169,6 +169,24 @@ export class CairoFelt extends CairoType {
...
+    super();
+  }
+  get fullStringRepresentation(): string {
+    return `[u${this.nBits}]`;
+  }
+  toString(): string {
+    return `u${this.nBits}`;
+  }
+  get width(): number {
+    return 1; // not sure about this width, but for the moment consider it as a felt
```
piwonskp: Where is the width used? I suppose when e.g. iterating over an array? Then I suppose it will always be 1 in cairo 1.0

src/utils/utils.ts
```diff
@@ -80,19 +80,49 @@ export function divmod(x: bigint, y: bigint): [bigint, bigint] {
...
+
+export function primitiveTypeToCairo(typeString: string): CairoPrimitiveIntType | 'felt' {
+  if (typeString === 'uint' || typeString === 'int') return 'u256';
+
+  if (typeString === 'fixed' || typeString === 'ufixed') {
+    throw new NotSupportedYetError('Fixed types not implemented');
+  }
+
+  if (typeString.startsWith('uint')) {
+    const bits = parseInt(typeString.slice(4));
```
piwonskp: In our code I've seen `'uint'.length` a few times and imo that is a great practice

src/utils/utils.ts
```diff
@@ -80,19 +80,49 @@ export function divmod(x: bigint, y: bigint): [bigint, bigint] {
...
+    if (bits > 256) {
+      throw new NotSupportedYetError('uint types larger than 256 bits not supported');
+    }
+    return `u${bits}` as CairoPrimitiveIntType;
   }
+
+  if (typeString.startsWith('int')) {
+    const bits = parseInt(typeString.slice(3));
```
piwonskp: Similarly to above

piwonskp (dismissed):
lgtm, some minor style changes

src/utils/cairoTypeSystem.ts
rjnrohit: I suppose it is used  as a felt slot in the memory and the storage system

rodrodros: In this case `width` should be equal to 2 when nBits === 256.

`width` represents felt size of the cairo type. For example a tuple of three felts has size 3, a tuple of three uint256 has size 6 etc. 

Reference types usually have size one representing their pointer to somewhere in memory or storage. But that can vary depending on the type and datalocation. Storage structs for example have the sum `width` of all its members, meanwhile memory structs are size one.

src/cairoWriter/writers/literalWriter.ts
```diff
@@ -31,16 +29,12 @@ export class LiteralWriter extends CairoASTNodeWriter {
...
-          case 'Uint256': {
-            return [
-              `Uint256(low=0x${node.hexValue.slice(32, 64)}, high=0x${node.hexValue.slice(0, 32)})`,
-            ];
-          }
-          case 'felt':
-            return [`0x${node.hexValue}`];
-          default:
-            throw new TranspileFailedError('Attempted to write unexpected cairo type');
+        if (primitiveTypeToCairo(node.typeString) as CairoPrimitiveIntType) {
```
rodrodros: This condition is always going to be truth, I guess you want to compare the string?
```suggestion
        if (primitiveTypeToCairo(node.typeString) === "smth?") {
```
Same thing happens in line 10

cicr99 (dismissed):

src/cairoWriter/writers/literalWriter.ts
```diff
@@ -31,16 +35,12 @@ export class LiteralWriter extends CairoASTNodeWriter {
...
-            return [
-              `u256_from_felts( ${node.hexValue.slice(32, 64)}, ${node.hexValue.slice(0, 32)} )`,
-            ];
-          }
-          case 'felt':
-            return [`0x${node.hexValue}`];
-          default:
-            throw new TranspileFailedError('Attempted to write unexpected cairo type');
+        if (isCairoPrimitiveIntType(type)) {
+          return [`0x${node.hexValue}_${type}`];
```
AlejandroLabourdette, line 46: Is it possible for this to write something like `0xsomeHex_u256`?. If so then we should consider this case because u256 are not avaible to create using this syntax since it is handled as a struct of 2 `u128`

rjnrohit: I have done the same before, but I had got some thing like: 
Screenshot 2023-04-04 at 1 35 37 PM
It seems `u256` literals are not supported yet

rjnrohit: good point  🙌 changes done!

AlejandroLabourdette: Nice 🚀

AlejandroLabourdette (dismissed):

warplib/memory.cairo
```diff
@@ -51,6 +206,163 @@ func wm_write_felt{warp_memory: DictAccess*}(loc: felt, value: felt) -> (res: fe
...
+func wm_write_240{warp_memory: DictAccess*}(loc: felt, value: u240) -> (res: u240) {
+    let (fvalue) = u240_to_felt(value);
+    dict_write{dict_ptr=warp_memory}(loc, fvalue);
+    return (value,);
+}
+func wm_write_248{warp_memory: DictAccess*}(loc: felt, value: u248) -> (res: u248) {
+    let (fvalue) = u248_to_felt(value);
+    dict_write{dict_ptr=warp_memory}(loc, fvalue);
+    return (value,);
+}
```
piwonskp, line 364: Why do you add cairo 0 code?

src/cairoUtilFuncGen/utils/uNselector.ts
```diff
@@ -0,0 +1,409 @@
...
+    case 'u240':
+      return WM_WRITE240;
+    case 'u248':
+      return WM_WRITE248;
+    case 'u256':
+      return WM_WRITE256;
+    default:
+      throw new Error('Invalid CairoUint type');
+  }
+};
```
piwonskp: We could do a function:
```
(bits) => [[...WARPLIB_MEMORY], `wm_read_${bits}`]
```
Would be less explicit but it wouldn't require us to write us so much code. I think it's a better option, WYT?

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -83,8 +84,8 @@ export class MemoryReadGen extends StringIndexedFuncGen {
       funcDef = this.requireImport(...WM_READ_ID, inputs, outputs);
     } else if (resultCairoType instanceof CairoFelt) {
       funcDef = this.requireImport(...WM_READ_FELT, inputs, outputs);
-    } else if (resultCairoType.fullStringRepresentation === CairoUint256.fullStringRepresentation) {
-      funcDef = this.requireImport(...WM_READ256, inputs, outputs);
+    } else if (resultCairoType instanceof CairoUint) {
+      funcDef = this.requireImport(...wmReaduNImport(resultCairoType), inputs, outputs);
```
piwonskp: Shouldn't it be `wmReadUNimport`?

warplib/memory.cairo
rjnrohit: This code is just place holder in `memory.cairo` file, since all codes in that file is in cairo0,  I added older format of cairo codes. When we work on memory.cairo file then we'll replace it

src/cairoUtilFuncGen/utils/uNselector.ts
rjnrohit: done, thank you!

src/cairoUtilFuncGen/memory/memoryWrite.ts
```diff
@@ -49,10 +49,15 @@ export class MemoryWriteGen extends StringIndexedFuncGen {
...
     if (cairoTypeToWrite instanceof CairoFelt) {
       return this.requireImport(...WM_WRITE_FELT, inputs, outputs);
-    } else if (
-      cairoTypeToWrite.fullStringRepresentation === CairoUint256.fullStringRepresentation
-    ) {
-      return this.requireImport(...WM_WRITE256, inputs, outputs);
+    }
+
+    if (cairoTypeToWrite instanceof CairoUint) {
+      this.requireImport(
```
cicr99: In the previous version there was a `return` statement here. Do we need to add one now? It appears you are creating two different functions

rjnrohit: Good catch!

piwonskp (approved):

src/utils/utils.ts
```diff
@@ -81,22 +81,79 @@ export function divmod(x: bigint, y: bigint): [bigint, bigint] {
...
+export const isCairoPrimitiveIntType = (x: string): x is CairoPrimitiveIntType => {
+  return cairoPrimitiveIntTypes.includes(x as CairoPrimitiveIntType);
+};
+
+export function primitiveTypeToCairo(
+  typeString: string,
+): CairoPrimitiveIntType | 'felt' | 'ContractAddress' {
+  if (typeString === 'address' || typeString === 'address payable') return 'ContractAddress';
+
+  if (typeString === 'uint' || typeString === 'int') return 'u256';
```
rodrodros, line 129: Wasn't `uint256` typestring "uint256" and not "uint"?

warplib/memory.cairo
```diff
@@ -40,7 +40,162 @@ func wm_read_felt{warp_memory: DictAccess*}(loc: felt) -> (val: felt) {
     return (res,);
 }
 
-func wm_read_256{warp_memory: DictAccess*}(loc: felt) -> (val: Uint256) {
+func wm_read_8{warp_memory: DictAccess*}(loc: felt) -> (val: u8) {
```
rodrodros, line 43: This functions are written in Cairo 0 👀.  The path to memory we are taking is defining a Struct (#1025) and then defining an implementation for it. You should extend the implementation of it using a Trait. I think using a separate file could be good as well. @piwonskp what do you think?

piwonskp: I wonder if it's possible to implement it with warp memory trait + parametrized method + into trait. It calls for a new PR tho

piwonskp:
Conflicts

src/utils/utils.ts
rjnrohit: anyway , It's being handled at line no .137 and 138

rjnrohit:
> Conflicts

resolved

src/utils/utils.ts
rodrodros: Right, so should we delete this case or not? I think `uint` only as a type string only does not exists

warplib/memory.cairo
rodrodros: Ok, created an issue #1031

src/utils/utils.ts
rjnrohit: It was there before, so I think we should keep it

src/passes/cairoUtilImporter.ts
```diff
@@ -42,7 +42,7 @@ export class CairoUtilImporter extends ASTMapper {
 
   visitElementaryTypeName(node: ElementaryTypeName, ast: AST): void {
     const cairoType = primitiveTypeToCairo(node.name);
-    if (cairoType === 'Uint256') {
+    if (primitiveTypeToCairo(node.name) === 'u256') {
```
rodrodros: You can use cairoType in this conditional as well

src/utils/utils.ts
rodrodros: Yes, we should keep it. I went and check and `uint` is used as a valid typename

rodrodros (approved):

PR #968, opened 2023-03-03
Imports path structure

cicr99:
Changes addressed in this PR:
- Change the imports writer to be compatible with Cairo1
- The path to the imports in the `CairoImportFunctionDefinition` node changed to be an array of strings instead of a raw string
- There is  a new file `src/utils/importFuncs` that contains the definitions for all the import paths. This way, if an import needs to be changed, it is only necessary to change it here and not in the entire code.

Issues:
- Warplib generated functions imports were not fixed, even though they are not compatible with cairo1, because warplib functions need to be modified as well, which is a different task.
- All of the parsing to get the implicits of warplib functions was also not changed, as implicits are probably being removed


src/utils/importFuncs.ts
```diff
@@ -0,0 +1,163 @@
+function cairoCommonPath(path: string[]): string[] {
+  return ['starkware', 'cairo', 'common', ...path];
+}
```
piwonskp: Let's have it as a const, and then we'd do:  `const ALLOC = [[...CAIRO_COMMON_PATH, 'alloc'], 'alloc']`

src/utils/importFuncGenerator.ts
```diff
@@ -24,7 +24,7 @@ const STARKWARE_CAIRO_COMMON_UINT256 = 'starkware.cairo.common.uint256';
 const STARKWARE_STARKNET_COMMON_SYSCALLS = 'starkware.starknet.common.syscalls';
```
piwonskp: Consts above should be removed

src/utils/importFuncGenerator.ts
```diff
@@ -64,12 +64,12 @@ export function createImport(
...
 
-  const warplibFunc = warplibImportInfo.get(path)?.get(name);
+  const warplibFunc = warplibImportInfo.get(path.join('.'))?.get(name);
   if (warplibFunc !== undefined) {
     return createFuncImport(...warplibFunc);
   }
 
-  switch (path + name) {
+  switch ([...path, name].join('.')) {
     case STARKWARE_CAIRO_COMMON_ALLOC + 'alloc':
```
piwonskp: Let's import from `importFuncs` and compare it as arrays instead

src/utils/importFuncs.ts
cicr99: At first I thought not to use constants in order to not cause reference issues. But you are right, these should be changed to variables instead.

src/utils/importFuncGenerator.ts
cicr99: I hadn't given too much thought about this as Implicits should be remove soon, but it's true, this can be improved as well

src/warplib/utils.ts
```diff
@@ -251,7 +249,7 @@ export function BoolxBoolFunction(node: BinaryOperation, name: string, ast: AST)
 
   const importedFunc = ast.registerImport(
     node,
-    `warplib.maths.${name}`,
+    ['warplib', 'maths', `${name}`],
```
piwonskp: I guess name is a string, so string interpolation is probably not necessary

src/warplib/utils.ts
```diff
@@ -205,7 +203,7 @@ export function IntFunction(
 
   const importedFunc = ast.registerImport(
     node,
-    `warplib.maths.${fileName}`,
+    ['warplib', 'maths', `${fileName}`],
```
piwonskp: Same as above

src/passes/builtinHandler/mathsOperationToFunction.ts
```diff
@@ -107,7 +107,7 @@ export class MathsOperationToFunction extends ASTMapper {
         const name = `warp_${node.vExpression.name}`;
         const importedFunc = ast.registerImport(
           node,
-          `warplib.maths.${node.vExpression.name}`,
+          ['warplib', 'maths', `${node.vExpression.name}`],
```
piwonskp: Can do `...WARPLIB_MATHS`

src/passes/bytesConverter.ts
```diff
@@ -109,7 +109,7 @@ export class BytesConverter extends ASTMapper {
 
     const importedFunc = ast.registerImport(
       node,
-      'warplib.maths.bytes_access',
+      ['warplib', 'maths', 'bytes_access'],
```
piwonskp: `...WARPLIB_MATHS`

src/utils/importFuncGenerator.ts
```diff
@@ -64,57 +77,45 @@ export function createImport(
     );
   const createStructImport = () => createCairoImportStructDefinition(name, path, ast, sourceUnit);
 
-  const warplibFunc = warplibImportInfo.get(path)?.get(name);
+  const warplibFunc = warplibImportInfo.get(path.join('.'))?.get(name);
```
piwonskp: If it doesn't break anything we can fix that in another PR. Let me know what you think @rodrigo-pino @cicr99

src/cairoWriter/writers/sourceUnitWriter.ts
```diff
@@ -57,13 +57,11 @@ export class SourceUnitWriter extends CairoASTNodeWriter {
...
-        .sort((funcA, funcB) =>
-          `${funcA.path}.${funcA.name}`.localeCompare(`${funcB.path}.${funcB.name}`),
-        )
-        .filter((func, index, importFuncs) => func.name !== importFuncs[index - 1]?.name),
-    ).reduce((writtenImports, importFunc) => `${writtenImports}\n${importFunc}`, '');
+    const writtenImportFuncs = importFunctions
+      .map((n) => writer.write(n))
+      .sort((importA, importB) => importA.localeCompare(importB))
+      .filter((func, index, importFuncs) => func !== importFuncs[index - 1])
+      .reduce((writtenImports, importFunc) => `${writtenImports}\n${importFunc}`, '');
```
piwonskp: Btw, do I understand correctly the reduce could be replaced by `.join('\n')` @rodrigo-pino?

src/utils/importPaths.ts
```diff
@@ -0,0 +1,103 @@
+const CAIRO_COMMON_PATH = ['starkware', 'cairo', 'common'];
```
rodrodros, line 1: It look good to me. The only thing I would change and it is a nitpick is naming of some constants:
* const which uses warp memory method I would prefix them with `WM_`

src/utils/importFuncGenerator.ts
cicr99: Right now that needs to be done because the map `warplibImportInfo` is constructed  that way. It makes sense that if we want the path to be the key, it needs to be a hashable value, hence, the path (which is an array) needs to be transformed into a string. I discussed about this with @rodrigo-pino  when I was implementing the changes and the main purpose of this logic is to collect implicits for the imported functions of warplib, so, to answer your question, yes, I think this could be left to do in another PR, maybe when addressing implicits.

src/cairoWriter/writers/sourceUnitWriter.ts
cicr99: Yep, totally. Again, I reused the old logic and didn't notice about this. Will change it now.

src/utils/importFuncGenerator.ts
cicr99: Just noticed we shouldn't compare arrays in the switch case as it checks for the references.

- If we want to keep the switch case we could generate a string for the corresponding path and then compare those
- Another idea might be to have an array storing the corresponding function for each path and instead of having a swith case, iterate through all the elements comparing the paths element by element until a match is found, then that function can be executed.

Do you have any other idea @piwonskp  @rodrigo-pino ?

piwonskp: I'll add my constant phrase: lodash would fix it :D [link](https://lodash.com/docs/#isEqual)

rodrodros: Do you mean having a Trie of paths? I think the switch case statement will live as long we update the code for not to need implicits. Which can be done at anytime, our Cairo1-ERC20 code already does not use implicits.

rodrodros: I am of the opinion of leaving as it is, since it works, and it won't exist soon. Dev time can be better spent somewhere else

src/utils/importFuncGenerator.ts
rodrodros: Yes I think all that code will get reduced considerably once we no longer need to track implicits anymore. So I think it's best to wait to update the way Warp treats implicits

src/cairoWriter/writers/sourceUnitWriter.ts
piwonskp: Yea, I've spotted that's not your logic. Caught my eye though, so had to mention :)

src/utils/importFuncGenerator.ts
piwonskp: If that's temporary and we won't need it in future then we can use strings, whatever is faster to implement for you @cicr99

src/cairoUtilFuncGen/storage/mappingIndexAccess.ts
```diff
@@ -229,8 +228,8 @@ export class MappingIndexAccessGen extends CairoUtilFuncGenBase {
           }
       `,
       functionsCalled: [
-        // this.requireImport(['array'], 'ArrayTrait'),
-        // this.requireImport(['warplib', 'string_hash'], 'string_hash'),
+        this.requireImport(['array'], 'ArrayTrait'),
```
rodrodros: Do you think this PR is a good place to start adding imports for Cairo1 code? 
```suggestion
        this.requireImport(...ARRAY_TRAIT),
```

piwonskp: Yeah I agree, let's just separate them by the comment line from the old imports

src/utils/importFuncGenerator.ts
piwonskp: I see, then let's leave it for later

piwonskp (approved):
Good job!

rodrodros (approved):

PR #973, opened 2023-03-07
modify syntax warplib memory

rjnrohit:


warplib/memory.cairo
```diff
@@ -35,367 +21,346 @@ from warplib.maths.utils import felt_to_uint256, narrow_safe
...
+
+fn wm_read_felt(ref warp_memory: DictFeltTo::<felt> ,  loc: felt) -> felt {
+    warp_memory.get(loc)
 }
 
-func wm_read_256{warp_memory: DictAccess*}(loc: felt) -> (val: Uint256) {
-    let (low) = dict_read{dict_ptr=warp_memory}(loc);
-    let (high) = dict_read{dict_ptr=warp_memory}(loc + 1);
-    return (Uint256(low, high),);
+fn wm_read_256(ref warp_memory: DictFeltTo::<felt> ,  loc: felt) -> u256 {
```
rodrodros, line 38: I think `warp_memory` should remain an implicit, if we are doing a direct translation, otherwise it would require extensive changes in multiple places in the code

rjnrohit: I didn't study much about implicts in cairo1 , will look into it and then I'll modify it

rjnrohit:
As we are writing a plugin for warp memory system, this is not needed anymore!

PR #975, opened 2023-03-07
Create u256 values from 2 felts

AlejandroLabourdette:


piwonskp (approved):
lgtm

warplib/integer.cairo
```diff
@@ -0,0 +1,15 @@
+use integer::u128_try_from_felt;
```
rodrodros, line 1: All logic related with conversions are in `warplib/math/utils.cairo` I would put these functions there

AlejandroLabourdette: If we move it there then we would need to translate the utils file. In there we have div operations on felts that are not currently supported by cairo1, so we would need to found a way around it. Since we are in a sprint to have erc20 transpiled and they are not used there I think would be best to leave it this way until we address the task to transpile the warplib/maths/utils library

rodrodros (approved):

PR #979, opened 2023-03-08
Erc20 storage fixes

piwonskp:


src/cairoUtilFuncGen/calldata/calldataToStorage.ts
```diff
@@ -169,8 +169,8 @@ export class CalldataToStorageGen extends StringIndexedFuncGen {
       `   let (index_uint256) = warp_uint256(index);`,
       `   let (elem_loc) = ${dynArray.name}.read(loc, index_uint256);`,
       `   if (elem_loc == 0){`,
-      `       let (elem_loc) = WARP_USED_STORAGE.read();`,
-      `       WARP_USED_STORAGE.write(elem_loc + ${cairoElementType.width});`,
+      `       let (elem_loc) = WARP_USED_STORAGE::read();`,
+      `       WARP_USED_STORAGE::write(elem_loc + ${cairoElementType.width});`,
```
rodrodros: Why do we changed just a part of the function, now it is in two different languages. I think is best to revert this changes here, unless we decide to update the whole function

rodrodros: This happens across several writers

src/cairoUtilFuncGen/serialisation.ts
```diff
@@ -48,18 +49,35 @@ function producePackExpression(type: CairoType): (string | Read)[] {
...
   if (type instanceof CairoStruct) {
+    if (type.fullStringRepresentation === CairoUint256.fullStringRepresentation) {
+      return [
+        type.name,
+        '{',
+        ...[...type.members.entries()]
+          .flatMap(([memberName, memberType]) => [
+            memberName,
+            ':',
+            'get_u128_try_from_felt_result(',
```
rodrodros: Inserting an imported function name without the import is prone to errors. What do you want to accomplish here?

src/cairoUtilFuncGen/calldata/calldataToStorage.ts
piwonskp: I've just used search and replace of the pattern. Forgot that there are functions not translated to cairo 1.0 yet

src/cairoUtilFuncGen/serialisation.ts
piwonskp: Well, warplib uint256 used felt, cairo 1.0 uint256 uses 2x uint128 so felts have to be casted to u128

piwonskp: @AlejandroLabourdette made the function `u256_from_felts` but we should start using u128 directly instead. The change imo deserves a separate PR tho

rjnrohit: Yeah, I'm thinking we should use `u128_from_felt` builtin function instead and for u256 we can use `u256{u128_from_felt(a), u128_from_felt(b)}`

piwonskp: Afaik `u128_from_felt` would be only necessary when transforming literals right now. So we could pretty much get rid of felts I think. In the long term literals will probably also allow to use them as u128 without the conversion so we could get rid of the function then

src/utils/importPaths.ts
```diff
@@ -107,3 +98,12 @@ export const WM_WRITE_FELT: [string[], string] = [[...WARPLIB_MEMORY], 'wm_write
 
 export const ARRAY_TRAIT: [string[], string] = [['array'], 'ArrayTrait'];
 export const U256_FROM_FELTS: [string[], string] = [['warplib', 'integer'], 'u256_from_felts'];
+export const UINT256: [string[], string] = [
```
cicr99: I would use a different name for this import, maybe something like `GET_U128` as the name of the function is pretty long.
There are many places in the code that uses `UINT256` import, I don't think this is what it is expected.

src/passes/builtinHandler/msgSender.ts
```diff
@@ -23,7 +30,23 @@ export class MsgSender extends ASTMapper {
...
+        ...ADDRESS_INTO_FELT,
+        [],
+        [['uint256', new ElementaryTypeName(ast.reserveId(), '', 'uint256', 'uint256')]],
+      );
+      const asFelt = new FunctionCall(
+        ast.reserveId(),
+        '',
+        'uint256',
+        FunctionCallKind.FunctionCall,
+        new MemberAccess(ast.reserveId(), '', 'uint256', replacementCall, 'into', IntoTrait.id),
```
cicr99, line 45: Why is the `INTO` import not created here instead of the `cairoUtilImporter`? It isn't being created anywhere else, so why have it separately?

src/cairoUtilFuncGen/serialisation.ts
rodrodros: StorageRead and MemoryRead function generators are the ones which call this function, maybe we can the imports on those files when there is uint256

src/utils/importPaths.ts
piwonskp: Actually that is expected, since it has to be imported everytime u256 struct is being constructed. I agree that's ugly and hackish, but as I suggested above I'd see a separate PR dealing with that. The function shouldn't be used, thus there is no point in debating over imports right now, we'll just remove them once we stop using it

cairo1/darwin_arm64/corelib/src/lib.cairo
```diff
@@ -413,3 +413,13 @@ mod test;
...
+
+mod warplib;
+use warplib::get_u128_try_from_felt_result;
+use warplib::u256_from_felts;
+
+use warplib::maths::warp_add256;
+use warplib::maths::warp_external_input_check_address;
+use warplib::maths::warp_external_input_check_int256;
+use warplib::maths::warp_ge256;
+use warplib::maths::warp_sub_unsafe256;
```
rjnrohit: new line

cairo1/darwin_arm64/corelib/src/warplib.cairo
```diff
@@ -0,0 +1,10 @@
+mod integer;
+use integer::get_u128_try_from_felt_result;
+use integer::u256_from_felts;
+
+mod maths;
+use maths::warp_add256;
```
rjnrohit, line 6: I think all these `use` statements are redudant, because anyway we are using full path during import (i.e `use warplib::maths::add::warp_add256;`)


I would suggest get rid of all `use` statments.
Only keep mod declarations

cairo1/darwin_arm64/corelib/src/warplib/maths.cairo
```diff
@@ -0,0 +1,10 @@
+mod add;
+use add::warp_add256;
+mod external_input_check_address;
+use external_input_check_address::warp_external_input_check_address;
+mod external_input_check_ints;
+use external_input_check_ints::warp_external_input_check_int256;
+mod ge;
+use ge::warp_ge256;
+mod sub_unsafe;
+use sub_unsafe::warp_sub_unsafe256;
```
rjnrohit: new line

rjnrohit (changes requested):
Still I'm not clear why don't we use builtin `u128_from_felt` function instead of `get_u128_try_from_felt_result`.

cairo1/darwin_arm64/corelib/src/warplib/maths/div_signed.cairo
```diff
@@ -0,0 +1,549 @@
...
...arp_int80_to_int256, warp_int88_to_int256, warp_int96_to_int256, warp_int104_to_int256, warp_int112_to_int256, warp_int120_to_int256, warp_int128_to_int256, warp_int136_to_int256, warp_int144_to_int256, warp_int152_to_int256, warp_int160_to_int256, warp_int168_to_int256, warp_int176_to_int256, warp_int184_to_int256, warp_int192_to_int256, warp_int200_to_int256, warp_int208_to_int256, warp_int216_to_int256, warp_int224_to_int256, warp_int232_to_int256, warp_int240_to_int256, warp_int248_to_int256, warp_int256_to_int8, warp_int256_to_int16, warp_int256_to_int24, warp_int256_to_int32, warp_int256_to_int40, warp_int256_to_int48, warp_int256_to_int56, warp_int256_to_int64, warp_int256_to_int72, warp_int256_to_int80, warp_int256_to_int88, warp_int256_to_int96, warp_int256_to_int104, warp_int256_to_int112, warp_int256_to_int120, warp_int256_to_int128, warp_int256_to_int136, warp_int256_to_int144, warp_int256_to_int152, warp_int256_to_int160, warp_int256_to_int168, warp_int256_to_int176, warp_int256_to_int184, warp_int256_to_int192, warp_int256_to_int200, warp_int256_to_int208, warp_int256_to_int216, warp_int256_to_int224, warp_int256_to_int232, warp_int256_to_int240, warp_int256_to_int248
```
rjnrohit, line 6: these are non-cairo1 codes

cairo1/darwin_arm64/corelib/src/warplib.cairo
piwonskp: Warplib is reworked by @rodrigo-pino, moreover we'll probably remove `maths` in near future as most functions there are covered by corelib. Let's leave it as it is for now

cairo1/darwin_arm64/corelib/src/warplib/maths/div_signed.cairo
piwonskp: We'll remove those with the proper support of cairo 1 uints. If we remove those in this PR we should also remove their imports etc.

src/utils/importPaths.ts
piwonskp: Name of constant changed

src/passes/builtinHandler/msgSender.ts
piwonskp: Unused imports are deleted. I cannot assign 2 imports to a single function call and that's what I have to do (I have to import Into trait and Into implementation for the struct). Thus I had to hack the system and do the import in cairoUtilImporter

piwonskp:
> Still I'm not clear why don't we use builtin `u128_from_felt` function instead of `get_u128_try_from_felt_result`.

I've replaced that. Anyway in a few weeks we shouldn't be using either of them

src/cairoUtilFuncGen/serialisation.ts
piwonskp: Imo we shouldn't overcomplicate things right now. We've got `u128` literals. In Solidity there is no such type as felt so abi inputs won't provide `felt` as well. We need it now only because the old warp operates on felts

piwonskp: The point is imo the function should be removed soon, so is there any sense in considering when to import it then?

rodrodros: Ok, I was worried that it could be an obscure bug further down the line. But if it won't exists later I agree

rodrodros (approved):

rjnrohit (approved):

src/passes/builtinHandler/msgSender.ts
cicr99: Oh, I see the problem. Maybe we could think of a better solution for this later, let's leave it as it is for now.

PR #969, opened 2023-03-03
Updates for Cairo Writer to get closer to Cairo1 syntax

AlejandroLabourdette:
This PR substitute #924 and target `cairo-1.0` branch

Writers updated:
- fix #939 
- fix #938
- fix #927
- fix #926
- fix #934 
- fix #931 
- fix #937 
- fix #934 
- fix #935
- fix #929
- fix #928 
- fix #930 
- fix #932 
- fix #940 
- fix #933
- fix #936

PR #983, opened 2023-03-09
Test erc20

piwonskp:


src/cairoWriter/writers/cairoAssertWriter.ts
```diff
@@ -6,6 +6,6 @@ export class CairoAssertWriter extends CairoASTNodeWriter {
   writeInner(node: CairoAssert, writer: ASTWriter): SrcDesc {
     const expression = writer.write(node.vExpression);
     const message = node.assertMessage ?? 'Assertion error';
-    return [`assert( ${expression}, "${message}" );`];
+    return [`assert( ${expression}, '${message}' )`];
```
piwonskp: Shall we remove space after `(` and before `)`?

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -128,13 +128,11 @@ export class CairoContractWriter extends CairoASTNodeWriter {
...
 
-    const writtenImportFuncs = getGroupedImports(
-      importFunctions
-        .sort((funcA, funcB) =>
-          `${funcA.path}.${funcA.name}`.localeCompare(`${funcB.path}.${funcB.name}`),
-        )
-        .filter((func, index, importFuncs) => func.name !== importFuncs[index - 1]?.name),
-    ).reduce((writtenImports, importFunc) => `${writtenImports}\n${importFunc}`, '');
+    const writtenImportFuncs = importFunctions
+      .map((n) => writer.write(n))
```
piwonskp, line 132: `.map(writer.write, writer)` would be more idiomatic

src/transpiler.ts
```diff
@@ -153,7 +153,6 @@ function applyPasses(
     ['An', AnnotateImplicits],
     ['Lv', IfStatementTempVarPostpender],
     ['Ci', CairoUtilImporter],
-    ['Rim', ReplaceIdentifierContractMemberAccess],
```
piwonskp, line 156: Unused import

src/cairoWriter/writers/cairoAssertWriter.ts
rodrigo-pino: Yes, I think it is better that way

src/cairoWriter/writers/cairoContractWriter.ts
cicr99: That's ok, but in the file it is always used the way it is currently written. I could change them all there, but would need to check about other files

src/cairoWriter/writers/cairoAssertWriter.ts
cicr99: Yes, they can be removed

src/cairoWriter/writers/cairoContractWriter.ts
cicr99: Besides, you get incompatible types doing it that way. I don't think it's worth it.

piwonskp: It's used both ways actually in the code, but if that causes incompatible type problems then let's leave it as it is and don't split hairs here right now

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -98,19 +100,84 @@ export class CairoContractWriter extends CairoASTNodeWriter {
...
+    const contractHeader = '#[contract] \n' + `mod ${node.name} {`;
+
     return [
       [
+        contractHeader,
         documentation,
-        ...events,
-        `namespace ${node.name}{\n\n${body}\n\n}`,
-        outsideNamespaceBody,
+        writtenImportFuncs,
```
rodrodros, line 165: Shouldn't the written importFuncs be on the top level of the script?

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
```diff
@@ -40,17 +40,16 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
...
         .join('\n'),
     ];
   }
 
   private getDecorator(node: CairoFunctionDefinition): string[] {
-    if (node.kind === FunctionKind.Constructor) return ['@constructor'];
+    if (node.kind === FunctionKind.Constructor) return ['#[constructor]'];
     const decorators: string[] = [];
     if (node.kind === FunctionKind.Fallback) {
       decorators.push('@raw_input');
```
piwonskp: @rjnrohit commented on the original PR https://github.com/NethermindEth/warp/pull/969#discussion_r1125955293

src/cairoWriter/writers/cairoContractWriter.ts
cicr99: Well, from what I saw in [ERC20 example in cairo1](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/test_data/erc20.cairo), apparently they use it this way. Please let me know otherwise so I can change it

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
cicr99: Done!

src/cairoWriter/writers/cairoContractWriter.ts
rodrodros: Yes it seems module related imports are defined inside the module. I agree with leaving them here

cicr99: I saw this other [example](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/test_data/account.cairo) where they use imports outside of the module as well, at the top of the file as you had suggested; but I think this is something else  we can work on later. I made a new issue to address this task and explain it better https://github.com/NethermindEth/warp/issues/997

rodrodros (approved):

PR #993, opened 2023-03-16
Cairo1 warplib infra

### LINKED ISSUE #1000: Build infrastructure to import from warplib

rodrodros:
Fix #1000 
Following Cairo1 requirements to have functional warplib:
Treat all transpiled contracts as cairo project with the right path
Treat warplib as a public library


src/warplib/generateWarplib.ts
```diff
@@ -64,3 +67,20 @@ export const warplibFunctions: WarplibFunctionInfo[] = [
...
+  `,
+);
+
+// generate warplib/maths lib
+fs.writeFileSync(
+  path.join('.', 'warplib', 'maths', 'lib.cairo'),
+  endent`
+    // Add here maths modules in cairo1
+  `,
+);
```
piwonskp, line 94: Is there any reason to have that generated? I mean won't static files do the job?

rodrodros: I am not sure I understand what you mean. 

My first attempt was to make `maths` a sub module of `warplib`. So to import anything from `maths` we would do something like `warplib::maths::func`, but when `maths` is set as a submodule of `warplib` it expects `math` to be a Cairo file and not a folder containing other cairo files

piwonskp: I mean why don't we simply commit those files to repo? Why do we need `fs.writeFileSync(`? You write static string to a static path anyway

rodrodros: Yes indeed, it makes sense for warplib. In case of maths, since it will change, is better to leave it as autogenerated

src/cli.ts
```diff
@@ -139,6 +141,25 @@ export function runTranspile(files: string[], options: CliOptions) {
   }
 }
 
+function createCairoProject(filePath: string): void {
+  // create cairo_project.toml
+  const projectRoot = path.dirname(__dirname);
+  const cairoProjectPath = path.join(path.dirname(filePath), '/cairo_project.toml');
```
piwonskp: If you use path.join I think `/` is unneeded

src/cli.ts
```diff
@@ -139,6 +141,25 @@ export function runTranspile(files: string[], options: CliOptions) {
   }
 }
 
+function createCairoProject(filePath: string): void {
+  // create cairo_project.toml
+  const projectRoot = path.dirname(__dirname);
```
piwonskp: `warpRoot` maybe?

src/cli.ts
```diff
@@ -139,6 +141,25 @@ export function runTranspile(files: string[], options: CliOptions) {
...
+  const cairoProjectPath = path.join(path.dirname(filePath), '/cairo_project.toml');
+  outputFileSync(
+    cairoProjectPath,
+    endent`
+      [crate_roots]
+      root = "."
+      warplib = "${projectRoot}/warplib"
+      maths = "${projectRoot}/warplib/maths"
+    `,
```
piwonskp, line 156: It could be module level constant, since project.toml contents are the same for all outputs throughout the run

piwonskp: Actually... should they? Is projectRoot absolute or relative path?

rodrodros: projectRoot is an absolute path. It varies depending on the machine

src/cli.ts
rodrodros: Removed `projectRoot`.

src/cli.ts
```diff
@@ -139,6 +141,24 @@ export function runTranspile(files: string[], options: CliOptions) {
...
+  outputFileSync(
+    cairoProjectPath,
+    endent`
+      [crate_roots]
+      root = "."
+      warplib = "${warplibRoot}"
+    `,
+  );
+  // create lib.cairo
+  const libPath = path.join(path.dirname(filePath), '/lib.cairo');
```
AlejandroLabourdette: The slash is automatically inserted by `path.join`
```suggestion
  const libPath = path.join(path.dirname(filePath), 'lib.cairo');
```

src/warplib/generateWarplib.ts
```diff
@@ -15,8 +15,14 @@ import { shr, shr_signed } from './implementations/maths/shr';
 import { sub_unsafe, sub_signed, sub_signed_unsafe } from './implementations/maths/sub';
 import { bitwise_not } from './implementations/maths/bitwiseNot';
 import { external_input_check_ints } from './implementations/external_input_checks/externalInputChecksInts';
+import path from 'path';
+import * as fs from 'fs';
+import endent from 'endent';
+import { warplibImportInfo } from './gatherWarplibImports';
```
AlejandroLabourdette: This import is unused

src/cli.ts
```diff
@@ -139,6 +141,24 @@ export function runTranspile(files: string[], options: CliOptions) {
...
 
+function createCairoProject(filePath: string): void {
+  // create cairo_project.toml
+  const cairoProjectPath = path.join(path.dirname(filePath), 'cairo_project.toml');
+  const warplibRoot = path.join(path.dirname(__dirname), 'warplib');
+  outputFileSync(
+    cairoProjectPath,
+    endent`
+      [crate_roots]
+      root = "."
```
piwonskp, line 152: Does it work when you run the command from outside warp root?

src/cli.ts
piwonskp: Yeah, during the run it won't change, so it's reasonable to calculate it once on the module level

rodrodros: Done

piwonskp: I meant that you can make this module level constant: 
```
const CAIRO_PROJECT_TOML_CONTENT = endent`
  [crate_roots]
  root = "."
  warplib = "${WARPLIB_ROOT}"
`
```

warplib/maths/sub.cairo
```diff
@@ -1,30 +1,27 @@
...
+    // // preemptively borrow from bit128
+    // let (low_safe) = bitwise_and(BOUND128 + lhs.low - rhs.low, MASK128);
+    // let low_unsafe = lhs.low - rhs.low;
+    // if (low_safe == low_unsafe) {
+    //     // the borrow was not used
+    //     return (Uint256(low_safe, lhs.high - rhs.high),);
+    // } else {
+    //     // the borrow was used
+    //     return (Uint256(low_safe, lhs.high - rhs.high - 1),);
+    // }
```
piwonskp: Let's remove commented lines. Imo we don't need them

src/warplib/generateWarplib.ts
```diff
@@ -75,11 +74,14 @@ const mathsContent: string = glob
...
-    const useFuncNames = parseMultipleRawCairoFunctions(rawCairoCode)
-      .map(({ name }) => `use ${fileName}::${name};`)
-      .join('\n');
-
-    return `mod ${fileName}\n${useFuncNames};`;
+    const funcNames = parseMultipleRawCairoFunctions(rawCairoCode).map(({ name }) => name);
+    return { fileName, funcNames };
+  })
+  // Filter required to deal with both cairo1 and cairo0.10 functions
+  .filter(({ funcNames }) => funcNames.length > 0)
```
piwonskp, line 81: Let's add `TODO:  remove once all warplib functions are ported to cairo 1`

piwonskp (approved):

AlejandroLabourdette (approved):

PR #1004, opened 2023-03-24
[Cairo 1] Static arrays implementation

piwonskp:
The PR main goal is to create a base on which we can work with warp and plugin on both ends. Next PRs will bring WarpMemory struct, Solidity structs and dynamic arrays to warp 

The code:
```solidity
pragma solidity ^0.8.14;

contract C {
  function f() public {
    uint8[3] memory t = [1, 2, 3];
  }
}
```
is translated into:
```
#[contract] 
mod C {
use array::Array;
use array::ArrayTrait;
use integer::u32_to_felt;

struct Storage {
  WARP_STORAGE: LegacyMap::<felt, felt>,
  WARP_USED_STORAGE: felt,
  WARP_NAMEGEN: felt,
  
}

fn readId(loc: felt) -> felt {
  let id = WARP_STORAGE::read(loc);
  if id == 0 {
    let id = WARP_NAMEGEN::read();
    WARP_NAMEGEN::write(id + 1);
    WARP_STORAGE::write(loc, id + 1);
    return id + 1;
  } 
  return id;
}
    // Dynamic variables - Arrays and Maps
    // Static variables
    #[external]
    fn f_26121ff0(){
    let mut warp_memory: Array::<felt> = ArrayTrait::new();
        let t = wm0_static_array(1, 2, 3);
        return ();
    }

    #[constructor]
    fn constructor(){
        return ();
    }

#[implicit(warp_memory)]
fn wm0_static_array(e0: felt, e1: felt, e2: felt) -> felt {
  let start = warp_memory.len();
  warp_memory.append(e0);
  warp_memory.append(e1);
  warp_memory.append(e2);
  return u32_to_felt(start);
}
}
```


src/utils/importPaths.ts
```diff
@@ -96,10 +96,13 @@ export const WM_WRITE_FELT: [string[], string] = [[...WARPLIB_MEMORY], 'wm_write
 
 //------------------------------------------------------
 
+export const ARRAY: [string[], string] = [['array'], 'Array'];
 export const ARRAY_TRAIT: [string[], string] = [['array'], 'ArrayTrait'];
 export const U256_FROM_FELTS: [string[], string] = [['warplib', 'integer'], 'u256_from_felts'];
 export const GET_U128: [string[], string] = [['integer'], 'u128_from_felt'];
 export const U128_TO_FELT: [string[], string] = [['integer'], 'u128_to_felt'];
+export const GET_U32: [string[], string] = [['integer'], 'u32_from_felt'];
+export const U32_TO_FELT: [string[], string] = [['integer'], 'u32_to_felt'];
```
rodrodros: nitpick: rename `GET_U32` to `U32_FROM_FELT`. 

I notice that we have a `GET_U128` as well. In case of accepting the suggestion, that should be renamed as well

src/utils/implicits.ts
```diff
@@ -41,9 +41,6 @@ export function registerImportsForImplicit(ast: AST, node: ASTNode, implicit: Im
     case 'bitwise_ptr':
       ast.registerImport(node, ...BITWISE_BUILTIN, [], []);
       break;
-    case 'warp_memory':
-      ast.registerImport(node, ...DICT_ACCESS, [], []);
-      break;
```
rodrodros, line 46: Why delete this import?

piwonskp: The import breaks compilation since it is not a valid (existing) path

piwonskp: Also I think warp memory imports will be handled on plugin side since it is adding initialization and warp memory arguments

JorikSchellekens:
Is the absence of an initialisation of warpmemory intentional?

piwonskp:
> Is the absence of an initialisation of warpmemory intentional?

Yes, I removed that in [9ee8878](https://github.com/NethermindEth/warp/pull/1004/commits/9ee88785bfbfe5dca8682a4747ecfdc43715e012) since we've decided the plugin will handle that.

JorikSchellekens:
I don't think the plugin should handle the case of instantiating implicits.

Handling instantiation in the plugin is not generalisable. Instantiation could be any number of arbitrary configuration. I see the plugin as something we can offer to the community and it should be independent of Warp. 

Also, not instantiating implicits in the plugin is safer. The only check the plugin should make is whether the implicit is defined in the scope. If it isn't it should err. Otherwise it will be super easy to get cases of "split brain" - if there are bugs we may instantiate multiple memories for each external function call.

@rodrigo-pino

rodrodros:
About default initialization I was thinking to force every implicit defined to implement a `CustomImplicit`  trait which would implement a function `default_initialize`. The plugin would initialize each implicit using `default_initialize`

piwonskp:
@rodrigo-pino but what if the initialization requires arguments? Is there a syntax for variable number of arguments in cairo 1? Aren't we overcomplicating things that can be easy?

rodrodros:
@piwonskp @JorikSchellekens I agree with both of your points. Leaving it out then since it can overcomplicate things.

piwonskp:
Ok, I'll revert the commit removing initialization

rodrodros (approved):

src/cairoWriter/writers/returnWriter.ts
```diff
@@ -6,24 +6,17 @@ import { CairoASTNodeWriter } from '../base';
...
-      : '';
-
     const finalizeKeccakPtr = this.usesImplicit(KECCAK, node)
       ? 'finalize_keccak(keccak_ptr_start, keccak_ptr);\n'
       : '';
 
-    return [
-      [documentation, finalizeWarpMemory, finalizeKeccakPtr, `return ${returns};`].join('\n'),
-    ];
+    return [[documentation, '', finalizeKeccakPtr, `return ${returns};`].join('\n')];
```
cicr99: Is it necessary to leave the blank line where `finalizeWarpMemory` used to be?

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -116,33 +117,36 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
     const argString = mapRange(size, (n) => `e${n}: ${elementCairoType.toString()}`).join(', ');
 
     // If it's dynamic we need to include the length at the start
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const alloc_len = dynamic ? size * elementCairoType.width + 2 : size * elementCairoType.width;
+    const writes = [
+      ...(dynamic ? [`wm_write_256{warp_memory=warp_memory}(start, ${uint256(size)});`] : []),
```
cicr99, line 123: Is this code still valid? I imagine this will be handled when addressing dynamic arrays, right?

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -116,33 +117,36 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
...
+        .flat()
+        .map(
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          (name, index) => `warp_memory.append(${name});`,
+          // `warp_memory.insert(
+          //   ${add(
+          //   'start',
+          //   dynamic ? index + 2 : index,
+          // )},
+          // ${name});`,
```
cicr99, line 134: What about these comments?

src/cairoWriter/writers/returnWriter.ts
```diff
@@ -6,24 +6,17 @@ import { CairoASTNodeWriter } from '../base';
...
     const documentation = getDocumentation(node.documentation, writer);
     const returns = node.vExpression ? writer.write(node.vExpression) : '()';
 
-    const finalizeWarpMemory = this.usesImplicit(WARP_MEMORY, node)
-      ? 'default_dict_finalize(warp_memory_start, warp_memory, 0);\n'
-      : '';
-
     const finalizeKeccakPtr = this.usesImplicit(KECCAK, node)
       ? 'finalize_keccak(keccak_ptr_start, keccak_ptr);\n'
       : '';
```
cicr99, line 17: Is this still necessary? I thought the initialization of these got removed from `getBody` in `cairoFunctionDefinitionWriter`

rjnrohit (approved):

src/cairoUtilFuncGen/memory/arrayLiteral.ts
piwonskp: Right

src/cairoWriter/writers/returnWriter.ts
piwonskp: Good catch

src/cairoUtilFuncGen/memory/arrayLiteral.ts
piwonskp: I'll uncomment that in memory struct PR, append is temporary solution

src/cairoWriter/writers/returnWriter.ts
piwonskp: Imo implicits deserve a separate PR so I only did the necessary changes to them in order to implement static arrays

cicr99 (approved):

PR #1005, opened 2023-03-27
Refactor on BytesConverter pass

AlejandroLabourdette:
This refactor set order over the updates that this pass do to the ast. This order will make the transition to Infertype smoother.


src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,72 @@
...
+  createUint8TypeName,
+  createUint256TypeName,
+} from '../../utils/nodeTemplates';
+import { safeGetNodeType } from '../../utils/nodeTypeProcessing';
+import { replaceBytesType } from './utils';
+
+export class ReplaceIndexAccessBytesConverter extends ASTMapper {
+  visitIndexAccess(node: IndexAccess, ast: AST): void {
+    const a = generalizeType(safeGetNodeType(node.vBaseExpression, ast.inference))[0];
+    if (!(node.vIndexExpression && a instanceof FixedBytesType)) {
```
rodrodros: Stricter condition here: `node.vIndexExpression !== undefined`

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,72 @@
...
+  createNumberLiteral,
+  createUint8TypeName,
+  createUint256TypeName,
+} from '../../utils/nodeTemplates';
+import { safeGetNodeType } from '../../utils/nodeTypeProcessing';
+import { replaceBytesType } from './utils';
+
+export class ReplaceIndexAccessBytesConverter extends ASTMapper {
+  visitIndexAccess(node: IndexAccess, ast: AST): void {
+    const a = generalizeType(safeGetNodeType(node.vBaseExpression, ast.inference))[0];
```
rodrodros: Give a meaningful name to the variable

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,72 @@
...
+      this.visitExpression(node, ast);
+      return;
+    }
+
+    const baseTypeName = typeNameFromTypeNode(
+      safeGetNodeType(node.vBaseExpression, ast.inference),
+      ast,
+    );
+
+    const width: string = baseTypeName.typeString.slice(5);
```
rodrodros: Why 5? Don't use a single number, use what your expecting to remove, like:
```ts
const width: string = baseTypeName.typeString.slice("uint".length)
```
This way is easier to understand

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
AlejandroLabourdette: Code was copy-pasted from the former approach but is a valid comment, it's fixed now

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
AlejandroLabourdette: Code was copy pasted from the former approach but is a valid comment, it's fixed now

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,72 @@
...
+      node.vIndexExpression instanceof Literal
+        ? createUint256TypeName(ast)
+        : typeNameFromTypeNode(safeGetNodeType(node.vIndexExpression, ast.inference), ast);
+
+    const stubParams: [string, TypeName][] = [
+      ['base', baseTypeName],
+      ['index', indexTypeName],
+    ];
+    const callArgs = [node.vBaseExpression, node.vIndexExpression];
+    if (baseTypeName.typeString !== 'bytes32') {
```
piwonskp: Can't we compare Nodetype instead? Why do we have to use typenames here?

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,72 @@
...
+  createNumberLiteral,
+  createUint8TypeName,
+  createUint256TypeName,
+} from '../../utils/nodeTemplates';
+import { safeGetNodeType } from '../../utils/nodeTypeProcessing';
+import { replaceBytesType } from './utils';
+
+export class ReplaceIndexAccessBytesConverter extends ASTMapper {
+  visitIndexAccess(node: IndexAccess, ast: AST): void {
+    const baseExprType = generalizeType(safeGetNodeType(node.vBaseExpression, ast.inference))[0];
```
piwonskp: `safeGetNodeType(node.vBaseExpression, ast.inference)` is used multiple times in the method so it can be extracted to a separate variable

AlejandroLabourdette: It was used 2 times with the same node, other calls were with different nodes. Changes Incoming

src/passes/bytesConverter/updateTypeStringBytesConverter.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  visitElementaryTypeName(node: ElementaryTypeName, ast: AST): void {
+    const typeNode = ast.inference.typeNameToTypeNode(node);
+    if (typeNode instanceof StringType || typeNode instanceof BytesType) {
+      ast.replaceNode(node, createArrayTypeName(createUint8TypeName(ast), ast));
+      return;
+    }
+    const replacementTypeNode = replaceBytesType(typeNode);
+    if (typeNode.pp() !== replacementTypeNode.pp()) {
+      const typeString = replacementTypeNode.pp();
```
piwonskp: Shouldn't it be something like `generateExpressionTypeString(replacementTypeNode)` instead?

src/passes/bytesConverter/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  } else if (type instanceof PointerType) {
+    return new PointerType(replaceBytesType(type.to), type.location, type.kind, type.src);
+  } else if (type instanceof TupleType) {
+    return new TupleType(type.elements.map(replaceBytesType), type.src);
+  } else if (type instanceof TypeNameType) {
+    return new TypeNameType(replaceBytesType(type.type), type.src);
+  } else if (type instanceof BytesType) {
+    return new ArrayType(new IntType(8, false, type.src), undefined, type.src);
+  } else if (type instanceof StringType) {
+    return new ArrayType(new IntType(8, false, type.src), undefined, type.src);
```
piwonskp:
```suggestion
  } else if (type instanceof BytesType || type instanceof StringType) {
    return new ArrayType(new IntType(8, false, type.src), undefined, type.src);
```

src/passes/bytesConverter/updateTypeStringBytesConverter.ts
AlejandroLabourdette: both generate same value, but `generateExpressionTypeString` keep same pattern than other functions

src/passes/bytesConverter/replaceIndexAccessBytesConverter.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const call = createCallToFunction(importedFunc, callArgs, ast);
+    ast.replaceNode(node, call, node.parent);
+    const callNodeType = replaceBytesType(safeGetNodeType(call, ast.inference));
+    call.typeString = generateExpressionTypeString(callNodeType);
+    this.commonVisit(call, ast);
+  }
+}
+
+function selectWarplibFunction(baseTypeName: TypeName, indexTypeName: TypeName): string {
```
piwonskp: Is it possible to use TypeNodes with instanceof instead of typenames and typestrings here?

piwonskp (approved):

rodrodros (approved):

PR #1022, opened 2023-04-07
remove the use of implicits that are not warp-memory

AlejandroLabourdette:


rjnrohit (changes requested):
I'd suggest you to add warp_memory implicit rather than completely getting rid of it
```rs
#[implicit(warp_memory)]
fn f(...) {
  ...
  warp_memory.read_u128(index);
  ...
}
```
It would be difficult later to track places where `warp_memory` needs to be added if we remove it from all of the places now.

AlejandroLabourdette:
> I'd suggest you to add warp_memory implicit rather than completely getting rid of it
> 
> ```rust
> #[implicit(warp_memory)]
> fn f(...) {
>   ...
>   warp_memory.read_u128(index);
>   ...
> }
> ```
> 
> It would be difficult later to track places where `warp_memory` needs to be added if we remove it from all of the places now.

I agree 100% with you, actually that's was what I think, that's why on `CairoFunctionDefinition` implementation I appended a comment of how we plan to use implicit and sort of what's happening with them.
https://github.com/NethermindEth/warp/blob/5cc3efe06dfff437b11b68929a0385b557bbbfba/src/ast/cairoNodes/cairoFunctionDefinition.ts#L18

Right now we still have a set of Implicits foreach function definition. Even considering that the only available implicit is `warplib_memory` the logic we had to infer the implicits of the current function worth it.

Nevertheless I do remove them from `cairoUtilFuncGen` scripts that were not updated to Cairo1 syntax, mainly because we had them hardcode in there (was just a string that we insert in the raw string definition of those functions). They didn't have any role at ast level and we do have to review each of of those script to make them Cairo1 compatible.
This is an example:
https://github.com/NethermindEth/warp/blob/db1b5987c7fe19c0f7b26dc80239f819be7f644e/src/cairoUtilFuncGen/memory/implicitConversion.ts#L51
https://github.com/NethermindEth/warp/blob/db1b5987c7fe19c0f7b26dc80239f819be7f644e/src/cairoUtilFuncGen/memory/implicitConversion.ts#L210

src/utils/importFuncGenerator.ts
```diff
@@ -116,14 +109,11 @@ export function createImport(
...
     case encodePath(EMIT_EVENT):
     case encodePath(GET_CALLER_ADDRESS):
     case encodePath(GET_CONTRACT_ADDRESS):
-      return createFuncImport('syscall_ptr');
-    // Import libraries from Cairo1
-    case encodePath(INTO):
+    case encodePath(INTO): // Import libraries from Cairo1
     case encodePath(ADDRESS_INTO_FELT):
     case encodePath(U128_TO_FELT):
     case encodePath(U256_FROM_FELTS):
```
piwonskp: I think we can remove the whole switch since now there are no other implicits than `warp_memory`

piwonskp:
> I'd suggest you to add warp_memory implicit rather than completely getting rid of it
> 
> ```rust
> #[implicit(warp_memory)]
> fn f(...) {
>   ...
>   warp_memory.read_u128(index);
>   ...
> }
> ```
> 
> It would be difficult later to track places where `warp_memory` needs to be added if we remove it from all of the places now.

I agree. Let's just put `#[implicit(warp_memory)]` in front of those functions, but without updating the rest of their implementation and syntax

src/cairoUtilFuncGen/enumInputCheck.ts
```diff
@@ -68,12 +68,12 @@ export class EnumInputCheck extends StringIndexedFuncGen {
       imports.push(this.requireImport(...NARROW_SAFE), this.requireImport(...U128_FROM_FELT));
     }
 
-    const implicits = '{range_check_ptr : felt}';
     const nMembers = enumDef.vMembers.length;
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: [
-        `func ${funcName}${implicits}(${
+        `#[implicit(warp_memory)]`,
```
rjnrohit: is it needed here?

rjnrohit (changes requested):
everything looks good to me, just left a comment there!

src/cairoUtilFuncGen/enumInputCheck.ts
AlejandroLabourdette: nop

rjnrohit:
Oops ! conflicts has come up since I have merged the uintN pr, could you resolve it?

rjnrohit (approved):

rodrodros (commented):
As part of this PR you should update `cairoParsing.ts` and make it parse the custom implicit attribute in order to detect when generated functions requires the `warp_memory` implicit

src/cairoUtilFuncGen/inputArgCheck/inputCheck.ts
```diff
@@ -154,7 +152,8 @@ export class InputCheckGen extends StringIndexedFuncGen {
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: [
-        `func ${funcName}${IMPLICITS}(arg : ${cairoType.toString()}) -> (){`,
+        `#[implicit(warp_memory)]`,
```
piwonskp: There wasn't warp memory in deleted `IMPLICITS`

AlejandroLabourdette: True

AlejandroLabourdette:
> As part of this PR you should update `cairoParsing.ts` and make it parse the custom implicit attribute in order to detect when generated functions requires the `warp_memory` implicit

Actually that's already done, we just have the remains of the old `parseImplicit`, but using `getRawCairoFunctionInfo` is already known the implicit used (if used any). I'm gonna remove that old function right now

src/cairoUtilFuncGen/storage/copyToStorage.ts
```diff
@@ -163,7 +160,8 @@ export class StorageToStorageGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: [
-        `func ${funcName}${IMPLICITS}(to_loc: felt, from_loc: felt) -> (retLoc: felt){`,
+        `#[implicit(warp_memory)]`,
```
piwonskp: IMPLICITS did not contain `warp_memory`

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
```diff
@@ -157,7 +153,8 @@ export class DynArrayPushWithArgGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: [
-        `func ${funcName}${implicits}(loc: felt, value: ${inputType}) -> (){`,
+        `#[implicit(warp_memory)]`,
```
piwonskp: The function had warp memory implicit added conditionally. Imo we should keep it that way. Wyt @rodrigo-pino?

src/cairoUtilFuncGen/storage/storageDelete.ts
```diff
@@ -127,7 +125,8 @@ export class StorageDeleteGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: [
-        `func ${funcName}${IMPLICITS}(loc: felt){`,
+        `#[implicit(warp_memory)]`,
```
piwonskp: IMPLICITS do not contain warp_memory

src/cairoUtilFuncGen/storage/storageToCalldata.ts
```diff
@@ -95,10 +95,10 @@ export class StorageToCalldataGen extends StringIndexedFuncGen {
       'member',
     );
 
-    const implicits = '{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}';
     const funcName = `ws_struct_${cairoStruct.toString()}_to_calldata`;
     const code = [
-      `func ${funcName}${implicits}(loc : felt) -> (${structName} : ${cairoStruct.toString()}){`,
+      `#[implicit(warp_memory)]`,
```
piwonskp: Again as above

src/cairoWriter/writers/returnWriter.ts
```diff
@@ -1,30 +1,12 @@
...
-  }
-
-  private usesImplicit(implicit: Implicits, node: Return): boolean {
-    const parentFunc = node.getClosestParentByType(CairoFunctionDefinition);
-    return (
-      parentFunc instanceof CairoFunctionDefinition &&
-      parentFunc.implicits.has(implicit) &&
-      isExternallyVisible(parentFunc)
-    );
+    return [[documentation, `return ${returns};`].join('\n')];
```
piwonskp, line 10: Since we're here, let's change it to use `endent`

src/utils/implicits.ts
```diff
@@ -1,55 +1 @@
...
-export const requiredBuiltin: { [key in Implicits]: CairoBuiltin | null } = {
-  bitwise_ptr: 'bitwise',
-  pedersen_ptr: 'pedersen',
-  range_check_ptr: 'range_check',
-  syscall_ptr: null,
-  warp_memory: null,
-  keccak_ptr: null,
-  dict_ptr: null,
-};
+export type Implicits = 'warp_memory';
```
piwonskp: One line file, where could we put that? @rodrigo-pino?

piwonskp:
Does `parseMultipleRawCairoFunctions` work properly?

src/utils/importFuncGenerator.ts
```diff
@@ -148,113 +72,30 @@ export function createImport(
...
-    case encodePath(U176_FROM_FELT):
-    case encodePath(U184_FROM_FELT):
-    case encodePath(U192_FROM_FELT):
-    case encodePath(U200_FROM_FELT):
-    case encodePath(U208_FROM_FELT):
-    case encodePath(U216_FROM_FELT):
-    case encodePath(U224_FROM_FELT):
-    case encodePath(U232_FROM_FELT):
-    case encodePath(U240_FROM_FELT):
-    case encodePath(U248_FROM_FELT):
```
piwonskp, line 198: Won't the lack of those break uintN PR functionality? Since we've got rid of switch we could also do it smarter now

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
```diff
@@ -107,25 +108,6 @@ export class CairoFunctionDefinitionWriter extends CairoASTNodeWriter {
     else return ''; // No return specified so nothing to print
   }
 
-  private getImplicits(node: CairoFunctionDefinition): string {
```
piwonskp, line 110: Was `getImplicits` never called?

piwonskp:
There is also `getImplicitsAt` in ast.ts. Is it needed?

src/cairoWriter/writers/cairoFunctionDefinitionWriter.ts
AlejandroLabourdette: Initially it was used to get the implicits subsection of a function definition header (`{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*...}`) but now we don't need it. We write now the implicits in the `getDecorator` function so that `getImplicits` was unreferenced

rodrodros (approved):
Thanks for this PR, looks good to me

src/utils/importFuncGenerator.ts
```diff
@@ -72,26 +57,89 @@ export function createImport(
...
+    Paths.U248_TO_FELT,
+    Paths.U256_FROM_FELTS,
+    Paths.ARRAY,
+    Paths.ARRAY_TRAIT,
+    Paths.WARP_MEMORY,
+    Paths.MEMORY_TRAIT,
+  ];
+  if (pathsForFunctionImport.some((i) => encodePath([path, name]) === encodePath(i))) {
     return createFuncImport();
   }
```
piwonskp, line 144: Let's merge all those ifs containing `return createFuncImport();` into one

piwonskp (approved):

PR #1025, opened 2023-04-07
Warp memory struct

piwonskp:
Solidity code:
```solidity
pragma solidity ^0.8.14;

contract C {
  function f() public {
    uint8[3] memory t = [1, 2, 3];
  }
}
```
transpiles to:
```
#[contract] 
mod C {


use array::Array;
use array::ArrayTrait;
use integer::u32_to_felt252;
use warplib::warp_memory::MemoryTrait;
use warplib::warp_memory::WarpMemory;


struct Storage {
  WARP_STORAGE: LegacyMap::<felt252, felt252>,
  WARP_USED_STORAGE: felt252,
  WARP_NAMEGEN: felt252,
  
}

fn readId(loc: felt252) -> felt252 {
  let id = WARP_STORAGE::read(loc);
  if id == 0 {
    let id = WARP_NAMEGEN::read();
    WARP_NAMEGEN::write(id + 1);
    WARP_STORAGE::write(loc, id + 1);
    return id + 1;
  } 
  return id;
}


    // Dynamic variables - Arrays and Maps

    // Static variables


    #[external]
    fn f_26121ff0(){
    let mut warp_memory: WarpMemory = MemoryTrait::initialize();

        
        let t = wm0_static_array(1, 2, 3);
        
        
        return ();
    }


    #[constructor]
    fn constructor(){

        
        
        return ();
    }

#[implicit(warp_memory)]
fn wm0_static_array(e0: felt252, e1: felt252, e2: felt252) -> felt252 {
  let start = warp_memory.pointer;
  warp_memory.insert(
              start,
              e0
            );
  warp_memory.insert(
              start + 1,
              e1
            );
  warp_memory.insert(
              start + 2,
              e2
            );
  return start;
}


}
```


warplib/warp_memory.cairo
rodrodros, line 1: Do you think we should left this primitive functions, or add more (perhaps in the form of new traits) like array allocation and static array indexing

piwonskp: I'd see much more helper functions in warplib. Ideally warplib would be static and warp would only create function calls to warplib. If that's possible of course

piwonskp: E.g. we could delete all those wm_write functions and pass parameters as arrays instead. It'd be much cleaner output then

rodrodros: We have two wm_write functions because they were the only static ones. Every other wm_write function depends on the context of the transpilation and it is dynamically generated. I think we should translate (in different PRs) all of warplib memory in different traits for `warp_memory`

piwonskp: Just to be clear, I meant that we can replace `func foo(e0: felt, e1:  felt, e2: felt)`  with generic `func foo (elements: Array::<felt>)` in warplib. It'd improve readability

rodrodros: Ok, I understand. There would be a cost on runtime then, that we should decide if it's worth it.

piwonskp: Actually it is not so obvious there will be any. Please note we're not compiling directly to binary code. Cairo compiler could optimize the code on its own

rodrodros (approved):

rjnrohit (approved):

PR #1034, opened 2023-04-17
Cross contract call

### LINKED ISSUE #964: uint/intX -> uX : cairo1

rjnrohit:
Add support for cairo1 cross contract calls.

It should closes #964 atm


src/cairoUtilFuncGen/serialisation.ts
rodrodros, line 1: There is a lot of code repetition here, is there any way to minimize this? I would think of changing the way reads work. The way it is working right now is that it generates a Cairo instruction which reads and convert to the desired type. 

Why not having serialize reads produce only read code (and not conversion). And then whatever that read call results in, add the conversion to the desired type. Just a suggestion though, it may not work

piwonskp:
Conflicts

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -134,6 +134,8 @@ export class CairoContractWriter extends CairoASTNodeWriter {
       .filter((func, index, importFuncs) => func !== importFuncs[index - 1])
       .join('\n');
 
+    // console.log(writtenImportFuncs, importFunctions);
```
piwonskp: leftover

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -158,8 +160,11 @@ export class CairoContractWriter extends CairoASTNodeWriter {
 
     const contractHeader = '#[contract] \n' + `mod ${node.name} {`;
 
+    const globalImports = ['use starknet::ContractAddress;'].join('\n');
```
piwonskp, line 161: Please use the function to print import instead of hardcoding it

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -206,14 +212,15 @@ export class CairoContractWriter extends CairoASTNodeWriter {
...
       : node.name;
 
     return [
-      [
-        documentation,
-        [`@contract_interface`, `namespace ${interfaceName}{`, ...functions, `}`].join('\n'),
-      ].join('\n'),
+      [documentation, [`#[abi]`, `trait ${interfaceName}{`, ...functions, `}`].join('\n')].join(
+        '\n',
+      ),
```
piwonskp: Please use endent

src/cairoWriter/writers/functionCallWriter.ts
```diff
@@ -55,9 +55,8 @@ export class FunctionCallWriter extends CairoASTNodeWriter {
                 nodeType.definition.name,
                 node,
                 interfaceNameMappings,
-              )}.${(isDelegateCall ? 'library_call_' : '') + memberName}(${firstArg}${
-                args ? ', ' : ''
-              }${args})`,
+                isDelegateCall,
+              )}{contract_address: ${firstArg}}.${memberName}(${args})`,
```
piwonskp: let's extract `getInterfaceNameForContract(...` to a variable

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -101,9 +112,20 @@ export class MemoryReadGen extends StringIndexedFuncGen {
...
+        isTrait: true,
+      });
+      createImport(...CONTRACT_ADDRESS, this.sourceUnit, this.ast);
+      functionsCalled.push(createImport(...CONTRACT_ADDRESS_FROM_FELT, this.sourceUnit, this.ast));
+    }
+
     const funcName = `WM${this.generatedFunctionsDef.size}_READ_${typeToRead.typeName}`;
     const resultCairoType = typeToRead.toString();
-    const [reads, pack] = serialiseReads(typeToRead, readFelt, readFelt);
+    const [reads, pack] = serialiseReads(typeToRead, readFelt, readFelt, readAddress, uNread);
```
AlejandroLabourdette: Can you use an approach similar to this one

https://github.com/NethermindEth/warp/blob/2a911fa0f38c15287f233030344228f939801832/src/cairoUtilFuncGen/serialisation.ts#L28-L31

to avoid the pass of extra arguments?

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -101,9 +112,20 @@ export class MemoryReadGen extends StringIndexedFuncGen {
   }
 
   private getOrCreate(typeToRead: CairoType): GeneratedFunctionInfo {
+    const functionsCalled: FunctionDefinition[] = [this.requireImport(...DICT_READ)];
+
+    // register require imports
```
AlejandroLabourdette: The `serializeRead` return this now

https://github.com/NethermindEth/warp/blob/2a911fa0f38c15287f233030344228f939801832/src/cairoUtilFuncGen/serialisation.ts#L21

so you can handle this import logic inside it

AlejandroLabourdette: also for `storageRead.ts` file

src/cairoUtilFuncGen/memory/memoryRead.ts
AlejandroLabourdette: also for `storageRead.ts` file

src/cairoUtilFuncGen/serialisation.ts
```diff
@@ -13,19 +14,119 @@ export function serialiseReads(
...
+          reads.push(readFelt(reads.length));
+          return `read${reads.length - 1}`;
+        case Read.Id:
+          reads.push(readId(reads.length));
+          return `read${reads.length - 1}`;
+        case Read.Address:
+          reads.push(readAddress(reads.length));
+          return `read${reads.length - 1}`;
+        case Read.U8:
+          reads.push(uNread(reads.length, 8));
```
AlejandroLabourdette: I think all those Read.UN can be written as one if we put that `N` value as type.nBits. Somethin like 
```
    case Read.UN:
        reads.push(uNread(reads.length, type.nBits));
        return `read${reads.length - 1}`;
```

perhaps would be needed an `assert(type instance of CairoUint)` at case begin

src/cairoWriter/writers/variableDeclarationStatementWriter.ts
```diff
@@ -71,12 +71,16 @@ export class VariableDeclarationStatementWriter extends CairoASTNodeWriter {
       return [
         [
           documentation,
-          `let (${declarations.join(', ')}) = ${writer.write(node.vInitialValue)};`,
+          `let (${declarations.map((decl) => `mut ${decl}`).join(', ')}) = ${writer.write(
```
AlejandroLabourdette: Why is it that we need to declare them as mutable?

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
```diff
@@ -138,5 +145,38 @@ export function genContractInterface(
     });
   sourceUnit.appendChild(contractInterface);
   ast.registerChild(contractInterface, sourceUnit);
+  createImport(
```
AlejandroLabourdette, line 149: Wouldn't be better if we use a call to this 
https://github.com/NethermindEth/warp/blob/2a911fa0f38c15287f233030344228f939801832/src/cairoUtilFuncGen/base.ts#L51

and hardcode the `['super']` as a constant in the script where we have all import constants defined?

src/cairoWriter/writers/variableDeclarationStatementWriter.ts
rjnrohit: Because at later stages we modify the value of the variables, So I thought it's good to make the variables mutable

src/passes/externalContractHandler/externalContractInterfaceInserter.ts
rjnrohit: well we can't use `this` here beacuse the class is of ASTMapper type, for `super` I can make a constant , good point

src/cairoWriter/writers/cairoContractWriter.ts
rjnrohit: I'm working on it! Writing redundant struct imports on file level

src/cairoUtilFuncGen/serialisation.ts
rjnrohit: Screenshot 2023-04-25 at 6 36 47 PM
Really? It doesn't seem to work

AlejandroLabourdette: Ye, you have to define it first in the enum were you defined all other Ux

https://github.com/NethermindEth/warp/blob/6ee56532862c1e21d18fe1553731be0e3c1b9678/src/cairoUtilFuncGen/serialisation.ts#L136-L144

rjnrohit: got it

src/cairoUtilFuncGen/memory/memoryRead.ts
rjnrohit: But for memory and storage there would be different line of snippet.

AlejandroLabourdette: The definition of the readFelt function can change but the Idea of reusing the returned value to convert later into the type you want (address, uN, bool...) is the same. Just to keep in mind that the name of the variable should be the same one.

src/cairoWriter/writers/cairoContractWriter.ts
rjnrohit: will handle when I address #1042

src/icf/genCairo.ts
```diff
@@ -55,7 +55,7 @@ function getStarknetLangDirective(): string[] {
 
 function getForwarderInterface(abi: AbiType): string[] {
   return [
-    '@contract_interface',
+    '#[abi]',
     'namespace Forwarder {',
```
piwonskp: Why is the change needed? The code still won't work since the rest of the code is cairo 0, right? Or am I missing something? Is there a `namespace` in cairo 1?

src/cairoWriter/writers/variableDeclarationStatementWriter.ts
```diff
@@ -71,12 +71,16 @@ export class VariableDeclarationStatementWriter extends CairoASTNodeWriter {
...
+            node.vInitialValue,
+          )};`,
         ].join('\n'),
       ];
     }
     return [
-      [documentation, `let ${declarations[0]} = ${writer.write(node.vInitialValue)};`].join('\n'),
+      [documentation, `let mut ${declarations[0]} = ${writer.write(node.vInitialValue)};`].join(
+        '\n',
+      ),
```
piwonskp: Please use `endent` instead of `join`

src/icf/genCairo.ts
rjnrohit: it must got changed because of the global regex replace, yes it doesn't work now

src/passes/functionPruner/functionRemover.ts
```diff
@@ -32,6 +40,13 @@ export class FunctionRemover extends ASTMapper {
     // Remove unreachable functions
     node.vFunctions
       .filter((func) => !this.reachableFunctions.has(func.id))
+      .filter(
+        (func) =>
+          !(
+            func instanceof CairoImportFunctionDefinition &&
+            func.functionStubKind === FunctionStubKind.TraitStructDefStub
+          ),
```
piwonskp: Let's extract it to a named function

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -206,13 +210,15 @@ export class CairoContractWriter extends CairoASTNodeWriter {
...
+          false,
         )
       : node.name;
 
     return [
       [
         documentation,
-        [`@contract_interface`, `namespace ${interfaceName}{`, ...functions, `}`].join('\n'),
+        [endent`#[abi]`, endent`trait ${interfaceName}{`, ...functions, `}`].join('\n'),
       ].join('\n'),
```
piwonskp: The joins are redundant

src/cairoUtilFuncGen/storage/storageRead.ts
```diff
@@ -61,14 +61,14 @@ export class StorageReadGen extends StringIndexedFuncGen {
 
   private getOrCreate(typeToRead: CairoType): GeneratedFunctionInfo {
     const functionsCalled: FunctionDefinition[] = [];
+    this.requireImport;
```
AlejandroLabourdette: left over?

src/cairoUtilFuncGen/serialisation.ts
```diff
@@ -9,24 +10,44 @@ import {
...
+] {
   const packExpression = producePackExpression(type);
   const reads: string[] = [];
-  const requiredImports: [string[], string][] = [];
+  const requiredImports: { import: [string[], string]; isTrait?: boolean }[] = [];
   const packString: string = packExpression
-    .map((elem: string | Read) => {
-      if (elem === Read.Felt) {
+    .map((elem: string | Read | (string | Read)[]) => {
+      if (elem instanceof Array) {
```
AlejandroLabourdette: What do you think if each of the elements of the returned packExpression is standarized in the way of `(elementType, dataRelated)` being `elementType` the enum value from `Read` and `dataRelated` an object were you store data needed to write correctly the reader for that type.

For address could be something like `(Read.Address, {})` but for uN could be like `(Read.UN, {nBytes: 8})`.

That way we could do the conditional match of the elem in a more elegant way and avoid the assumption that if `elem` is an array, then it will always belong to an UN

src/passes/functionPruner/functionRemover.ts
rjnrohit: I didn't get it, do you mean to split 46, 47 in two filter functions?

rjnrohit: done!

src/passes/functionPruner/functionRemover.ts
```diff
@@ -15,24 +16,31 @@ export class FunctionRemover extends ASTMapper {
...
+
   visitSourceUnit(node: SourceUnit, ast: AST): void {
     node.vFunctions.filter((func) => isExternallyVisible(func)).forEach((func) => this.dfs(func));
 
     node.vContracts.forEach((c) => this.visitContractDefinition(c, ast));
 
-    node.vFunctions
-      .filter((func) => !this.reachableFunctions.has(func.id))
-      .forEach((func) => node.removeChild(func));
+    this.importFuncFilter(node.vFunctions).forEach((func) => node.removeChild(func));
```
piwonskp, line 35: `visitContractDefinition` and `visitSourceUnit` are very similiar. Would that break any logic if we reordered instructions to reuse more code? I mean changing `visitSourceUnit` to: 
```
this.removeUnreachableFunctions();
node.vContracts.forEach((c) => this.visitContractDefinition(c, ast));
```
Or to:
```
node.vContracts.forEach((c) => this.visitContractDefinition(c, ast));
this.removeUnreachableFunctions();
```

rjnrohit: I guess, it'd make the difference in the dfs graph visitor, what if a function inside anyone of the contracts makes a call to the source unit level function. @cicr99 your opinion?

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -206,13 +210,18 @@ export class CairoContractWriter extends CairoASTNodeWriter {
...
 
     return [
       [
         documentation,
-        [`@contract_interface`, `namespace ${interfaceName}{`, ...functions, `}`].join('\n'),
+        endent`#[abi]
+        trait ${interfaceName}{
+          ${functions.join('\n')}
+          }`,
       ].join('\n'),
```
piwonskp: Unnecessary join
```
endent`
${documentation}
#[abi]
trait ${interfaceName}{
  ${functions.join('\n')}
}`
```

rjnrohit: Thanks for pointing it out 🙌, updated !

src/passes/functionPruner/functionRemover.ts
cicr99: Source Unit functions are supposed to be file-level function definitions (free functions). I believe, currently those function get inline in the contracts, but I'm not sure if they are removed then. Anyway, there is still the question whether to remove them or not given that you can use them in Cairo 1, so I wouldn't remove this code yet.
I agree with @rjnrohit that we can not separate the order of operations that way: the dfs has to start with the free functions, but this will only visit the called functions inside their bodies, so it is still necessary to check if there is any function inside the contracts that is using those free functions, hence, the contracts in the source unit are visited. After that, is when the functions can be removed.

rjnrohit: Thanks @cicr99 for clarification!

src/cairoUtilFuncGen/serialisation.ts
```diff
@@ -41,13 +62,23 @@ export function serialiseReads(
...
+  metadata: {
+    nBits?: number;
+  };
+}
+
+function producePackExpression(type: CairoType): packExpReturnType[] {
+  if (type instanceof WarpLocation) return [{ dataOrDataType: Read.Id, metadata: {} }];
+  if (type instanceof CairoFelt) return [{ dataOrDataType: Read.Felt, metadata: {} }];
+  if (type instanceof CairoContractAddress) return [{ dataOrDataType: Read.Address, metadata: {} }];
+  if (type instanceof CairoBool) return [{ dataOrDataType: Read.Bool, metadata: {} }];
```
piwonskp: Imo it'd be more readable to create a function wrapping `Read.<type>` within the `packExpReturnType[]` instead of repeating it in each lane

piwonskp: Similarly the code below

piwonskp (approved):

rodrodros (approved):

Issue #1026, opened 2023-04-07
dynamic arrays 


piwonskp:
The code should compile:
```solidity
pragma solidity ^0.8.6;

//SPDX-License-Identifier: MIT

contract WARP {
    function uint8new() pure public {
        uint8[] memory x = new uint8[](2);
    }
}
```

PR #1035, opened 2023-04-17
Storage Dynamic Arrays

AlejandroLabourdette:
Issue: #1026.


src/cairoUtilFuncGen/storage/dynArrayIndexAccess.ts
```diff
@@ -71,29 +71,22 @@ export class DynArrayIndexAccessGen extends StringIndexedFuncGen {
...
-        `}`,
-      ].join('\n'),
-      functionsCalled: [
-        this.requireImport(...U128_FROM_FELT),
-        this.requireImport(...UINT256_LT),
-        arrayDef,
-        arrayLength,
-      ],
+      code: endent`
+        fn ${funcName}(ref: felt252, index: u256) -> felt252 {
```
rodrodros: `ref` is a keyword in cairo1, it cannot/shoudln't be used as  a parameter

src/cairoUtilFuncGen/storage/dynArrayPop.ts
```diff
@@ -88,31 +89,27 @@ export class DynArrayPopGen extends StringIndexedFuncGen {
...
       isDynamicArray(elementType) || isMapping(elementType)
-        ? [
-            `let (elem_loc) = ${arrayName}.read(loc, newLen);`,
-            `let (elem_loc) = readId(elem_loc);`,
-          ].join('\n')
-        : `let (elem_loc) = ${arrayName}.read(loc, newLen);`;
+        ? endent`
+            let elem_loc = ${arrayName}::read((loc, newLen));
+            let elem_loc = readId(elem_loc);
+            `
```
piwonskp, line 95: Does double definition of the same variable work in Cairo 1?

AlejandroLabourdette: yes, you can take this as an example:
```
fn main() -> felt252 {
    let a = 3;
    let a = a + 1;
    a
}
```
output:
```
Run completed successfully, returning [4]
```

piwonskp: Even though it's possible, let's do:
```suggestion
            let elem_loc = ${arrayName}::read((loc, newLen));
            elem_loc = readId(elem_loc);
            `
```

AlejandroLabourdette: In order to do that we would need to specify variable as mutable. I think, if it cause so much confusion, that the best idea would be to just rename it

src/cairoUtilFuncGen/storage/dynArrayPop.ts
```diff
@@ -90,8 +90,8 @@ export class DynArrayPopGen extends StringIndexedFuncGen {
     const getElemLoc =
       isDynamicArray(elementType) || isMapping(elementType)
         ? endent`
-            let elem_loc = ${arrayName}::read((loc, newLen));
-            let elem_loc = readId(elem_loc);
+            let elem_loc_id = ${arrayName}::read((loc, newLen));
+            let elem_loc = readId(elem_loc_id);
```
rodrodros, line 94: nitpick, but I would go for:
* elem_id_loc  for the first one, since is the location of the id or pointer
* elem_id for the second one, since it holds the actual id

src/cairoUtilFuncGen/storage/dynArrayPop.ts
piwonskp: Oh you are probably right

src/cairoUtilFuncGen/storage/dynArrayIndexAccess.ts
```diff
@@ -71,29 +71,22 @@ export class DynArrayIndexAccessGen extends StringIndexedFuncGen {
...
+            let length = ${lengthName}::read(warp_storage_array_ptr);
+            assert(index < length, 'Index out of bounds');
+            let existing = ${arrayName}::read(warp_storage_array_ptr, index);
+            if (existing == 0) {
+                let used = WARP_USED_STORAGE::read();
+                WARP_USED_STORAGE::write(used + ${valueCairoType.width});
+                ${arrayName}::write((warp_storage_array_ptr, index), used);
+                return used;
+            } else {
+                return existing;
```
rodrodros: Since Cairo is very rust-like, and Cairo 1 devs are following the same guidelines that Rust follows, I think that our generated code should match it. You should change the `return expr;` for just `expr`

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
```diff
@@ -144,40 +145,34 @@ export class DynArrayPushWithArgGen extends StringIndexedFuncGen {
     const arrayName = dynArray.name;
     const lengthName = dynArrayLength.name;
     const funcName = `${arrayName}_PUSHV${this.generatedFunctionsDef.size}`;
-    const implicits =
-      argLoc === DataLocation.Memory
-        ? '{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt, warp_memory: DictAccess*}'
-        : '{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt, bitwise_ptr: BitwiseBuiltin*}';
+    const implicit = argLoc === DataLocation.Memory ? '#[implicit(warp_memory)]' : '';
```
rodrodros: Update the implicit syntax

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
```diff
@@ -144,40 +145,34 @@ export class DynArrayPushWithArgGen extends StringIndexedFuncGen {
...
-        `    return ();`,
-        `}`,
-      ].join('\n'),
+      code: endent`
+        ${implicit}
+        fn ${funcName}(loc: felt252, value: ${inputType}) {
+            let len = ${lengthName}::read(loc);
+            ${lengthName}::write(loc, len + u256_from_felts(1,0));
+            let existing = ${arrayName}::read((loc, len));
+            if (existing == 0) {
```
rodrodros: redundant `()` in `if`

src/cairoUtilFuncGen/storage/dynArrayPushWithoutArg.ts
```diff
@@ -71,30 +72,22 @@ export class DynArrayPushWithoutArgGen extends StringIndexedFuncGen {
...
-        this.requireImport(...UINT256_ADD),
-        dynArray,
-        dynArrayLength,
-      ],
+      code: endent`
+        fn ${funcName}(loc: felt252) -> felt252 {
+            let len = ${lengthName}::read(loc);
+            ${lengthName}::write(loc, len + u256_from_felts(1,0));
+            let existing = ${arrayName}::read((loc, len));
+            if (existing == 0) {
```
rodrodros: redundant `()`

src/cairoUtilFuncGen/storage/dynArrayPushWithoutArg.ts
```diff
@@ -71,30 +72,22 @@ export class DynArrayPushWithoutArgGen extends StringIndexedFuncGen {
...
+                let used = WARP_USED_STORAGE::read();
+                WARP_USED_STORAGE::write(used + ${cairoElementType.width});
+                ${arrayName}::write((loc, len), used);
+                return used;
+            } else {
+                return existing;
```
rodrodros: Avoid using `return` if unnecesary

rodrodros (commented):
It looks good to me. Minor comments and it's good to merge. Could you please remove the changes you did to warplib except for integers. They are not very important for this PR and it will break other branches working with memory.,

src/cairoWriter/writers/cairoContractWriter.ts
```diff
@@ -85,7 +85,7 @@ export class CairoContractWriter extends CairoASTNodeWriter {
         WARP_STORAGE: LegacyMap::<felt252, felt252>,
         WARP_USED_STORAGE: felt252,
         WARP_NAMEGEN: felt252,
-        ${otherStorageVars.map((v) => `${writer.write(v)}`).join('\n')}
+        ${otherStorageVars.map((v) => `${writer.write(v)},`).join('\n')}
```
piwonskp, line 88: Why? Won't that break when there is more than two otherStorageVars?

src/cairoUtilFuncGen/storage/storageDelete.ts
```diff
@@ -126,12 +127,13 @@ export class StorageDeleteGen extends StringIndexedFuncGen {
...
-        `func ${funcName}${IMPLICITS}(loc: felt){`,
-        ...mapRange(cairoType.width, (n) => `    WARP_STORAGE.write(${add('loc', n)}, 0);`),
-        `    return ();`,
-        `}`,
-      ].join('\n'),
+      code: endent`
+        fn ${funcName}(loc: felt252){
+            ${mapRange(cairoType.width, (n) => `WARP_STORAGE::write(${add('loc', n)}, 0);`).join(
+              '\n',
+            )}
```
piwonskp: Let's do that in a variable and then just do `${var}`. It will be more readable

piwonskp:
Conflicts

src/cairoWriter/writers/cairoContractWriter.ts
AlejandroLabourdette: No, It actually broke the former way with more than 2 vars because it will write something like this
```
    struct Storage {
    WARP_STORAGE: LegacyMap::<felt252, felt252>,
    WARP_USED_STORAGE: felt252,
    WARP_NAMEGEN: felt252,
    a: S
    b: S
```
With one it work just fine because that last comma is kind of optional but with more than 2 there's no separator.
This will be written with the change:
```
    struct Storage {
    WARP_STORAGE: LegacyMap::<felt252, felt252>,
    WARP_USED_STORAGE: felt252,
    WARP_NAMEGEN: felt252,
    a: S,
    b: S,
```

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
AlejandroLabourdette: In Cairo source code they actually use it both ways indistinctly

https://github.com/starkware-libs/cairo/blob/74dbf871c1125a591591519b75f987df1fb2d08d/corelib/src/ecdsa.cairo#L87

https://github.com/starkware-libs/cairo/blob/74dbf871c1125a591591519b75f987df1fb2d08d/examples/fib_loop.cairo#L4

but I will include the changes

src/cairoUtilFuncGen/storage/storageDelete.ts
AlejandroLabourdette: These are instructions that are void function calls to update the storage, In this case to set it to 0 deleting all data written there. I don't understand the save in variable approach you suggest

src/cairoWriter/writers/cairoContractWriter.ts
piwonskp: My bad, I've read the diff the reverse way, thought you were deleting the comma

src/cairoUtilFuncGen/storage/storageDelete.ts
piwonskp: I've just meant to do:
```
let var = mapRange(...);
return [endent`
fn [../] {
 ${var}
`
```
a lot of nested brackets and backquotes make it hard to read

AlejandroLabourdette:
> t looks good to me. Minor comments and it's good to merge. Could you please remove the changes you did to warplib except for integers. They are not very important for this PR and it will break other branches working with memory.,

Actually they are, we need to have those `memory` warplib functions working to be able to handle storage dynamic arrays for example. Initial issue #1026 was just to be able to create memory dynamic arrays, but it wouldn't do much if we can't access or update them. Also seems logic to include support for storage dynamic arrays in this pr.
All old Cairo0 code from `memory.cairo` was added to `memory.old.cairo` file and in the `memory.cairo` were left just the Cairo1 equivalent of the used util functions. The end will be to translate utils functions as they are needed or perhaps a separate issue to translate all at once (I prefer first approach because they are wide thematic and would be hard to test) and then delete that `memory.old.cairo` file

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
AlejandroLabourdette: Can you please specify the new decorator syntax for warp-memory implicit?

src/cairoUtilFuncGen/storage/storageDelete.ts
AlejandroLabourdette: Oh ok, I thought you mean at cairo level. Changes incoming

src/cairoUtilFuncGen/storage/dynArrayPushWithArg.ts
piwonskp: Rodrigo meant to add type annotation I guess

src/cairoUtilFuncGen/storage/storageDelete.ts
```diff
@@ -123,16 +123,17 @@ export class StorageDeleteGen extends StringIndexedFuncGen {
...
-        func ${funcName}(loc: felt){
-            ${mapRange(cairoType.width, (n) => `    WARP_STORAGE.write(${add('loc', n)}, 0);`).join(
-              '\n',
-            )}
-            return ();
+        fn ${funcName}(loc: felt252){
+            ${setToZeroInstructions}
         }
-      `,
+        `,
```
piwonskp: The change on the line seems redundant

warplib/src/warp_memory.cairo
```diff
@@ -28,10 +29,14 @@ impl WarpMemoryImpl of MemoryTrait {
 
     fn insert(ref self: WarpMemory, position: felt252, value: felt252) {
         self.memory.insert(position, value);
-        self.pointer += 1;
```
piwonskp, line 31: The change breaks the logic in arrayLiteral.ts. Please consider the insert call there

AlejandroLabourdette: Is this still an issue @piwonskp after what we discuss in daily?

piwonskp: Not anymore, now we just should use `create` for static arrays

PR #1036, opened 2023-04-19
Plugin Integration

rodrodros:
This PR accomplish the following tasks:

1. Add the `warp` binary. A an "enhanced" version of Scarb which features the parsing of custom implicits.
2. Since `warp` is based on Scarb, warplib and the way files were transpiled were adapted to work the same way as Scarb
3. `warp` uses Alpha7 release so this PR serves as an update from Alpha6 as well.

Example of an annotated cairo contract:
```rs
﻿#[contract]
mod ContractUsesImplicit {
    use warplib::memory::WarpMemory;
    use warplib::memory::WarpMemoryTrait;
    use warplib::memory::WarpMemoryImpl;

	#[implicit(warp_memory: WarpMemory)]
	fn call_to_insert(value: felt252) {
    	insert_to_warp_memory(value);
    }

	#[implicit(warp_memory: WarpMemory)]
	fn insert_to_warp_memory(value: felt252) {
		warp_memory.append(value);
	}
}
```

Result after expanding the `implict` attribute:
```rs
#[contract]
mod ContractUsesImplicit {
    use warplib::memory::WarpMemory;
    use warplib::memory::WarpMemoryTrait;
    use warplib::memory::WarpMemoryImpl;

    fn call_to_insert(ref warp_memory: WarpMemory, value: felt252) {
        insert_to_warp_memory(ref warp_memory, value);
    }
    fn insert_to_warp_memory(ref warp_memory: WarpMemory, value: felt252) {
        warp_memory.append(value);
    }
}

```


tests/testing.ts
```diff
@@ -68,12 +68,15 @@ export function runSolFileTest(
...
-      ([file, cairo]) => {
-        outputFileSync(path.join(warpTest, file), cairo);
-        createCairoProject(path.join(warpTest, file));
-        const baseDir = path.dirname(path.join(warpTest, file));
+      ([fileName, cairoCode]) => {
+        const testLocation = path.join(warpTest, fileName);
+        outputFileSync(testLocation, cairoCode);
+        createCairoProject(testLocation);
+
+        const baseDir = path.dirname(path.dirname(testLocation));
```
piwonskp, line 79: Wouldn't it be better to call `path.dirname(warpTest)` here?

src/warplib/utils.ts
```diff
@@ -61,7 +61,7 @@ export function msbAndNext(width: number): string {
 // const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', '..', 'warp_venv', 'bin')}:$PATH`;
 
 export function generateFile(warpFunc: WarplibFunctionInfo): void {
-  const pathToFile = path.join('.', 'warplib', 'maths', `${warpFunc.fileName}.cairo`);
+  const pathToFile = path.join('.', 'warplib', 'src', 'maths', `${warpFunc.fileName}.cairo`);
```
piwonskp: Let's keep the path to warplib as const and reuse it

src/starknetCli.ts
```diff
@@ -29,14 +29,7 @@ const NETWORK = 'network';
...
 const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', 'warp_venv', 'bin')}:$PATH`;
-const CAIRO1_COMPILE_BIN = path.resolve(
-  __dirname,
-  '..',
-  'cairo1',
-  getPlatform(),
-  'bin',
-  'starknet-compile',
-);
+const CAIRO1_COMPILE_BIN = path.resolve(__dirname, '..', 'cairo1', getPlatform(), 'bin', 'warp');
```
piwonskp: We could have BASE_PATH to warp directory as  const and then just do `WARPLIB_PATH = path.join(BASE_PATH, 'warplib', 'src')`, same with venv and compile bin

src/warplib/gatherWarplibImports.ts
```diff
@@ -8,10 +8,12 @@ export const warplibImportInfo = glob
   .reduce((warplibMap, pathToFile) => {
     const rawCairoCode = fs.readFileSync(pathToFile, { encoding: 'utf8' });
 
-    const importPath = pathToFile
+    let importPath = pathToFile
       .split('/')
```
piwonskp:
```suggestion
      .split(path.sep)
```

piwonskp: It requires an import tho

src/warplib/gatherWarplibImports.ts
```diff
@@ -8,10 +8,12 @@ export const warplibImportInfo = glob
...
-      .join('.')
-      .slice(0, pathToFile.length - '.cairo'.length);
+      // We need to remove the `src` folder where scarb store the contracts
+      .filter((dir) => dir !== 'src')
+      .join('.');
+    importPath = importPath.slice(0, importPath.length - '.cairo'.length);
```
piwonskp: Regex seems like a better option. Something like `^warplib/src/\w+.cairo$` to match everything in between, then just use `.replace(path.sep, '.')`

tests/testing.ts
piwonskp: Or even better have BASE_DIR as const and import it

src/passes/sourceUnitSplitter.ts
```diff
@@ -109,12 +109,8 @@ export function mangleFreeFilePath(path: string): string {
...
 }
 
-export function mangleContractFilePath(
-  path: string,
-  contractName: string,
-  suffix = '.cairo',
-): string {
-  return join(path, contractName + suffix);
+export function mangleContractFilePath(path: string, contractName: string, suffix: string): string {
+  return join(path, 'src', contractName + suffix);
```
piwonskp: `extension` would be more explicit name than `suffix`

src/cli.ts
```diff
@@ -144,15 +144,21 @@ function runTranspile(files: string[], options: CliOptions) {
 }
 
 export function createCairoProject(filePath: string): void {
-  // create cairo_project.toml
-  const cairoProjectPath = path.join(path.dirname(filePath), 'cairo_project.toml');
-  const warplibRoot = path.join(PROJECT_ROOT, 'warplib');
+  const dirname = path.dirname(path.dirname(filePath));
```
piwonskp: Maybe `outputRoot` or `outputProjectRoot`

tests/testing.ts
rodrodros: I don't think it will work because: `warpTest = "warpTest"`. There are two dirname applications on `testLocation` because it's value is _warpTest/contract.sol/src/c.cairo_  and we want the path without the contract name and the src folder

src/warplib/gatherWarplibImports.ts
rodrodros: I did not fully understand what you meant here, specially how this regex would remove the `src` from the path. I've put a solution using regex nonetheless

piwonskp: I meant regex with match method. You could extract everything between `src/` and `.cairo`, then do `warplib.${variable.replace(path.sep, '.')}`

rodrodros: I see now. I am not sure I see it as a better solution than the current one. Using a match would also require to handle a possible `null` case.

piwonskp: I see, then let's use slice: `importPath.slice('warplib/src/'.length,  -'.cairo'.length)`. Then use `warplib.${importPath.replace(path.sep, '.')}`

src/warplib/gatherWarplibImports.ts
```diff
@@ -4,14 +4,11 @@ import { parseMultipleRawCairoFunctions } from '../utils/cairoParsing';
...
-  .sync('warplib/**/*.cairo')
+  .sync('warplib/src/**/*.cairo')
   .reduce((warplibMap, pathToFile) => {
     const rawCairoCode = fs.readFileSync(pathToFile, { encoding: 'utf8' });
 
-    const importPath = pathToFile
-      .split('/')
-      .join('.')
-      .slice(0, pathToFile.length - '.cairo'.length);
+    const importPath = 'warplib/' + pathToFile.slice('warplib/src/'.length, -'.cairo'.length);
```
piwonskp: What has changed that we don't use `.` as separator here anymore?

rodrodros: I change the separator to /

cicr99: Why is the `src` part not needed in the import path?

piwonskp: @cicr99 It's because it's not the system path, it's import path separated by slashes

piwonskp: And since it's an import path it would be better to write it that way:
```suggestion
    const importPath = encodePath(['warplib', pathToFile.slice('warplib/src/'.length, -'.cairo'.length).split(path.sep)]);
```

Though paths mapping/encoding functionality will require a refactor in future

rodrodros: I agree. I would like to postpone the addition of  encodePath to a future PR because it is not that straightforward, due to circular dependency between importFuncGenerator and gatherWarplibImports. Moving the `encodePath` function to gatherWarplib would solve the issue, but I think it is alright where it is located.

I think the best solution is a small movement of files where we joined all three major files: imporFuncGenerator, importPaths and gatherWarplibImports, to their own import folder, inside utils.

Tried doing it inside of this PR but there are lot of updating to do across the whole project

cicr99 (approved):
lgtm

src/cli.ts
```diff
@@ -144,15 +143,21 @@ function runTranspile(files: string[], options: CliOptions) {
...
     endent`
-      [crate_roots]
-      root = "."
-      warplib = "${warplibRoot}"
+    [package]
+    name = "${packageName}"
+    version = "1.0.0"
+
+    [dependencies]
+    warplib = { path = "${warplib}" }
```
piwonskp, line 158: Could we use `PATH_TO_WARPLIB` here as well?

src/starknetCli.ts
```diff
@@ -28,15 +28,9 @@ const MAX_FEE = 'max_fee';
...
-const CAIRO1_COMPILE_BIN = path.resolve(
-  __dirname,
-  '..',
-  'cairo1',
-  getPlatform(),
-  'bin',
-  'starknet-compile',
-);
+export const BASE_PATH = path.dirname(__dirname);
+const warpVenvPrefix = `PATH=${path.resolve(BASE_PATH, 'warp_venv', 'bin')}:$PATH`;
```
piwonskp: Is there any specific reason why we call `resolve` here instead of `join`?

src/cli.ts
rodrodros: They are not the same path. The one in `util` is for when you execute `yarn warplib` which you do from the root. The second one is done inside the bin folder (`bin/warp`). So you need to resolve first to get to the root.

src/warplib/gatherWarplibImports.ts
```diff
@@ -2,16 +2,17 @@ import fs from 'fs';
...
     const rawCairoCode = fs.readFileSync(pathToFile, { encoding: 'utf8' });
 
-    const importPath = pathToFile
-      .split('/')
-      .join('.')
-      .slice(0, pathToFile.length - '.cairo'.length);
+    const importPath = [
+      'warplib',
+      ...pathToFile.slice('warplib/src/'.length, -'.cairo'.length).split(path.sep),
+    ].join('/');
```
piwonskp, line 17: If we can't call `encodePath` here now let's add `TODO` comment

src/starknetCli.ts
rodrodros: I think it is because resolve gives the absolute path no matter the arguments

piwonskp: Is BASE_PATH relative path then?  if not it can be `join` as well

rodrodros: Nop, used `join`

src/cli.ts
piwonskp: Can't we use absolute paths? Relative paths are always error prone

rodrodros: It is an absolute path

src/starknetCli.ts
```diff
@@ -28,15 +28,9 @@ const MAX_FEE = 'max_fee';
...
-  __dirname,
-  '..',
-  'cairo1',
-  getPlatform(),
-  'bin',
-  'starknet-compile',
-);
+export const BASE_PATH = path.dirname(__dirname);
+const warpVenvPrefix = `PATH=${path.join(BASE_PATH, 'warp_venv', 'bin')}:$PATH`;
+const CAIRO1_COMPILE_BIN = path.resolve(BASE_PATH, 'cairo1', getPlatform(), 'bin', 'warp');
```
piwonskp: Same as above

piwonskp (approved):

Issue #1019, opened 2023-04-05
Implement retrieving data from warp memory


piwonskp:
The code below, copied from memory/staticArrays.sol in behaviour tests should be working properly:
```pragma solidity ^0.8.6;

//SPDX-License-Identifier: MIT

contract WARP {
    function uint8default() pure public returns (uint8, uint8) {
        uint8[2] memory x;
        return (x[0], x[1]);
    }

    function uint8write(uint8 v) pure public returns (uint8, uint8) {
        uint8[2] memory x;
        x[1] = v;
        return (x[0], x[1]);
    }

    function uint256default() pure public returns (uint256, uint256) {
        uint256[2] memory x;
        return (x[0], x[1]);
    }

    function uint256write(uint256 v) pure public returns (uint256, uint256) {
        uint256[2] memory x;
        x[1] = v;
        return (x[0], x[1]);
    }

    function assignToTuple(uint8 a, uint8 b) pure public returns (uint8, uint8) {
        uint8[2] memory arr = [a, b];
        return (arr[0], arr[1]);
    }
}
```

PR #1038, opened 2023-04-25
Warp Memory basic functionality

### LINKED ISSUE #988: Felt as a Cairo Custom Type
Cairo 1 introduces `uX`, an unsigned integer equivalent to each of Soiditys `uintX`.  It is expected from the Warp of the future to take advantage of this new types due to all the advantages they offer compared to simple `felt`s. Warp 3.0 should transpile `uintX` types as `uX`. 

The problem arise when there is need of the Felt type, Warp 3.0 is left without a mechanism to write simple felts. Examples where felts are needed:
* storage locations are felts
* memory locations are felts
* string_hash result is a felt

rodrodros:
This PR introduces Warp memory basic functionality plus other utilities

Fixes #988 #1019


warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
...
+        let length_256 = u256_from_felt252(length);
+        if index_256 >= length_256 {
+             panic_with_felt252('Index out of range');
+        } 
+
+        let index_location = array_ptr + index * elem_width;
+        self.read(index_location) 
+    }
+
+    fn lenght_dyn(ref self: WarpMemory, array_ptr: felt252) -> felt252 {
```
piwonskp:
```suggestion
    fn length_dyn(ref self: WarpMemory, array_ptr: felt252) -> felt252 {
```

warplib/src/warp_memory/write.cairo
```diff
@@ -0,0 +1,12 @@
...
+
+trait WarpMemoryWriteTrait {
+    fn write_felt(ref self: WarpMemory, index: felt252, value: felt252);
+}
+
+impl WarpMemoryWriteImpl of WarpMemoryWriteTrait {
+    fn write_felt(ref self: WarpMemory, index: felt252 , value: felt252) {
+        self.write(index, value)
+    }
+}
```
piwonskp: As we talked I think write and read traits should be removed, since we've got MemoryAccessor which is able to store/retrieve any kind of data

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -130,7 +131,8 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
       ...mapRange(size, (n) => elementCairoType.serialiseMembers(`e${n}`))
         .flat()
         .map(
-          (name, index) => `warp_memory.insert(
+          // TODO: use create from WarpMemoryAccessor instead
+          (name, index) => `warp_memory.insert(start,
```
piwonskp:
```suggestion
          (name, index) => `warp_memory.write(
```

warplib/src/warp_memory/bytes.cairo
```diff
@@ -0,0 +1,6 @@
+use warplib::warp_memory::WarpMemory;
+use warplib::warp_memory::WarpMemoryTrait;
+
+trait WarpMemoryBytesTrait{
+    fn bytes_to_fixed_bytes32(ref self: WarpMemory, location: felt252) -> u256;
```
piwonskp, line 5: Shall we create our own `bytes32` type? Might help with readability and allow to implement specific traits for the type

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
+use warplib::warp_memory::WarpMemory;
+use warplib::warp_memory::WarpMemoryTrait;
+use warplib::warp_memory::WarpMemoryImpl;
+use integer::u256_from_felt252;
+use integer::u128_try_from_felt252;
+use integer::Into;
+use option::OptionTrait;
+
+trait WarpMemoryArraysTrait {
+    /// reads the pointer of an array. If it does not exists, it creates a new one
```
piwonskp:
```suggestion
    /// reads the pointer of an array. If it does not exist, it creates a new one
```

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
...
+    fn read_id(ref self: WarpMemory, location: felt252, size: felt252) -> felt252;
```
piwonskp: It's rather `read_or_create` . `read` could be a separate method

warplib/src/warp_memory.cairo
```diff
@@ -14,24 +31,69 @@ impl DestructWarpMemory of Destruct::<WarpMemory> {
...
+        self.memory.insert(position, value);
+    }
+
+    fn write(ref self: WarpMemory, position: felt252, value: felt252){
+        let position_256 = u256_from_felt252(position);
+        let pointer_256 = u256_from_felt252(self.pointer);
+        if position_256 >= pointer_256 {
+           panic_with_felt252('Writing on unreserved position');
+        }
         self.memory.insert(position, value);
```
piwonskp, line 58: Is optimization the reason why you don't use `unsafe_write` here?

warplib/src/warp_memory.cairo
```diff
@@ -14,24 +31,69 @@ impl DestructWarpMemory of Destruct::<WarpMemory> {
...
+        self.memory.get(position)
+    }
+
+    fn unsafe_alloc(ref self: WarpMemory, size: felt252) -> felt252 {
+        let reserved_pointer = self.pointer;
+        self.pointer += size;
+        
+        reserved_pointer
+    }
+    fn alloc(ref self: WarpMemory, size: felt252) -> felt252 {
```
piwonskp:
```suggestion
    }
    
    fn alloc(ref self: WarpMemory, size: felt252) -> felt252 {
```

warplib/src/warp_memory.cairo
```diff
@@ -14,24 +31,69 @@ impl DestructWarpMemory of Destruct::<WarpMemory> {
...
+        let pointer_256 = u256_from_felt252(self.pointer);
+        let size_256 = u256_from_felt252(size);
+
+        let MAX_FELT_256: u256 = u256_from_felt252(-1);
+        if (pointer_256 + size_256 > MAX_FELT_256){
+            panic_with_felt252('Maximum memory size exceded');
+        }
+
+        let reserved_pointer= self.pointer;
```
piwonskp:
```suggestion
        let reserved_pointer = self.pointer;
```

warplib/src/warp_memory/bytes.cairo
rodrodros: Yes, I think that is a good idea!

warplib/src/warp_memory/arrays.cairo
rodrodros: it creates an empty pointer for reading solidity uninitialized memory. It does not create anything meaningful, also `read_id` is all over the code. I think `read_or_create` can be misleading. `read_id` gets a pointer to an array, if there is none, then it returns a default one.

warplib/src/warp_memory.cairo
rodrodros: Yes, no need for the extra call

warplib/src/warp_memory/write.cairo
rodrodros: Oops, I thought it got deleted on the merge

warplib/src/warp_memory/bytes.cairo
JorikSchellekens: More generally, can we preserve all solidity types?

warplib/src/warp_memory/arrays.cairo
piwonskp: Imo `read_id` is misleading since nobody expects `read_id` to write to the memory. Maybe `read_or_initialize`?

warplib/src/warp_memory.cairo
piwonskp: We should investigate if that's necessary with cairo 1. In any case imo we could inline those functions compile-time via e.g. plugin. Anyway, let's not split hairs on that now, there are higher priority tasks

warplib/src/warp_memory/arrays.cairo
rodrodros: Yeah you are right actually, let's use `get_or_create_id`

warplib/src/warp_memory/bytes.cairo
rodrodros: Yes, and with their own trait system which implements all solidity operations over them: like index for `bytesX`

warplib/src/warp_memory.cairo
```diff
@@ -14,24 +25,70 @@ impl DestructWarpMemory of Destruct::<WarpMemory> {
...
+        
+        reserved_pointer
+    }
+
+    fn alloc(ref self: WarpMemory, size: felt252) -> felt252 {
+        let pointer_256 = u256_from_felt252(self.pointer);
+        let size_256 = u256_from_felt252(size);
+
+        let MAX_FELT_256: u256 = u256_from_felt252(-1);
+        if (pointer_256 + size_256 > MAX_FELT_256){
```
AlejandroLabourdette, line 86: You can directly use `assert` function here

https://github.com/starkware-libs/cairo/blob/74dbf871c1125a591591519b75f987df1fb2d08d/corelib/src/lib.cairo#L391

warplib/src/warp_memory.cairo
```diff
@@ -14,24 +25,70 @@ impl DestructWarpMemory of Destruct::<WarpMemory> {
...
-    fn append(ref self: WarpMemory, value: felt252) {
-        self.insert(self.pointer, value);
+    fn unsafe_read(ref self: WarpMemory, position: felt252) -> felt252 {
+        self.memory.get(position)
+    }
+
+    fn read(ref self: WarpMemory, position: felt252) -> felt252 {
+        let position_256 = u256_from_felt252(position);
+        let pointer_256 = u256_from_felt252(self.pointer);
+        if position_256 >= pointer_256 {
```
AlejandroLabourdette, line 68: also here

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
...
+    /// Given the pointer to a dynamic array, it returns it's length
+    fn length_dyn(ref self: WarpMemory, array_ptr: felt252) -> felt252;
+}
+
+
+impl WarpMemoryArraysImpl of WarpMemoryArraysTrait {
+    fn get_or_create_id(ref self: WarpMemory, location: felt252, size: felt252) -> felt252{
+        // Get the id at the location
+        let id = self.read(location);
+        if id != 0 {
```
AlejandroLabourdette: What if the array was found but it doesn't have the specified size? It's assumed that is not a static array?

warplib/src/warp_memory.cairo
AlejandroLabourdette: This approach is also used in other pr's scripts

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        // If no id was found create a new one
+        let id = self.alloc(size);
+        self.unsafe_write(location, id);
+        id
+    }
+
+    fn new_dynamic_array(ref self: WarpMemory, len: felt252, elem_width: felt252) -> felt252 {
+        // The max space needed for the array in memory
```
AlejandroLabourdette, line 51: Why the max possible space if it's a dynamic array?

piwonskp:
Let's implement some basic unit tests for warplib

warplib/src/warp_memory.cairo
rodrodros: yes using assert is more readable. In this specific case where we are dealing with functions that have a high probability of being called  a lot, reducing an extra function call is noticeable performance wise

warplib/src/warp_memory/arrays.cairo
rodrodros: I don't understand the question. This is a very low level methods which purpose is to create an empty array on the fly when reading from uninitialized one. This is mimicking the solidity way. It trusts that is being called correctly since there is no way of checking it

warplib/src/warp_memory/arrays.cairo
rodrodros: solidity memory arrays work the same as C# arrays. The size can be dynamically set once, after that you cannot change them. A memory dynamic array cannot alter it's size once created

AlejandroLabourdette: Solidity memory arrays are kind of fixed size, just like c# as you said. But we are now at Cairo level and this is a function named `new_dynamic_array` so when i review it I was expecting the creation of an array that can grow and shrink during execution; perhaps the confusion is with that function name and what we are doing here is the creation of a fixed array. What do you think?

rodrodros: They are not a fixed size, they have a dynamic size which is set on runtime. Fixed size arrays are static arrays whose size is known during compilation.

warplib/src/warp_memory/arrays.cairo
AlejandroLabourdette: Oh ok, you can't check if in that location is stored an N length fixed array

warplib/src/warp_memory/arrays.cairo
rodrodros: What you say is true though, dynamic arrays are generally regarded as capable of resizing and this memory ones cannot. We are using `dynamic` here to specify we are referring to it's size is dynamically set. On further though a new name can be used

rodrodros: Or not. We have been referring to them as dynamic for a year now. So the whole code refers them as dynamic and for a rename here to be less confusing, we should then rename many other things to appropriately  refer to this kind of array. 
Also in solidity there is no semantic difference when creating storage arrays vs memory arrays, and the former is truly dynamic.

I think if we rename we should go and rename everything that deals with memory and arrays, and that's a lot of code. I think renaming this single function can cause confusion since this type is treated as dynamic arrays everywhere.

AlejandroLabourdette: I think we at least should do a comment there for future contributors. However I'm in favor of avoid confusions and do a proper naming, I understand it involve some more changes so perhaps could be a good Idea to create a ticket for it

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,281 @@
+use warplib::warp_memory::WarpMemory;
+use warplib::warp_memory::WarpMemoryTrait;
+use warplib::warp_memory::WarpMemoryImpl;
+
+// ==================== Basic Functions ====================
+
+#[test]
+fn test_init(){
```
piwonskp: I'd say it's rather testing alloc

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,281 @@
...
+    assert(warp_memory.pointer == 15, 'Pointer is not 15');
+}
+
+#[test]
+#[should_panic]
+fn test_alloc(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    warp_memory.alloc(10);
+    assert(warp_memory.pointer == 10, 'Pointer is not 10');
```
piwonskp: It was already tested in the previous function, I'd remove that and call the function `test_incorrect_alloc` or `negative_alloc`

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,281 @@
...
+#[test]
+fn test_init(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    assert(warp_memory.pointer == 0, 'Pointer is not 0');
+
+    warp_memory.unsafe_alloc(5);
+    assert(warp_memory.pointer == 5, 'Pointer is not 5');
+
+    warp_memory.alloc(10);
+    assert(warp_memory.pointer == 15, 'Pointer is not 15');
```
piwonskp: I think the better messages would be `X should be y`

warplib/src/warp_memory/accessors.cairo
```diff
@@ -58,8 +102,7 @@ impl WarpMemoryAccesssor of WarpMemoryAccesssorTrait {
         let mut serialization_array: Array<felt252> = ArrayImpl::<felt252>::new();
         TSerde::serialize(ref serialization_array, value);
 
-        let position = self.pointer;
-        self.alloc(u32_to_felt252(serialization_array.len()));
+        let position = self.alloc(u32_to_felt252(serialization_array.len()));
```
piwonskp: Should we prefer explicit functions or `.into()` syntax in your opinion? In such cases

rodrodros: I agree

tests/warplib.test.ts
```diff
@@ -1,31 +1,23 @@
...
-describe('Warplib files should compile', function () {
+describe('Warplib files should compile and execute correctly', function () {
   this.timeout(1800000);
 
-  let compileResults: SafePromise<{ stderr: string }>[];
-
+  let compileResult: SafePromise<{ stderr: string }>;
   before(function () {
-    compileResults = paths.map((file) => wrapPromise(starknetCompile(file, `${file}.json`)));
+    compileResult = wrapPromise(cairoTest(path.resolve(__dirname, '..', 'warplib')));
```
piwonskp: Use `BASE_PATH` instead

tests/util.ts
```diff
@@ -49,6 +50,11 @@ export function starknetCompile(
   );
 }
 
+export function cairoTest(cairoProjectPath: string): Promise<{ stdout: string; stderr: string }> {
+  const cairoTestBin = path.resolve(__dirname, '..', 'cairo1', getPlatform(), 'bin', 'cairo-test');
```
piwonskp: Use `BASE_PATH`. Perhaps we should make a function within warp that would construct the path on top of `BASE_PATH`? Seems like a cleaner option than always doing `path.resolve(BASE_PATH, ...)`

tests/warplib.test.ts
```diff
@@ -1,31 +1,23 @@
...
+import { cairoTest, SafePromise, wrapPromise } from './util';
 
-const paths = findCairoSourceFilePaths('warplib', true);
-describe('Warplib files should compile', function () {
+describe('Warplib files should compile and execute correctly', function () {
   this.timeout(1800000);
 
-  let compileResults: SafePromise<{ stderr: string }>[];
-
+  let compileResult: SafePromise<{ stderr: string }>;
```
piwonskp: Is there a reason not to define it directly within `it`? I don't see any reason to run tests within `before`

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    warp_memory.write(3, 5); 
+}
+
+#[test] 
+#[should_panic]
+fn test_unallocated_read() {
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    warp_memory.read(3); 
+}
```
piwonskp, line 54: I suggest rearranging functions to group `#[should_panic]` together. It's hard to track whether the test should throw exception or not

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+
+#[test]
+fn test_access(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let ptr1 = warp_memory.unsafe_alloc(2);
+    assert(ptr1 == 0, 'Pointer is not 0');
+    assert(warp_memory.pointer == 2, 'Pointer is not 2');
```
piwonskp:
```suggestion
    assert(ptr1 == 0, 'Pointer should be 0');
    assert(warp_memory.pointer == 2, 'Pointer should be 2');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+
+    warp_memory.write(ptr1, 2); 
+    warp_memory.write(ptr1 + 1, 3); 
+    assert(warp_memory.read(ptr1) == 2, 'Invalid read value');
+    assert(warp_memory.read(ptr1 + 1) == 3, 'Invalid read value');
```
piwonskp:
```suggestion
    assert(warp_memory.read(ptr1) == 2, 'Invalid read value on position 0');
    assert(warp_memory.read(ptr1 + 1) == 3, 'Invalid read value on position 1');
```

piwonskp:
```suggestion
    assert(warp_memory.read(ptr1) == 2, 'Invalid read value on position 0');
    assert(warp_memory.read(ptr1 + 1) == 3, 'Invalid read value on position 1');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+
+    let ptr2 = warp_memory.unsafe_alloc(3);
+    assert(ptr2 == 2, 'Pointer is not 2');
+    assert(warp_memory.pointer == 5, 'Pointer is not 5');
```
piwonskp:
```suggestion
    assert(ptr2 == 2, 'Pointer should be 2');
    assert(warp_memory.pointer == 5, 'Pointer should be 5');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+
+    warp_memory.write(ptr2, 4); 
+    warp_memory.write(ptr2 + 1, 5); 
+    warp_memory.write(ptr2 + 2, 6); 
+    assert(warp_memory.read(ptr2) == 4, 'Invalid read value');
+    assert(warp_memory.read(ptr2 + 1) == 5, 'Invalid read value');
+    assert(warp_memory.read(ptr2 + 2) == 6, 'Invalid read value');
```
piwonskp:
```suggestion
    assert(warp_memory.read(ptr2) == 4, 'Invalid read value on position 2');
    assert(warp_memory.read(ptr2 + 1) == 5, 'Invalid read value on position 3');
    assert(warp_memory.read(ptr2 + 2) == 6, 'Invalid read value on position 4');
```

warplib/src/warp_memory.cairo
```diff
@@ -1,5 +1,16 @@
...
+
+mod bytes;
+use bytes::WarpMemoryBytesTrait;
+
+mod accessors;
+use accessors::WarpMemoryAccesssorTrait;
+use accessors::WarpMemoryMultiCellAccessorTrait;
 
 
 struct WarpMemory {
```
piwonskp, line 16: Since it's more or less the final version of warp memory we should come up with a better name than `pointer` imo. It's meaningless. Do you have any preferred name? Maybe we should just name it `length`? Or at least `free_space_pointer`?

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    assert(warp_memory.pointer == -1, 'Invalid pointer value');
+    assert(warp_memory.read(dyn_array) == -2, 'Invalid length value');
+}
+
+#[test]
+#[should_panic]
+fn test_new_dynamic_array_overflow_1(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let dyn_array = warp_memory.new_dynamic_array(-1, 1);
```
piwonskp: If you assign `-1` to a constant named `MAX_FELT252` it will be much more readable

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    warp_memory.index_static(static_array, 5, 2, length);
+}
+
+#[test]
+#[should_panic]
+fn test_dyn_length() {
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    let dyn_array = warp_memory.new_dynamic_array(5, 3);
+
+    assert(warp_memory.length_dyn(dyn_array) == 16, 'Invalid lenght');
```
piwonskp:
```suggestion
    assert(warp_memory.length_dyn(dyn_array) == 16, 'Invalid length');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+fn test_index_static_out_of_range(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    let length = 10;
+    let static_array = warp_memory.unsafe_alloc(length);
+
+    warp_memory.index_static(static_array, 5, 2, length);
+}
+
+#[test]
+#[should_panic]
```
piwonskp: Why should panic?

warplib/src/warp_memory/accessors.cairo
```diff
@@ -11,31 +13,75 @@ use warplib::warp_memory::WarpMemoryImpl;
 
 trait WarpMemoryAccesssorTrait {
     fn store<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, position: felt252, value: T);
-    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, start_pos: felt252, end_pos: felt252) -> T;
+    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, start_pos: felt252, size: felt252) -> T;
```
piwonskp:
```suggestion
    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, position: felt252, size: felt252) -> T;
```

warplib/src/warp_memory/accessors.cairo
```diff
@@ -11,31 +13,75 @@ use warplib::warp_memory::WarpMemoryImpl;
...
     fn create<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, value: T);
 }
 
 trait WarpMemoryMultiCellAccessorTrait {
     fn write_multiple(ref self: WarpMemory, position: felt252, ref value: Array::<felt252>);
-    fn read_multiple(ref self: WarpMemory, start_pos: felt252, end_pos: felt252) -> Array::<felt252>;
+    fn read_multiple(ref self: WarpMemory, start_pos: felt252, size: felt252) -> Array::<felt252>;
```
piwonskp:
```suggestion
    fn read_multiple(ref self: WarpMemory, position: felt252, size: felt252) -> Array::<felt252>;
```

warplib/src/warp_memory/accessors.cairo
```diff
@@ -11,31 +13,75 @@ use warplib::warp_memory::WarpMemoryImpl;
...
                 break();
             }
-            self.write(position, value.pop_front().unwrap());
+            self.unsafe_write(position, value.pop_front().unwrap());
+            position += 1;
         }
     }
 
-    fn read_multiple(ref self: WarpMemory, start_pos: felt252, end_pos: felt252) -> Array::<felt252> {
+    fn read_multiple(ref self: WarpMemory, start_pos: felt252, size: felt252) -> Array::<felt252> {
```
piwonskp:
```suggestion
    fn read_multiple(ref self: WarpMemory, position: felt252, size: felt252) -> Array::<felt252> {
```

warplib/src/warp_memory/accessors.cairo
```diff
@@ -58,13 +104,12 @@ impl WarpMemoryAccesssor of WarpMemoryAccesssorTrait {
...
 
-        let position = self.pointer;
-        self.alloc(u32_to_felt252(serialization_array.len()));
+        let position = self.alloc(serialization_array.len().into());
         self.write_multiple(position, ref serialization_array);
     }
 
-    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, start_pos: felt252, end_pos: felt252) -> T {
-        let serialization_array: Array<felt252> = self.read_multiple(start_pos, end_pos);
+    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, start_pos: felt252, size: felt252) -> T {
```
piwonskp:
```suggestion
    fn retrieve<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, position: felt252, size: felt252) -> T {
```

piwonskp:
I've seen the word "location" and "position" used interchangeably. I think we should build a proper dictionary and differentiate between free space pointer, the address to an individual cell (WarpMemoryTrait), the address to a group of cells (WarpMemoryAccessor) and  between static and dynamic (retrieved from memory e.g. when accessing dynamic arrays) addresses. Imo we could use the word `position` to refer to individual cells and `location` to refer to possibly larger data that may span across multiple cells

warplib/src/warp_memory.cairo
rodrodros: I like `free_space_pointer`

warplib/src/test/warp_memory_test.cairo
rodrodros: I think every rearregenment has it's pros and cons. The intent here was to arrange according to functionality since in my opinion is more natural and helpful. Indeed with this organization finding panicables test it would take more time, but in my experience I don't know many situations where that would be necessary.

piwonskp: I agree, what I meant that you could rearrange panicables within the specific functionality to be near each other. Eg. it could be in the order: test_access,  test_unsafe_access, test_alloc, test_alloc_overflow, test_unallocated_write,  test_unallocated_read. 

I didn't want to group them  for the purpose of searching them, it's rather the fact it's easy to miss the `#[should_panic]` annotation as I've been reading through it.  If you prefer any other solution to make it more visible the code in the function is expected to fail it's fine for me. E.g. names of tests could be more verbose like: `test_write_to_unallocated_space_should_fail`. That would make it clear the code within test is incorrect

piwonskp: Also I think `test_` prefix is not necessary. It makes the name only longer. I'd stick to mocha's way of defining tests so we could name functions in the manner of `x_should_y`

rodrodros: I think `test` is a good prefix and `should_panic` is a good suffix even if they are both described by attributes. I left the first and added the second

piwonskp: Ok, fine by me

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,93 @@
...
+            return id;
+        }
+
+        // If no id was found create a new one
+        let id = self.alloc(size);
+        self.unsafe_write(location, id);
+        id
+    }
+
+    fn new_dynamic_array(ref self: WarpMemory, len: felt252, elem_width: felt252) -> felt252 {
```
piwonskp, line 50: I think we can find a better name, since it's not adding elements instantly. Shall we name it `allocate_dynamic_array`? Maybe `declare_dynamic_array` or `reserve_memory_for_dynamic_array`?

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,308 @@
...
+}
+
+#[test]
+fn test_unsafe_access() {
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    warp_memory.unsafe_write(3, 5); 
+    
+    let read_from_unallocated = warp_memory.unsafe_read(3);
+    assert(read_from_unallocated == 5, 'Invalid read value');
+    assert(warp_memory.free_space_pointer == 0, 'Pointer should be 0');
```
piwonskp:
```suggestion
    assert(warp_memory.free_space_pointer == 0, 'Free space pointer should be 0');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,308 @@
...
+    warp_memory.read(3); 
+}
+
+#[test]
+fn test_access(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let ptr1 = warp_memory.unsafe_alloc(2);
+    assert(ptr1 == 0, 'Pointer should be 0');
+    assert(warp_memory.free_space_pointer == 2, 'Pointer should be 2');
```
piwonskp:
```suggestion
    assert(ptr1 == 0, 'Pointer to allocated memory should be 0');
    assert(warp_memory.free_space_pointer == 2, 'Free space pointer should be 2');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,308 @@
...
+
+    warp_memory.write(ptr1, 2); 
+    warp_memory.write(ptr1 + 1, 3); 
+    assert(warp_memory.read(ptr1) == 2, 'Invalid read value on pos 0');
+    assert(warp_memory.read(ptr1 + 1) == 3, 'Invalid read value on pos 1');
+
+    let ptr2 = warp_memory.unsafe_alloc(3);
+    assert(ptr2 == 2, 'Pointer should be 2');
+    assert(warp_memory.free_space_pointer == 5, 'Pointer should be 5');
```
piwonskp:
```suggestion
    assert(ptr2 == 2, 'Pointer to allocated memory should be 2');
    assert(warp_memory.free_space_pointer == 5, 'Free space pointer should be 5');
```

warplib/src/warp_memory/arrays.cairo
rodrodros: I think all of those are good names. But I don't see how are they better than the current one

piwonskp: The problem is: how would you name a function that both allocates memory and writes it's contents at the same time? We should be able to semantically differentiate between the two

piwonskp: It's hard to tell what exactly `new` means then

warplib/src/test/warp_memory_test.cairo
rodrodros: Yeah, changed it so it does not panic

warplib/src/warp_memory/arrays.cairo
rodrodros: As i see it `new` means there is a dynamic array ready to be used

rodrodros: Anyone can go see the function for it's specific behaviour

piwonskp: Anyone always can analyze the code. The question is if it's easier or harder to do so. The purpose of name is to be a short description of function's functionalities.

 As I've asked how do we name a function that allocates memory and writes contents? Currently it's `wm${this.generatedFunctionsDef.size}_dynamic_array`. So e.g. `wm3_dynamic_array` would be calling `new_dynamic_array` followed by writes. Should we keep it that way?

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    let read_id2 = warp_memory.get_or_create_id(ptr1 + 1, 10);
+    assert(read_id2 == 2, 'Invalid read id value');
+    assert(warp_memory.free_space_pointer == 12, 'Invalid pointer value');
+}
+
+#[test]
+fn test_new_dynamic_array(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let dyn_array = warp_memory.new_dynamic_array(-2, 1);
```
piwonskp:
```suggestion
    let dyn_array = warp_memory.new_dynamic_array(MAX_FELT-1, 1);
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    assert(warp_memory.free_space_pointer == -1, 'Invalid pointer value');
```
piwonskp:
```suggestion
    assert(warp_memory.free_space_pointer == MAX_FELT, 'Invalid pointer value');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    assert(warp_memory.read(dyn_array) == -2, 'Invalid length value');
```
piwonskp:
```suggestion
    assert(warp_memory.read(dyn_array) == MAX_FELT - 1, 'Invalid length value');
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+#[should_panic]
+fn test_new_dynamic_array_overflow_1_should_panic(){
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let dyn_array = warp_memory.new_dynamic_array(MAX_FELT, 1);
+}
+
+#[test]
+#[should_panic]
+fn test_new_dynamic_array_overflow_2_should_panic(){
```
piwonskp: Let's be a little more verbose:
```suggestion
fn test_complex_element_dynamic_array_overflow__should_panic(){
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    let mut warp_memory = WarpMemoryTrait::initialize();
+
+    let dyn_array = warp_memory.new_dynamic_array(-2, 1);
+    assert(warp_memory.free_space_pointer == -1, 'Invalid pointer value');
+    assert(warp_memory.read(dyn_array) == -2, 'Invalid length value');
+}
+
+#[test]
+#[should_panic]
+fn test_new_dynamic_array_overflow_1_should_panic(){
```
piwonskp:
```suggestion
fn test_new_dynamic_array_overflow_should_panic(){
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+
+    let mut values = ArrayTrait::new();
+    values.append(2);
+    values.append(3);
+    values.append(5);
+
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    warp_memory.unsafe_alloc(MAX_FELT);
+    
+    warp_memory.write_multiple(-3, ref values);
```
piwonskp:
```suggestion
    warp_memory.write_multiple(MAX_FELT - 2, ref values);
```

warplib/src/test/warp_memory_test.cairo
```diff
@@ -0,0 +1,307 @@
...
+    warp_memory.write_multiple(pointer, ref values);
+}
+
+#[test]
+#[available_gas(1000000)]
+#[should_panic]
+fn test_read_multiple_overflow_should_panic() {
+    let mut warp_memory = WarpMemoryTrait::initialize();
+    warp_memory.unsafe_alloc(MAX_FELT);
+    warp_memory.read_multiple(-3, 3);
```
piwonskp:
```suggestion
    warp_memory.read_multiple(MAX_FELT - 2, 3);
```

warplib/src/warp_memory/arrays.cairo
rodrodros: I agree that names are not the best sometimes. I would like to postpone naming schemes after this part of memory is complete, and we have the whole picture. Renaming here is affecting the work in #1045 since it is a follow up to this one

piwonskp: Ok, I'm fine with postponing it

warplib/src/warp_memory/arrays.cairo
```diff
@@ -0,0 +1,91 @@
...
+        let id = self.read(location);
+        if id != 0 {
+            return id;
+        }
+
+        // If no id was found create a new one
+        let id = self.alloc(size);
+        self.unsafe_write(location, id);
+        id
+    }
```
piwonskp, line 48: This is the place I was talking about which can improve naming of static/dynamic addresses. `id` and `location` are  unintuitive, especially `id`. But as we spoke, let's fix it in later PRs

warplib/src/warp_memory/arrays.cairo
```diff
@@ -83,8 +82,7 @@ impl WarpMemoryArraysImpl of WarpMemoryArraysTrait {
              panic_with_felt252('Index out of range');
         } 
 
-        let index_location = array_ptr + index * elem_width;
-        self.read(index_location) 
+        array_ptr + index * elem_width
```
piwonskp, line 90: Why the change to return absolute pointer instead of the value itself?

piwonskp (approved):

warplib/src/warp_memory/accessors.cairo
```diff
@@ -0,0 +1,118 @@
...
+            }
+            array.append(self.read(index));
+            index += 1;
+        };
+        array
+    }
+}
+
+
+impl WarpMemoryAccesssor of WarpMemoryAccesssorTrait {
```
rjnrohit, line 95: can we write this `WarpMemoryAccesssor ` -> `WarpMemoryAccesssorImpl` as rather , not matter gonna anyway but let's keep this as a naming convention for traits, wdyt ?

warplib/src/warp_memory/accessors.cairo
```diff
@@ -0,0 +1,118 @@
...
+    fn create<T, impl TSerde: Serde<T>, impl TDrop: Drop<T>>(ref self: WarpMemory, value: T);
+}
+
+trait WarpMemoryMultiCellAccessorTrait {
+    fn write_multiple(ref self: WarpMemory, position: felt252, ref value: Array::<felt252>);
+    fn read_multiple(ref self: WarpMemory, position: felt252, size: felt252) -> Array::<felt252>;
+}
+
+
+impl WarpMemoryMultiCellAccessor of WarpMemoryMultiCellAccessorTrait {
```
rjnrohit, line 26: can we write this `WarpMemoryMultiCellAccessor`  -> `WarpMemoryMultiCellAccessorImpl` as rather , not matter gonna anyway but let's keep this as a naming convention for traits, wdyt ?

rjnrohit (approved):
lgtm

warplib/src/warp_memory/accessors.cairo
rodrodros: We talked and decided to remove the `Impl`  suffix whenever possible. Only leaving the `Trait` suffix for traits. I would think adding `Impl` as a suffix is important when multiple types implement the same trait since in this case only WarpMemory will be doing it, I wouldn't mind leaving it as it is.

warplib/src/warp_memory/arrays.cairo
rodrodros: Yeah, that was actually an error on my side. The index function returns the address of the "object" you want to read. What it was doing was getting the address and then reading from it which is a wrong behaviour since it should only get it and let other functions (such as accessors) read from that position

warplib/src/warp_memory/arrays.cairo
rodrodros: yes, I understand. Which name you think would be descriptive enough without being overly verbose?

piwonskp: Maybe instead of `location` it could be `pointer_position` and instead of `id` it could be `array_location`? So the name of function could be `get_array_location_or_alloc_new`. WYT?

PR #1040, opened 2023-04-26
Use Serde trait to store/retrieve from memory

### LINKED ISSUE #1012: Warp memory translation layer
Warp memory stores any data type as a list of felts. When retrieving data from memory an adapter is needed to translate them back into respective types

piwonskp:
Fixes #1012


warplib/src/warp_memory.cairo
```diff
@@ -1,6 +1,9 @@
 use dict::Felt252Dict;
 use dict::Felt252DictTrait;
-
+use array::Array;
```
AlejandroLabourdette: You and @rodrigo-pino are working over the same file with significant changes in diferent PR's. 
Please take a look at #1038

piwonskp: Yea, I'd like the changes to be consistent with each other. I'd be happy to hear @rodrigo-pino opion on write memory traits and read memory traits since they seem contradictory to the approach I propose in the PR. Do we plan to operate on felts directly? My idea would be to separate existing memory trait into low level access trait operating on felts and an external interface which is high level operating on actual types. In my view only the external interface should be used directly (kind of public-private distinction)

warplib/src/warp_memory.cairo
rodrodros, line 1: What I had in mind was some how different. The sructure I had in mind is in #1038 where we separated all the memory old functionality into different traits, each with its own file:
* Basic functionality would go on _warp_memory.cairo_ which includes: allocation of memory cells, as well as reading and writing to those a single value.
* Reading stuff more complicated than a felt would go to to _warp_memory/read.cairo_ (I would place the read code of memory here: `retrieve` and `retrieve_raw`)
* Writing stuff on _warp_memory/write.cairo_ (I would place the part of your PR which writes here: `store` and `store_raw`)
* array functions on _warp_memory/arrays.cairo_
* etc ...

Memory inter-operations (memory working with memory, or memory working with storage) does not need Serde. That is one of the reasons I wouldn't include them in the main trait.

Finally, it would be good to compare Serde's performance against generated code before removing it completely

piwonskp: It seems we differ in the interpretation of basic functionality. For me `store` and `retrieve` are very basic high level functionality for functions that operate on typed values. `write` and `read` are on the other hand basic functionalities for functions operating directly on memory. Therefore we got 2 levels here, high level is the default, low level access is for potential optimization reasons. Both traits contain basic functionalities.

As per `store_raw_array` I do not interpret that as an array function per se, rather as a util/helper function to store multiple values in memory at once. It'd be weird to call ArrayTrait to store e.g. u256 thus putting it there would be a bad idea since it'd be misleading.

As for memory-to-memory or memory-to-storage those are internal warp functions. By internal I mean the outside world shouldn't know anything about their implementation, nor about the encoding of data within the memory. In my point of view functions like `store` and `retrieve` are the "gate to cairo world". The isolation of memory is important in order for it to be easily replaceable. If we code a lot of functions to use the memory directly it will be significantly harder to replace it since we'll have to rewirte their logic which otherwise we could simply reuse replacing only the source of data

As for optimization `store` and `retrieve` could be optimized by removing the buffer. The change however would probably require a change in the corelib to `Serde` trait first. I haven't explored the topic in details though

warplib/src/warp_memory.cairo
```diff
@@ -34,4 +41,45 @@ impl WarpMemoryImpl of MemoryTrait {
     fn append(ref self: WarpMemory, value: felt252) {
         self.insert(self.pointer, value);
     }
+
+    fn store_raw_array(ref self: WarpMemory, ref value: Array::<felt252>) {
```
rodrodros: perhaps a `store_raw` name

rodrodros (approved):

warplib/src/warp_memory.cairo
piwonskp: I would like to adhere to your naming convention so `write_cells` or `write_multiple` will be bettter. It'd reflect that it is a util function same as `write`, operating on the same level, but for multiple cells

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -102,23 +108,13 @@ export class MemoryReadGen extends StringIndexedFuncGen {
...
       name: funcName,
-      code: [
-        `func ${funcName}{range_check_ptr, warp_memory : DictAccess*}(loc: felt) ->(val: ${resultCairoType}){`,
-        `    alloc_locals;`,
-        ...reads.map((s) => `    ${s}`),
-        `    return (${pack},);`,
-        '}',
-      ].join('\n'),
-      functionsCalled: [this.requireImport(...DICT_READ)],
+      code: endent`fn ${funcName}(loc: felt) -> ${typeToRead.toString()}{
```
rodrodros: I think the `#[implicits(warp_memory: WarpMemory)]` is missing here

piwonskp: True

PR #1044, opened 2023-05-09
Update warplib generator to not only write in math

AlejandroLabourdette:


warplib/src/external_input_check.cairo
rodrodros, line 1: oops, this shouldn't be here

AlejandroLabourdette: where do you think is a best location?

AlejandroLabourdette: > oops, this shouldn't be here

oh, forget to put it in gitignore

src/warplib/generateWarplib.ts
```diff
@@ -61,34 +61,49 @@ const warplibFunctions: WarplibFunctionInfo[] = [
...
-    return `mod ${fileName};\n${useFuncNames}`;
-  })
-  .join('\n\n');
+generateWarplibFor('maths', mathWarplibFunctions);
+generateWarplibFor('conversions', conversionWarplibFunctions);
+generateWarplibFor('external_input_check', inputCheckWarplibFunctions);
+
+function generateWarplibFor(folderName: string, functions: WarplibFunctionInfo[]) {
+  functions.forEach((warpFunc: WarplibFunctionInfo) => generateFile(warpFunc, folderName));
+  const mathsContent: string = folderContent(folderName);
```
rodrodros: nit mathsContent -> content or a more general name

rodrodros (approved):

src/warplib/generateWarplib.ts
rodrodros, line 1: I haven't noticed at first, but you should update gatherWarplib.ts as well. Since it won't read warplib functions outside of Maths

rodrodros: I've checked and it is not required

rjnrohit (approved):
lgtm

PR #1043, opened 2023-04-28
fixed bytes operations in warplib

AlejandroLabourdette:

PR #1045, opened 2023-05-09
Update warp memory functionality to Cairo 1

rodrodros:


warplib/src/maths/utils.cairo
```diff
@@ -1,72 +1,76 @@
-from starkware.cairo.common.math import split_felt, unsigned_div_rem
-from starkware.cairo.common.math_cmp import is_le
-from starkware.cairo.common.uint256 import Uint256, uint256_le
+//from starkware.cairo.common.math import split_felt, unsigned_div_rem
```
temyurchenko: Let's delete instead of commenting. If need to look this up, we can look in older revisions

src/warplib/implementations/conversions/dynBytesToFixed.ts
```diff
@@ -23,7 +22,7 @@ export function functionaliseBytesToFixedBytes(
         ['width', createUint8TypeName(ast)],
       ];
 
-  const importedFunc = ast.registerImport(node, WARPLIB_MEMORY, funcName, args, [
+  const importedFunc = ast.registerImport(node, ['this needs to be set'], funcName, args, [
```
temyurchenko: Is that a FIXME that needs to be finished?

src/utils/utils.ts
```diff
@@ -141,6 +144,10 @@ export function primitiveTypeToCairo(
 
   if (uintMatch) {
     const bits = BigInt(uintMatch[1]);
+    if (bits === 252n) {
```
temyurchenko, line 147: I'm not 100% sure it's correct, felt is actually lesser than 252-bit wide integer. Maybe there is a convention that 'u252' always stands for felt252 and u252, since there is no u252 type in Solidity? In that case, why can't we straight up use 'felt252' instead of 'u252'? It seems that conflating base-2 integers and felts might be undesirable

src/utils/utils.ts
```diff
@@ -155,7 +162,13 @@ export function primitiveTypeToCairo(
...
   }
 
-  return 'felt';
+  console.log(
+    warning('Warning:'),
+    `Unknown translation for primitive type ${typeString}.`,
+    'Defaulting to felt252',
+  );
+  return `felt252`;
+  // throw new Error(`Unknown translation for cairo type: ${typeString}`);
```
temyurchenko, line 171: Why not throw an error? Incorrect type assignment might lead to unexpected errors further down the pipeline. While the warning might help to find the source, it is less direct and easier to not notice, especially when running in CI when the output is large. What is the benefit over throwing?

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -117,44 +113,53 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
...
     ];
+    const calledFuncs: FunctionDefinition[] = dynamic
+      ? [this.requireImport(...WM_WRITE), funcCall]
+      : [funcCall];
+
     return {
       name: funcName,
       code: endent`
-        #[implicits(warp_memory: WarpMemory)]
+        #[implicit(warp_memory: WarpMemory)]
```
temyurchenko, line 155: Why? What is the difference?

temyurchenko (commented):
I haven't looked at everything yet, but just some preliminary stuff

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -96,44 +96,40 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
-        #[implicit(warp_memory)]
-        func ${funcName}(calldata : ${callDataType}) -> (mem_loc: felt){
-            alloc_locals;
-            let (len256) = felt_to_uint256(calldata.len);
-            let (mem_start) = wm_new(len256, ${uint256(memoryElementWidth)});
-            ${funcName}_elem(calldata.ptr, mem_start + 2, calldata.len);
-            return (mem_start,);
+
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(calldata : ${callDataType}) -> felt{
```
piwonskp:
```suggestion
        fn ${funcName}(calldata : ${callDataType}) -> felt252 {
```

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -96,44 +96,40 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
+            let mem_start = warp_memory.new_dynamic_array(calldata.len, ${memoryElementWidth});
+            ${funcName}_elem(calldata, mem_start + 1, calldata.len);
```
piwonskp, line 131: ${funcName}_copy or _fill imo would be a better name

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -211,8 +205,8 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
         if (isReferenceType(type)) {
           const recursiveFunc = this.getOrCreateFuncDef(type);
           const code = [
-            `let (member_${decl.name}) = ${recursiveFunc.name}(calldata.${decl.name});`,
-            `dict_write{dict_ptr=warp_memory}(${add('mem_start', offset)}, member_${decl.name});`,
+            `let member_${decl.name} = ${recursiveFunc.name}(calldata.${decl.name});`,
+            `warp_memory.unsafe_write(${add('mem_start', offset)}, member_${decl.name});`,
           ].join('\n');
```
piwonskp: Use endent

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -221,16 +215,12 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
             ? [
-                `dict_write{dict_ptr=warp_memory}(${add('mem_start', offset)}, calldata.${
-                  decl.name
-                }.low);`,
-                `dict_write{dict_ptr=warp_memory}(${add('mem_start', offset + 1)}, calldata.${
+                `warp_memory.unsafe_write(${add('mem_start', offset)}, calldata.${decl.name}.low);`,
+                `warp_memory.unsafe_write(${add('mem_start', offset + 1)}, calldata.${
                   decl.name
                 }.high);`,
               ].join('\n')
```
piwonskp: Use endent

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -105,11 +104,9 @@ export class MemoryArrayConcat extends StringIndexedFuncGen {
       return {
         name: funcName,
         code: endent`
-          #[implicit(warp_memory)]
-          func ${funcName}() -> (res_loc : felt){
-             alloc_locals;
-             let (res_loc) = wm_new(${uint256(0)}, ${uint256(1)});
-             return (res_loc,);
+          #[implicit(warp_memory: WarpMemory)]
+          fn ${funcName}() -> felt{
```
piwonskp:
```suggestion
          fn ${funcName}() -> felt252 {
```

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -149,18 +146,16 @@ export class MemoryArrayConcat extends StringIndexedFuncGen {
     );
 
     const code = endent`
-      #[implicit(warp_memory)]
-      func ${funcName}(${cairoArgs}) -> (res_loc : felt){
-          alloc_locals;
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}(${cairoArgs}) -> (res_loc : felt){
```
piwonskp:
```suggestion
      fn ${funcName}(${cairoArgs}) -> felt252 {
```

piwonskp:
Is there any behaviour test we could uncomment?

src/cairoUtilFuncGen/memory/arrayLiteral.ts
```diff
@@ -117,44 +113,53 @@ export class MemoryArrayLiteralGen extends StringIndexedFuncGen {
...
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const alloc_len = dynamic ? size * elementCairoType.width + 2 : size * elementCairoType.width;
+    const funcCall =
+      elementCairoType instanceof CairoFelt
+        ? this.requireImport(...WM_WRITE)
+        : this.requireImport(...WM_STORE);
     const writes = [
-      ...(dynamic ? [`wm_write_256{warp_memory=warp_memory}(start, ${uint256(size)});`] : []),
-      ...mapRange(size, (n) => elementCairoType.serialiseMembers(`e${n}`))
+      dynamic ? `warp_memory.write(start, ${arraySize});` : '',
```
piwonskp: Let's move it directly to the returned object

src/warplib/implementations/conversions/dynBytesToFixed.ts
rodrodros: yes, they should be introduced by #1043. Update it to make it clear

src/utils/utils.ts
rodrodros: Yes felt is lesser than 252 bits, but bigger than 251. The word size in Cairo is actually 252 bits.

The reason for this change is that previously (old warp) we could easily represent felts in the transpiled contract by using a any uint lesser than 256. 

Now in Cairo 1 there is an intention of having a `ux` per `uintx`, which means we have no way of representing `felts` now in the transpiled contracts. 

The use of `252n` is a hacky way of the transpiler to understand that we mean this type is going to be a felt. We cannot create our own `felt252` typestring because the solc typed ast which we use to parse all types everywhere would not know where it come from and error.

We decided this way because it was the much lesser evil among other options.  Also is easy to revert back if we need it.

src/utils/utils.ts
rodrodros: Yes, I added the warning because I thought you were working on that, and as a temporay patch. Also I don't think untyped literals will actually give errors on the transpiled contract.

rodrodros: By using `felt252` is the same as ignoring it. If it is not `felt252` since we do not annotate it, it will be inferred correctly by the transpiler.

I am not saying that I am willing to defend this solution on the future, only that for now it seems a safe compromise

src/cairoUtilFuncGen/memory/arrayLiteral.ts
rodrodros: `implicitS` is a reserved word by Cairo. `implict` is the attribute that triggers our plugin. It was misswritten.

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
rodrodros: My only issue against this is that all copy semantics have the same name. I would postpone it because to do it the right way we should update the name of every copy functions

rodrodros:
> Is there any behaviour test we could uncomment?

I wouldn't do it just yet. During memory development we found a bunch of small errors scattered everywhere we are currently solving. I think we should start testing behaviour with non memory/storage/calldata stuff first if possible, to also discard possible errors on the testnet side

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -187,7 +186,7 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
     let sourceLocationCode: string;
     if (targetType.elementT instanceof PointerType) {
       const idAllocSize = isDynamicArray(sourceType.elementT) ? 2 : cairoSourceElementType.width;
-      sourceLocationFunc = this.requireImport(...WM_READ_ID);
+      sourceLocationFunc = this.requireImport(...WM_GET_ID);
       sourceLocationCode = `let (source_elem) = wm_read_id(${sourceLoc}, ${uint256(idAllocSize)});`;
```
piwonskp: Shouldn't it be updated as well?

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -254,24 +250,14 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
     } else {
-      sourceLocationCode.push(
-        `let (source_elem) = ${memoryRead.name}(${getOffset(
-          'source',
-          'felt_index',
-          sourceTWidth,
-        )});`,
-      );
+      const offset = getOffset('source', 'index', sourceTWidth);
+      sourceLocationCode = `let (source_elem) = ${memoryRead.name}(${offset});`;
```
piwonskp:
```suggestion
      sourceLocationCode = `let source_elem = ${memoryRead.name}(${offset});`;
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -281,31 +267,27 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
 
     const memoryWrite = this.memoryWrite.getOrCreateFuncDef(targetType.elementT);
     const targetCopyCode = [
-      `let (target_elem_loc) = wm_index_dyn(target, index, ${uint256(targetTWidth)});`,
+      `let target_elem_loc = warp_memory.index_dyn(target, index, ${targetTWidth});`,
       `${memoryWrite.name}(target_elem_loc, target_elem);`,
-    ];
+    ].join('\n');
```
piwonskp: Please use `endent`

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -281,31 +267,27 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
         ${conversionCode}
-        ${targetCopyCode.join('\n')}
-        let (next_index, _) = uint256_add(index, ${uint256(1)});
-        return ${funcName}_copy(source, target, next_index, len);
+        ${targetCopyCode}
+        return ${funcName}_copy(source, target, next_index + 1, len);
       }
-      #[implicit(warp_memory)]
+      #[implicit(warp_memory: WarpMemory)]
       func ${funcName}(source : felt) -> (target : felt){
```
piwonskp:
```suggestion
      fn ${funcName}(source : felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -357,33 +339,30 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
 
-      #[implicit(warp_memory)]
-      func ${funcName}(source : felt) -> (target : felt){
-        alloc_locals;
-        let (len) = wm_dyn_array_length(source);
-        let (target) = wm_new(len, ${uint256(targetWidth)});
-        ${funcName}_copy(source, target, Uint256(0, 0), len);
-        return (target=target,);
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}(source: felt252) -> (target: felt252) {
```
piwonskp:
```suggestion
      fn ${funcName}(source: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -441,12 +420,12 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
       const conversionFunc = `warp_int${sourceType.nBits}_to_int${targetType.nBits}`;
       return [
-        `let (${targetVar}) = ${conversionFunc}(${sourceVar});`,
+        `let ${targetVar} = ${conversionFunc}(${sourceVar});`,
         [this.requireImport(INT_CONVERSIONS, conversionFunc)],
       ];
     } else if (!targetType.signed && targetType.nBits === 256 && sourceType.nBits < 256) {
       return [
-        `let (${targetVar}) = felt_to_uint256(${sourceVar});`,
+        `let ${targetVar} = felt_to_uint256(${sourceVar});`,
```
piwonskp: This whole conversion stuff can be replaced by a single call to `into()` in future, right?

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -1,57 +1,47 @@
...
       // for the contents of the newly accessed subobject
       args.push(
         createNumberLiteral(
           isDynamicArray(valueType)
-            ? 2
+            ? 1
             : CairoType.fromSol(valueType, this.ast, TypeConversionContext.MemoryAllocation).width,
           this.ast,
-          'uint256',
+          'uint252',
```
piwonskp: Let's put it as a constant somewhere and write a comment what exactly it is

piwonskp:
> > Is there any behaviour test we could uncomment?
> 
> I wouldn't do it just yet. During memory development we found a bunch of small errors scattered everywhere we are currently solving. I think we should start testing behaviour with non memory/storage/calldata stuff first if possible, to also discard possible errors on the testnet side

We are already running behaviour tests that are non memory/storage/calldata

src/cairoUtilFuncGen/memory/arrayLiteral.ts
rodrodros: Do you mean the `writes`? I prefer it outside, inside it makes it hard to understand what happens in the generated function

src/cairoUtilFuncGen/memory/implicitConversion.ts
rodrodros: Probably

src/utils/utils.ts
temyurchenko: Oh, does solc understand uint252? I'm surprised given it's not a legal Solidity type, it's not divisible by 8. If that's the case though, feel free to resolve

src/utils/utils.ts
temyurchenko: Oh, I did have a patch for other things for that function, but as you pointed out they were better handled in other places, so I don't really do anything with it.

src/cairoUtilFuncGen/memory/arrayLiteral.ts
piwonskp: I meant only the write of dynamic array's length. I agree writing elements outside is better. The length though is just a single write so I think it will be more visible if it is there

warplib/src/maths/utils.cairo
```diff
@@ -1,72 +1,10 @@
...
+use integer::u256_from_felt252;
+use integer::u128_to_felt252;
 
-func get_max{range_check_ptr}(op1, op2) -> (result: felt) {
-    let le = is_le(op1, op2);
-    if (le == 1) {
-        return (op2,);
-    } else {
-        return (op1,);
+fn u256_to_felt252(x: u256) -> felt252 {
```
piwonskp: Let's keep the convention of corelib `felt252_try_from_u256`

piwonskp: Also shouldn't it be in integer.cairo? We've got a file dealing with integers already

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -96,44 +96,40 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
+        let memory_elem = ${recursiveFunc.name}(calldata_elem);
+        warp_memory.unsafe_write(calldata_elem, memory_elem)
       `;
       auxFunc = recursiveFunc;
     } else if (memoryElementWidth === 2) {
       copyCode = endent`
-        dict_write{dict_ptr=warp_memory}(mem_start, calldata[0].low);
-        dict_write{dict_ptr=warp_memory}(mem_start+1, calldata[0].high);
+        warp_memory.unsafe_write(mem_ptr, calldata[index].low)
+        warp_memory.unsafe_write(mem_ptr + 1, calldata[index].high)
```
piwonskp, line 107: Wouldn't `warp_memory.store(mem_ptr, calldata[index])` be better here?

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -161,37 +157,35 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
+          warp_memory.unsafe_write(${loc(index)}, memory_elem);
         `;
       funcCalls = [recursiveFunc];
     } else if (memoryElementWidth === 2) {
       copyCode = (index) =>
         endent`
-          dict_write{dict_ptr=warp_memory}(${loc(index)}, calldata[${index}].low);
-          dict_write{dict_ptr=warp_memory}(${loc(index)} + 1, calldata[${index}].high);
+            warp_memory.unsafe_write(${loc(index)}, calldata[${index}].low);
+            warp_memory.unsafe_write(${loc(index)} + 1, calldata[${index}].high);
```
piwonskp, line 169: Again, wouldn't store be better?

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
```diff
@@ -210,27 +204,21 @@ export class CallDataToMemoryGen extends StringIndexedFuncGen {
...
-                  decl.name
-                }.high);`,
-              ].join('\n')
-            : `dict_write{dict_ptr=warp_memory}(${add('mem_start', offset)}, calldata.${
+            ? endent`
+                warp_memory.unsafe_write(${add('mem_start', offset)}, calldata.${decl.name}.low);
+                warp_memory.unsafe_write(${add('mem_start', offset + 1)}, calldata.${
                 decl.name
-              });`;
+              }.high);`
```
piwonskp, line 220: Again, wouldn't store be better?

src/cairoUtilFuncGen/memory/arrayConcat.ts
```diff
@@ -149,18 +156,16 @@ export class MemoryArrayConcat extends StringIndexedFuncGen {
     );
 
     const code = endent`
-      #[implicit(warp_memory)]
-      func ${funcName}(${cairoArgs}) -> (res_loc : felt){
-          alloc_locals;
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}(${cairoArgs}) -> felt252{
```
piwonskp:
```suggestion
      fn ${funcName}(${cairoArgs}) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -206,24 +203,22 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
     const allocSize = narrowBigIntSafe(targetType.size) * cairoTargetElementType.width;
     const funcName = `memory_conversion_static_to_static${this.generatedFunctionsDef.size}`;
     const code = endent`
-      #[implicit(warp_memory)]
-      func ${funcName}_copy(source : felt, target : felt, index : felt){
-         alloc_locals;
-         if (index == ${sourceType.size}){
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}_copy(source : felt252, target : felt252, index : felt252) {
```
piwonskp:
```suggestion
      fn ${funcName}_copy(source: felt252, target: felt252, index: felt252) {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -206,24 +203,22 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
-         return ${funcName}_copy(source, target, index + 1);
+         ${funcName}_copy(source, target, index + 1)
       }
 
-      #[implicit(warp_memory)]
-      func ${funcName}(source : felt) -> (target : felt){
-         alloc_locals;
-         let (target) = wm_alloc(${uint256(allocSize)});
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}(source : felt252) ->  felt252{
```
piwonskp:
```suggestion
      fn ${funcName}(source: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -206,24 +203,22 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
+         let target = wm_alloc(${allocSize});
```
piwonskp: `warp_memory.alloc`

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -280,32 +264,27 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
+      ${memoryWrite.name}(target_elem_loc, target_elem);`;
 
     const funcName = `memory_conversion_static_to_dynamic${this.generatedFunctionsDef.size}`;
     const code = endent`
-      #[implicit(warp_memory)]
-      func ${funcName}_copy(source : felt, target : felt, index : Uint256, len : Uint256){
-        alloc_locals;
-        if (len.low == index.low and len.high == index.high){
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}_copy(source : felt252, target : felt252, index : felt252, len : felt252){
```
piwonskp:
```suggestion
      fn ${funcName}_copy(source: felt252, target: felt252, index: felt252, len: felt252){
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -280,32 +264,27 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
       }
-      #[implicit(warp_memory)]
-      func ${funcName}(source : felt) -> (target : felt){
-        alloc_locals;
-        let len = ${uint256(sourceType.size)};
-        let (target) = wm_new(len, ${uint256(targetTWidth)});
-        ${funcName}_copy(source, target, Uint256(0, 0), len);
-        return (target=target,);
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}(source : felt252) -> felt252{
```
piwonskp:
```suggestion
      fn ${funcName}(source: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -357,33 +336,30 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
 
     const targetWidth = cairoTargetElementType.width;
     const funcName = `memory_conversion_dynamic_to_dynamic${this.generatedFunctionsDef.size}`;
     const code = endent`
-      #[implicit(warp_memory)]
-      func ${funcName}_copy(source : felt, target : felt, index : Uint256, len : Uint256){
-        alloc_locals;
-        if (len.low == index.low and len.high == index.high){
+      #[implicit(warp_memory: WarpMemory)]
+      fn ${funcName}_copy(source : felt252, target : felt252, index : felt252, len : felt252){
```
piwonskp:
```suggestion
      fn ${funcName}_copy(source: felt252, target: felt252, index: felt252, len: felt252){
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -441,14 +417,11 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
...
-        `let (${targetVar}) = ${conversionFunc}(${sourceVar});`,
+        `let ${targetVar} = ${conversionFunc}(${sourceVar});`,
         [this.requireImport(INT_CONVERSIONS, conversionFunc)],
       ];
     } else if (!targetType.signed && targetType.nBits === 256 && sourceType.nBits < 256) {
-      return [
-        `let (${targetVar}) = felt_to_uint256(${sourceVar});`,
-        [this.requireImport(...FELT_TO_UINT256)],
-      ];
+      return [`let ${targetVar}: u256 = ${sourceVar}.into();`, []];
```
piwonskp: It requires an import of `Into` trait and the implementation of into trait

src/cairoUtilFuncGen/memory/memoryRead.ts
```diff
@@ -1,57 +1,46 @@
...
 import { isDynamicArray, safeGetNodeType } from '../../utils/nodeTypeProcessing';
-import { GeneratedFunctionInfo, locationIfComplexType, StringIndexedFuncGen } from '../base';
+import { locationIfComplexType, StringIndexedFuncGen } from '../base';
 
 /*
   Produces functions that when given a start location in warp_memory, deserialise all necessary
   felts to produce a full value. For example, a function to read a Uint256 reads the given location
   and the next one, and combines them into a Uint256 struct
 */
-// TODO: Replace the whole class with a call to warp_memory.retrieve?
```
piwonskp, line 37: The class doesn't generate any cairo code, only returns an instance of imported function now, right? Thus in future it shouldn't be a generated function. Am I right or I miss something?

src/utils/utils.ts
rodrodros: Yes, luckily for us at this time

src/utils/utils.ts
rodrodros: So do you agree on leaving it as a warning for now, and if it becomes a reason for failure (or there are no longer other pressing matter), fix it?

warplib/src/maths/utils.cairo
rodrodros: > Also shouldn't it be in integer.cairo? We've got a file dealing with integers already

Yes it's a better location, I was respecting the status quo where this kind of function was located in

>  Let's keep the convention of corelib felt252_try_from_u256

You can see both types in corelib, a try and a not try. e.g: `u32_to_felt252`

src/cairoUtilFuncGen/memory/memoryStruct.ts
```diff
@@ -79,38 +79,33 @@ export class MemoryStructGen extends StringIndexedFuncGen {
...
-        `    let (start) = wm_alloc(${uint256(structType.width)});`,
-        mangledStructMembers
-          .flatMap(([name, type]) => type.serialiseMembers(name))
-          .map(write)
-          .join('\n'),
-        `    return (start,);`,
-        `}`,
+        endent`
+        #[implicits(warp_memory: WarpMemory)]
+        fn ${funcName}(${argString}) -> felt252{
```
piwonskp:
```suggestion
        fn ${funcName}(${argString}) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -119,14 +112,10 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(mem_loc : felt) -> (ret_data: ${outputType.toString()}){
-          alloc_locals;
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(mem_loc : felt) -> (ret_data: ${outputType.toString()}){
```
piwonskp:
```suggestion
        fn ${funcName}(mem_loc: felt252) -> ${outputType.toString()} {
```

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -191,20 +180,16 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
-        #[implicit(warp_memory)]
-        func ${funcName}(mem_loc: felt) -> (retData: ${outputType.toString()}){
-            alloc_locals;
-            let (len_256) = wm_read_256(mem_loc);
-            let (ptr : ${outputType.vPtr.toString()}) = alloc();
-            let (len_felt) = narrow_safe(len_256);
-            ${dynArrayReaderInfo.name}(len_felt, ptr, mem_loc + 2);
-            return (${calldataDynArrayStruct.name}(len=len_felt, ptr=ptr),);
+        #[implicit(warp_memory: WarpMemory)]
+        func ${funcName}(mem_loc: felt) -> (ret_data: ${outputType.toString()}){
```
piwonskp:
```suggestion
        fn ${funcName}(mem_loc: felt252) -> ${outputType.toString()} {
```

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -191,20 +180,16 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
+            let lenght = warp_memory.read(mem_loc);
```
piwonskp:
```suggestion
            let length = warp_memory.read(mem_loc);
```

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -191,20 +180,16 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
+            ${dynArrayReaderInfo.name}(length, ArrayTrait::new(), mem_loc + 1)
         }
         `,
       functionsCalled: [
-        this.requireImport(...ALLOC),
-        this.requireImport(...NARROW_SAFE),
-        this.requireImport([...WARPLIB_MEMORY], 'wm_read_256'),
+        this.requireImport(...ARRAY_TRAIT),
+        this.requireImport(...U256_TO_FELT252),
+        this.requireImport(...WM_READ),
```
piwonskp, line 191: Shouldn't array impl be also imported?

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -217,40 +202,38 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
-        `assert ptr[0] = mem_read1;`,
+        `let mem_id_location = ${readFunc.name}(mem_loc, ${allocSize});`,
+        `let mem_value = ${auxFunc.name}(mem_id_location);`,
+        `result.append(mem_value);`,
       ];
-      funcCalls = [this.requireImport(...U128_FROM_FELT), auxFunc, readFunc];
+      funcCalls = [auxFunc, readFunc];
     } else {
-      code = [`let (mem_read0) = ${readFunc.name}(mem_loc);`, 'assert ptr[0] = mem_read0;'];
+      code = [`let mem_value = ${readFunc.name}(mem_loc);`, 'result.append(mem_value);'];
```
piwonskp: The code parts can be written using endent

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -217,40 +202,38 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
...
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(len: felt, ptr: ${ptrString}*, mem_loc: felt) -> (){
-            alloc_locals;
-            if (len == 0){
-                 return ();
+        #[implicit(warp_memory: WarpMemory)]
+        func ${funcName}(result: Array<${cairoElementT}>, index: felt252, length: felt252 mem_loc: felt252) -> (){
```
piwonskp:
```suggestion
        fn ${funcName}(result: Array<${cairoElementT}>, index: felt252, length: felt252, mem_loc: felt252) {
```

src/cairoUtilFuncGen/calldata/calldataToMemory.ts
rodrodros: Better in a readability way, I don't think it would be better in any other way.

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -119,11 +111,10 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt, mem_loc: felt) -> (loc: felt){
-            alloc_locals;
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt, mem_loc: felt) -> felt {
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -149,11 +140,10 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
```
piwonskp:
```suggestion
        fn ${funcName}(loc : felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -200,36 +188,33 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
+    const memoryOffset = add('mem_loc', elementMemoryWidth);
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}_elem(storage_loc: felt, mem_loc : felt, length: felt) -> (){
-          alloc_locals;
-          if (length == 0){
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}_elem(storage_loc: felt, mem_loc : felt, length: felt) {
```
piwonskp:
```suggestion
        fn ${funcName}_elem(storage_loc: felt252, mem_loc : felt252, length: felt252) {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -200,36 +188,33 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
+          ${funcName}_elem(${storageOffset}, ${memoryOffset}, index);
         }
 
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt, mem_loc : felt) -> (loc : felt){
-            alloc_locals;
-            ${funcName}_elem(loc, mem_loc, ${length});
-            return (loc,);
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(storage_loc : felt, mem_loc : felt) -> felt {
```
piwonskp:
```suggestion
        fn ${funcName}(storage_loc: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -200,36 +188,33 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
+            ${funcName}_elem(storage_loc, mem_loc, ${length});
+            stroage_loc
```
piwonskp:
```suggestion
            storage_loc
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -289,48 +274,43 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}_elem(storage_name: felt, mem_loc : felt, length: Uint256) -> (){
-            alloc_locals;
-            if (length.low == 0 and length.high == 0){
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}_elem(storage_id: felt252, mem_loc: felt252, length: felt252) -> (){
```
piwonskp:
```suggestion
        fn ${funcName}_elem(storage_id: felt252, mem_loc: felt252, length: felt252) {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -289,48 +274,43 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
-            alloc_locals;
-            let (length) = ${lengthMappingName}.read(loc);
-            let (mem_length) = wm_dyn_array_length(mem_loc);
-            ${lengthMappingName}.write(loc, mem_length);
-            let (narrowedLength) = narrow_safe(mem_length);
-            ${funcName}_elem(loc, mem_loc + 2 + ${elementMemoryWidth} * narrowedLength, mem_length);
-            let (lesser) = uint256_lt(mem_length, length);
-            if (lesser == 1){
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(storage_id: felt, mem_loc: felt) -> felt {
```
piwonskp:
```suggestion
        fn ${funcName}(storage_id: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/implicitConversion.ts
rodrodros: yes, I think it would be a good issue to fix. Update the whole `generateScalingCode` family of funcs to Cairo 1. What do you think?

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
rodrodros: Impl are not needed. Traits are enough

src/utils/utils.ts
temyurchenko: I don't see any benefit of not raising an Error here: if we don't know what it is, it might not do what we want. I might be missing something. Do however you judge is the best.

src/cairoUtilFuncGen/memory/memoryRead.ts
rodrodros: Yes you are right. We could remove it in the near future, once serde is proven to be sufficient for whatever type we would like to read

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -119,11 +111,10 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt, mem_loc: felt) -> (loc: felt){
-            alloc_locals;
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt252, mem_loc: felt252) -> felt252 {
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -149,11 +140,10 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -200,36 +188,33 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
+    const memoryOffset = add('mem_loc', elementMemoryWidth);
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}_elem(storage_loc: felt, mem_loc : felt, length: felt) -> (){
-          alloc_locals;
-          if (length == 0){
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}_elem(storage_loc: felt252, mem_loc : felt252, length: felt252) {
```
piwonskp:
```suggestion
        fn ${funcName}_elem(storage_loc: felt252, mem_loc: felt252, length: felt252) {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -200,36 +188,33 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
+          ${funcName}_elem(${storageOffset}, ${memoryOffset}, index);
         }
 
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt, mem_loc : felt) -> (loc : felt){
-            alloc_locals;
-            ${funcName}_elem(loc, mem_loc, ${length});
-            return (loc,);
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(storage_loc : felt252, mem_loc : felt252) -> felt252 {
```
piwonskp:
```suggestion
        fn ${funcName}(storage_loc: felt252, mem_loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/memory/memoryToStorage.ts
```diff
@@ -350,14 +330,14 @@ export class MemoryToStorageGen extends StringIndexedFuncGen {
...
               ]
             : [
-                `let (${elemLoc}) = ${readFunc.name}(${add('mem_loc', memOffset)}, ${uint256(
-                  CairoType.fromSol(type, this.ast, TypeConversionContext.Ref).width,
-                )});`,
+                `let ${elemLoc} = ${readFunc.name}(${add('mem_loc', memOffset)}, ${
+                  CairoType.fromSol(type, this.ast, TypeConversionContext.Ref).width
+                });`,
                 `${auxFunc.name}(${add('loc', storageOffset)}, ${elemLoc});`,
               ];
```
piwonskp: Let's use endent instead

src/utils/utils.ts
rodrodros: I undestand your point and I agree with it. If we raise an error then we must make a parser for literals like "int_const 12345" and assign them a type. Also since we have not yet the full scope of which typestring is arriving to this functions, the warning serves as way to probe/learn the types that come into here. 

The purpose of this PR is not to introduce a fully implemented feature but a ground base. It will get very big if all were to be fixed here

temyurchenko: I see

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -118,17 +118,16 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt) -> (mem_loc: felt){
-          alloc_locals;
-          let (mem_start) = wm_alloc(${uint256(memoryType.width)});
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt252) -> felt252 {
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -164,17 +163,15 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt) -> (mem_loc : felt){
-          alloc_locals;
-          let length = ${uint256(memoryType.width)};
-          let (mem_start) = wm_alloc(length);
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt252) -> felt252{
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -207,38 +204,26 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
...
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}_elem(mem_start: felt, loc : felt, length: Uint256) -> (){
-          alloc_locals;
-          if (length.low == 0){
-            if (length.high == 0){
-              return ();
-            }
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}_elem(mem_start: felt, loc : felt, length: felt252) -> (){
```
piwonskp:
```suggestion
        fn ${funcName}_elem(mem_start: felt252, loc: felt252, length: felt252) {
```

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -207,38 +204,26 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
...
+      )}, index - 1)
         }
 
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt) -> (mem_loc : felt){
-            alloc_locals;
-            let length = ${uint256(length)};
-            let (mem_start) = wm_alloc(length);
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt) -> felt {
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252) -> felt252 {
```

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -267,26 +252,24 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
     const funcInfo: GeneratedFunctionInfo = {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}_elem(storage_name: felt, mem_start: felt, length: Uint256) -> (){
-            alloc_locals;
-            if (length.low == 0 and length.high == 0){
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}_elem(storage_name: felt252, mem_start: felt252, length: felt252) -> () {
```
piwonskp:
```suggestion
        fn ${funcName}_elem(storage_name: felt252, mem_start: felt252, length: felt252) {
```

src/cairoUtilFuncGen/storage/storageToMemory.ts
```diff
@@ -267,26 +252,24 @@ export class StorageToMemoryGen extends StringIndexedFuncGen {
...
+            ${funcName}_elem(storage_name, mem_start, index)
         }
 
-        #[implicit(warp_memory)]
-        func ${funcName}(loc : felt) -> (mem_loc : felt){
-            alloc_locals;
-            let (length: Uint256) = ${lengthMappingName}.read(loc);
-            let (mem_start) = wm_new(length, ${uint256(memoryElementType.width)});
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(loc : felt252) -> fetl252{
```
piwonskp:
```suggestion
        fn ${funcName}(loc: felt252) -> felt252 {
```

src/cairoWriter/writers/cairoImportFunctionDefinitionWriter.ts
```diff
@@ -5,6 +5,10 @@ import { CairoASTNodeWriter } from '../base';
 // Not being used as for now
 export class CairoImportFunctionDefinitionWriter extends CairoASTNodeWriter {
   writeInner(node: CairoImportFunctionDefinition, _writer: ASTWriter): SrcDesc {
+    if (node.path[node.path.length - 1].endsWith('Trait')) {
+      return [`use ${node.path.join('::')};`];
+    }
```
piwonskp, line 13: Why is it needed?

src/cli.ts
```diff
@@ -122,18 +122,18 @@ function runTranspile(files: string[], options: CliOptions) {
       .forEach((file: string) => {
         createCairoProject(path.join(options.outputDir, file));
         if (options.compileCairo) {
-          const { success, resultPath, abiPath } = compileCairo(
-            path.join(options.outputDir, file),
-            BASE_PATH,
-            options,
+          const { success, outputDir /*abiPath*/ } = compileCairo1(
```
piwonskp, line 125: Why do you comment that? In cairo 1 there is no separate abi file. The abi is a part of output. I suppose we should delete it then

src/cli.ts
```diff
@@ -122,18 +122,18 @@ function runTranspile(files: string[], options: CliOptions) {
...
+            path.dirname(path.dirname(path.join(options.outputDir, file))),
+            // BASE_PATH,
+            //options,
```
piwonskp, line 128: Again why is it commented?

src/io.ts
```diff
@@ -87,11 +89,10 @@ export function outputResult(
 
     const codeOutPath = path.join(outputLocation.dir, outputLocation.base);
     outputFileSync(codeOutPath, code);
-    // Cairo-format is disabled, as it has a bug
-    // if (options.formatCairo || options.dev) {
-    //   const warpVenvPrefix = `PATH=${path.resolve(__dirname, '..', 'warp_venv', 'bin')}:$PATH`;
-    //   execSync(`${warpVenvPrefix} cairo-format -i ${fullCodeOutPath}`);
-    // }
+    if (options.formatCairo || options.dev) {
+      const formatBin = path.join(CAIRO1_BINS, 'cairo-format');
```
piwonskp, line 93: Can we use the one provided by `cairo-lang`? it should be globally available I presume

src/utils/importFuncGenerator.ts
```diff
@@ -12,6 +12,102 @@ import {
...
+  Paths.WM_RETRIEVE,
+  Paths.WM_INDEX_DYN,
+  Paths.WM_UNSAFE_READ,
+  Paths.WM_INDEX_STATIC,
+  Paths.WM_UNSAFE_ALLOC,
+  Paths.WM_UNSAFE_WRITE,
+  Paths.WM_READ_MULTIPLE,
+  Paths.WM_WRITE_MULTIPLE,
+  Paths.WM_DYN_ARRAY_LENGTH,
+];
```
piwonskp, line 109: :ok_hand:

tests/behaviour/expectations/behaviour.ts
piwonskp, line 1: The same problem as #1055. The diff is large due to the change of indentation

warplib/src/maths/utils.cairo
piwonskp: `try_` means the function may panic. `try` are always functions where bits(sourceType) < bits(targetType)

piwonskp: `u32` don't need to have `try` in name since it will always succeed

src/cairoUtilFuncGen/memory/implicitConversion.ts
```diff
@@ -357,33 +338,30 @@ export class MemoryImplicitConversionGen extends StringIndexedFuncGen {
 
     const memoryWrite = this.memoryWrite.getOrCreateFuncDef(targetType.elementT);
     const targetCopyCode = [
-      `let (target_elem_loc) = wm_index_dyn(target, index, ${uint256(targetTWidth)});`,
+      `let target_elem_loc = warp_memory.index_dyn(target, index, ${targetTWidth});`,
       `${memoryWrite.name}(target_elem_loc, target_elem);`,
     ];
```
piwonskp: Endent

src/cairoUtilFuncGen/memory/memoryStruct.ts
```diff
@@ -79,38 +79,33 @@ export class MemoryStructGen extends StringIndexedFuncGen {
...
-        `    return (start,);`,
-        `}`,
+        endent`
+        #[implicits(warp_memory: WarpMemory)]
+        fn ${funcName}(${argString}) -> felt252 {
+            let start = warp_memory.unsafe_alloc(${structType.width});
+            ${writeStructCode}
+            return (start,);
+        }`,
       ].join('\n'),
```
piwonskp: Unnecessary join

src/cairoUtilFuncGen/memory/memoryToCalldata.ts
```diff
@@ -119,14 +112,10 @@ export class MemoryToCallDataGen extends StringIndexedFuncGen {
     return {
       name: funcName,
       code: endent`
-        #[implicit(warp_memory)]
-        func ${funcName}(mem_loc : felt) -> (ret_data: ${outputType.toString()}){
-          alloc_locals;
+        #[implicit(warp_memory: WarpMemory)]
+        fn ${funcName}(mem_loc : felt252) -> ${outputType.toString()} {
```
piwonskp:
```suggestion
        fn ${funcName}(mem_loc: felt252) -> ${outputType.toString()} {
```

src/cairoUtilFuncGen/memory/staticIndexAccess.ts
```diff
@@ -30,20 +34,28 @@ export class MemoryStaticArrayIndexAccessGen extends CairoUtilFuncGenBase {
...
     const width = CairoType.fromSol(arrayType.elementT, this.ast).width;
     return createCallToFunction(
       importFunc,
       [
         indexAccess.vBaseExpression,
-        indexAccess.vIndexExpression,
-        createNumberLiteral(width, this.ast, 'uint256'),
-        createNumberLiteral(arrayType.size, this.ast, 'uint256'),
+        createCallToFunction(u256ToFelt252, [indexAccess.vIndexExpression], this.ast),
+        createNumberLiteral(width, this.ast, 'uint252'),
```
piwonskp: A constant with a comment what `uint252` exactly is and why it was introduced would be useful. I've seen you've added functions like createtypename and it's good. However since it's unintuitive and there is also `felt252` it should be explained why it is needed and also relation between the two

src/io.ts
rodrodros: Because it is not arm friendly

src/utils/importPaths.ts
```diff
@@ -251,11 +242,60 @@ export const UNSAFE_CONTRACT_ADDRESS_FROM_U256: [string[], string] = [
...
+export const WM_WRITE_MULTIPLE: [string[], string] = [
+  MULTICELL_ACCESSOR_TRAIT,
+  'warp_memory.write_multiple',
+];
+export const WM_READ_MULTIPLE: [string[], string] = [
+  MULTICELL_ACCESSOR_TRAIT,
+  'warp_memory.read_multiple',
+];
+
+const ACCESSOR_TRAIT: string[] = [...MEMORY_MODULE, 'accessors', 'WarpMemoryAccesssorTrait'];
```
temyurchenko: Triple "s" in "acce_sss_or" probably needs to be changed

src/cairoWriter/writers/cairoImportFunctionDefinitionWriter.ts
rodrodros: Functions inside traits are not imported directly. The path is cut short at the name. Without the condition we would import `import::path::Trait::funcName` which is wrong. We should only use  `import::path::Trait`.

src/cli.ts
rodrodros: Because I am not sure if we are going to need it or not, and leaving it there as a bread crumb trail in case it was needed further down the line

temyurchenko:
https://github.com/NethermindEth/warp/pull/1045#discussion_r1208123402

I've made a search for functions that start with "try_" in the Cairo codebase and in all cases it's functions that might return a failure (`Option`), but not panic. 
Since this function may panic, it might probably make sense to prefix it with "unsafe". 

cc @piwonskp @rodrigo-pino 


(for some reason, I can't reply in the thread itself)

src/utils/importPaths.ts
```diff
@@ -291,7 +291,7 @@ export const WM_READ_MULTIPLE: [string[], string] = [
   'warp_memory.read_multiple',
 ];
 
-const ACCESSOR_TRAIT: string[] = [...MEMORY_MODULE, 'accessors', 'WarpMemoryAccesssorTrait'];
+const ACCESSOR_TRAIT: string[] = [...MEMORY_MODULE, 'accessors', 'WarpMemoryAccessorTrait'];
```
temyurchenko, line 294: You have to fix it in all of the Cairo files as well(

rodrodros: Done, thanks!

src/cairoWriter/writers/cairoImportFunctionDefinitionWriter.ts
piwonskp: Oh I see, let's explain it in the comment, I suppose it's only a temporary fix

warplib/src/maths/utils.cairo
piwonskp: > [#1045 (comment)](https://github.com/NethermindEth/warp/pull/1045#discussion_r1208123402)
> 
> I've made a search for functions that start with "try_" in the Cairo codebase and in all cases it's functions that might return a failure (`Option`), but not panic. Since this function may panic, it might probably make sense to prefix it with "unsafe".
> 
> cc @piwonskp @rodrigo-pino
> 
> (for some reason, I can't reply in the thread itself)

I see, you're right. There's barely any function using panic in the corelib. The convention is established for functions returning Option

src/utils/nodeTemplates.ts
```diff
@@ -32,6 +32,8 @@ import { safeGetNodeTypeInCtx, specializeType } from './nodeTypeProcessing';
 import { notNull } from './typeConstructs';
 import { toHexString, toSingleExpression } from './utils';
 
+const FELT_TYPESTRING = 'uint252';
```
piwonskp, line 41: Let's explain why it can't be e.g. `felt252` in comment

.starknet_accounts_devnet/starknet_open_zeppelin_accounts.json
piwonskp, line 1: Maybe let's add the file to gitignore. The file gets overwritten each time behaviour tests are run thus your prs gonna have a conflict https://github.com/NethermindEth/warp/pull/1043/files#diff-ea8e560f656c39d48f9e28f674d13ed2a8c8435a876cc5c72322e55f1bb7509a

piwonskp: Actually I'll try to fix it in another pr

piwonskp, line 1: Please revert the changes to the file

src/cairoUtilFuncGen/memory/implicitConversion.ts
piwonskp: Imo we would be able to move most, if not all, of the conversion code to static traits (or reuse existing from corelib). I'm not familiar with the `generateScalingCode` specifically though

piwonskp (approved):
Apart from the few simple changes (deleting starknet accounts, adding a few comments, indentation) it's ready to be merged

src/cairoUtilFuncGen/memory/implicitConversion.ts
rodrodros: Yes, we probably can. That is what I meant

tests/behaviour/expectations/behaviour.ts
temyurchenko, line 1: Can we revert everything in this file? You will have to temporarily (locally) disable the commit hook for that

temyurchenko (approved):

Issue #1046, opened 2023-05-09
Research and implement runtime tests for cairo 1


piwonskp:
Tests need a major rework, the solution implemented right now is temporary. 

The changes  necessary:
- [ ] Use a CLI that provides a machine readable output (e.g. protostar)
- [ ] Change the output of warp to be compatible with the newly chosen CLI
- [x] #1052
- [ ] Remove flask server and all leftovers of the old solution
- [x] #1053
- [ ] Fix linting errors
- [ ] Research https://github.com/dojoengine/katana

PR #1047, opened 2023-05-09
Implement behaviour testing for cairo 1.0

piwonskp:
This PR provides a temporary solution to run behaviour tests for cairo 1.0. 

For more details see #1046.


tests/testnetInterface.ts
```diff
@@ -102,33 +132,42 @@ export async function invoke(
...
+      status: 500,
+      steps: null,
+      threw: true,
+      return_data: null,
+      error_message: result,
+      events: [],
+    };
+  }
+  const transactionAsBuffer = await execSync(
+    `starknet get_transaction_trace --gateway_url http://devnet:5050 --feeder_gateway_url http://devnet:5050 --hash ${matches[1]}`,
```
cicr99: Why not use `starknetCliCall` here?

tests/testnetInterface.ts
```diff
@@ -102,33 +132,42 @@ export async function invoke(
...
+    steps: null,
+    threw: false,
+    return_data: transaction.function_invocation.result,
+    error_message: undefined,
+    events: transaction.function_invocation.events,
+  };
 }
 
 export async function ensureTestnetContactable(timeout: number): Promise<boolean> {
+  return Promise.resolve(true);
```
cicr99, line 170: What are we doing with the rest of the code in this function?

tests/testnetInterface.ts
piwonskp: Good catch

tests/testnetInterface.ts
piwonskp: I left it for later, I think the logic of verifying server (devnet in this case) availability might be needed. I'll comment that tho to make it clear it's not used right now

tests/behaviour/behaviour.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
rodrodros, line 1: I think there is no need to disable `eslint` in this specific file

piwonskp: There is unused import `compileCluster` that breaks  linting. Currently the function is replaced with `compileCairo1`, but I'd leave the decision on what to do with it for a PR refactoring tests. The function should be either removed or updated to work with cairo 1

piwonskp: I think it's better to leave a trace that such a function was used in the code so when fixing linting it'll have to be dealt with, not just silently ignored

rodrodros (approved):

cicr99 (approved):

PR #1051, opened 2023-05-17
Cairo1 bump up to v1.0.0-rc0

AlejandroLabourdette:


piwonskp:
The PR requires https://github.com/0xSpaceShard/starknet-devnet/pull/472 and an update to dependencies before it gets merged

piwonskp:
It's merged now, let's update dependencies for starknet devnet

piwonskp:
@rodrigo-pino do you think we could use `corelib` bundled by `cairo-lang` from this point on?

rodrodros:
I've added the new Warp binary for 1.0.0-rc for arm. Also updated warplib code to be compatible with new version. As soon as someone updates for x86_64, test should start passing

temyurchenko (approved):
Everything looks in order to me, however I want to learn more about the approaches taken.

Why do we keep two versions of the corelib for different architectures? What is the difference? 
Have we considered the alternative of having a submodule pointing to a particular tag in the StarkWare repository?
Are we set on keeping the binaries in the repo long-term?

piwonskp (approved):

piwonskp:
> Everything looks in order to me, however I want to learn more about the approaches taken.
> 
> Why do we keep two versions of the corelib for different architectures? What is the difference? Have we considered the alternative of having a submodule pointing to a particular tag in the StarkWare repository? Are we set on keeping the binaries in the repo long-term?

We plan to use normal package management long-term, either `cairo-lang` from pip or cairo compiler from cargo.  Probably it will be cargo long term

piwonskp:
I've made an issue for that https://github.com/NethermindEth/warp/issues/1068

temyurchenko:
> I've made an issue for that #1068

I see, thank you. Is that also the reason that we duplicate the library? No reason not too, given that this is temporary anyway?

piwonskp:
Exactly, I guess it was easier that way

temyurchenko:
> Exactly, I guess it was easier that way

I see the point, but just as a note one symlink would serve well to lessen the code/changes blowup

PR #1054, opened 2023-05-17
CI to build and push docker image on release

### LINKED ISSUE #1050: CI to push image to dockerhub

cicr99:
Fixes #1050


.github/workflows/build-push-docker-image.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Docker image
+
+on:
+  release:
+    types: [published]
```
piwonskp: Shall we change it also in `publish package to npmjs`? Or maybe we should have one workflow `publish.yaml` with two jobs?

.github/workflows/build-push-docker-image.yml
```diff
@@ -0,0 +1,36 @@
...
+        shell: bash
+        run: echo "TAG_FROM_REPO=$(git describe --tags --always | cut -d- -f1)" >> $GITHUB_OUTPUT
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: nethermind/warp:${{ steps.get_version.outputs.TAG_FROM_REPO }}
```
piwonskp: I think we don't need the release version. Do you think we could use `tag_name`?
```suggestion
      - name: Build and push Docker image
        uses: docker/build-push-action@v4
        with:
          context: .
          file: ./Dockerfile
          push: true
          tags: nethermind/warp:${{ github.event.release.tag_name }}
```

.github/workflows/build-push-docker-image.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Get release version
+        id: get_version
+        shell: bash
+        run: echo "TAG_FROM_REPO=$(git describe --tags --always | cut -d- -f1)" >> $GITHUB_OUTPUT
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ./Dockerfile
```
piwonskp: I think the context and file is not needed since they're defaults

.github/workflows/build-push-docker-image.yml
cicr99: Yes, it should probably be changed to `published` as well. I decided to use this type based on the [notes](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) of the documentation.
As both workflows will be triggered on the same event, I guess there's no problem on having them as two jobs under the same workflow. I'll do that

.github/workflows/build-push-docker-image.yml
cicr99: Is it possible that `tag_name` is an empty string? I thought to get the release version to have a different image after every new release

.github/workflows/build-push-docker-image.yml
piwonskp: Yea, I think it is a better option since they do the same, just to different repositories

.github/workflows/build-push-docker-image.yml
piwonskp: `tag_name` should be whatever you set when creating a release in github UI. I don't think it can be empty, it wouldn't make sense to create a release with a blank version number

temyurchenko (approved):
Seems perfectly reasonable, but I'm not particularly an expert on Docker infrastructure

.github/workflows/publish.yaml
```diff
@@ -20,3 +20,26 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_RELEASE_KEY }}
+
+  build_and_push_docker_image:
```
piwonskp:
```suggestion
  dockerhub:
```

.github/workflows/publish.yaml
```diff
@@ -1,10 +1,10 @@
-name: Publish Package to npmjs
+name: Publish Package to npmjs and Docker image
 on:
   release:
-    types: [created]
+    types: [published]
 
 jobs:
-  build:
+  build_package:
```
piwonskp:
```suggestion
  npm:
```

.github/workflows/publish.yaml
```diff
@@ -20,3 +20,26 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_RELEASE_KEY }}
+
+  build_and_push_docker_image:
+    name: Push Docker image to Docker Hub
+    if: github.repository_owner == 'NethermindEth'
```
piwonskp, line 26: Why do we need the condition?

temyurchenko: I would assume, so that forks don't try to publish?

piwonskp: They shouldn't be able anyway since they won't have the access to secrets

piwonskp: If they could somehow it would be a serious security breach

temyurchenko: I think the point is that their CI doesn't even try to do that. If it does, it will result in extra-work, a failure and paint CI red. So, we're just making life a bit less confusing for contributors. At least, that is the reason for why I did stuff like that before. In that case, let's wait for Carmen's response

cicr99: Yes, I think @murcake explained perfectly, even if this workflow won't work for other repos because of the secrets, it shouldn't run for them at all.
When I implemented this workflow I also had a look into [nethermind](https://github.com/NethermindEth/nethermind/blob/master/.github/workflows/build-nethermind-docker-images.yml) and [juno](https://github.com/NethermindEth/juno/blob/main/.github/workflows/docker-image-build-push.yml) repositories as they both have an implementation of it, and both of them have the condition as well. I thought it was a very good idea to do the check.

piwonskp: I wouldn't mind painting CI red in the fork since trying to publish a release on a fork would be incomprehensible. I prefer if such code is introduced by contributors based on their real-world use case, not arbitrarily decided up front just in case. By providing the use case or reasoning it is easy to asses if the change is justified. Otherwise we don't know it the code is useful at all

Having that said it's not a big deal so I'm fine with merging it with the change

piwonskp (approved):

PR #1055, opened 2023-05-17
Check for passing behavior tests

AlejandroLabourdette:


AlejandroLabourdette (commented):
address this issue #1052

tests/behaviour/expectations/behaviour.ts
```diff
@@ -25,4784 +25,4758 @@ export const expectations = flatten(
...
-//               ['balance', [], ['338', '0'], '0'],
-//               ['closeEvent', [], [], '0'],
-//               [
-//                 'donate',
-//                 ['500', '0'],
-//                 null,
-//                 '0',
-//                 'The event must be open in order to receive donations',
-//               ],
-//
```
piwonskp: Did you try changing plain arrays to `toCairoUint256(x)` to test if constructor tests work?

tests/behaviour/expectations/behaviour.ts
```diff
@@ -25,4784 +25,4758 @@ export const expectations = flatten(
           Expect.Simple('test', [], []),
           Expect.Simple('returnTest', [], toCairoUint256(12)),
         ]),
```
piwonskp: Let's remove the example test. There was a duplicate somewhere below so let's use the other one

.github/workflows/test-behaviour.yml
```diff
@@ -118,6 +118,47 @@ jobs:
...
+      - name: Test code execution
+        run: |
+          starknet-devnet &
+          yarn test:behaviour
+
+      - name: Save benchmarks in artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: benchmark
+          path: benchmark
```
piwonskp: I suppose that is used only for `yarn testnet:benchmark`?

tests/behaviour/expectations/behaviour.ts
AlejandroLabourdette: that function do same, diference is that it create the hex representation. Tests keep failing

.github/workflows/test-behaviour.yml
```diff
@@ -118,6 +118,45 @@ jobs:
           name: benchmark
           path: benchmark
 
+  build-cairo1:
```
piwonskp:
```suggestion
  test-behaviour:
```

.github/workflows/test-behaviour.yml
```diff
@@ -118,6 +118,45 @@ jobs:
...
+    # only runs on pull requests to the cairo-1.0 branch
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref == 'cairo-1.0' }}
```
piwonskp: It's unneeded since the yaml will be only present on `cairo-1.0` branch

.github/workflows/test-behaviour.yml
```diff
@@ -118,6 +118,45 @@ jobs:
...
+        with:
+          python-version: 3.9 # Cairo was tested with this version
+
+      - name: Install dependencies
+        run: |
+          make
+          make compile
+
+      - name: Setup warp_venv
+        run: bin/warp install --python python --verbose
```
temyurchenko: Why do we need this step?

tests/behaviour/behaviour.test.ts
```diff
@@ -33,10 +33,11 @@ interface AsyncTestCluster {
...
 }
 
-fs.writeFileSync(
-  path.join(BASE_PATH, '.starknet_accounts_devnet', 'starknet_open_zeppelin_accounts.json'),
-  '{}',
-);
+const accountDir = '.starknet_accounts_devnet';
+if (!fs.existsSync(accountDir)) {
+  fs.mkdirSync(accountDir);
+}
```
temyurchenko:
```suggestion
fs.mkdirSync(accountDir, { recursive: true })
```

temyurchenko (commented):
Can we remove tests that don't pass at the moment, or move them to a different file? In the first approach, we can resurrect fixed tests from older revisions. As it is now, it's hard to see what has been uncommented and is working and the diff blowup is impressive

piwonskp:
> Can we remove tests that don't pass at the moment, or move them to a different file? In the first approach, we can resurrect fixed tests from older revisions. As it is now, it's hard to see what has been uncommented and is working and the diff blowup is impressive

I think the huge diff is an issue of indentation change, or sth like that in the PR. The diff should only include uncommented tests so it shouldn't be that huge

temyurchenko:
> > Can we remove tests that don't pass at the moment, or move them to a different file? In the first approach, we can resurrect fixed tests from older revisions. As it is now, it's hard to see what has been uncommented and is working and the diff blowup is impressive
> 
> I think the huge diff is an issue of indentation change, or sth like that in the PR. The diff should only include uncommented tests so it shouldn't be that huge

I see. Then the indentation change should preferably be fixed, otherwise it is genuinely hard to navigate thousands of lines of diffs

piwonskp:
Yea I think what has changed is the indentation of `//` comments, previously they weren't indented, now they are

temyurchenko:
I tried to reduce the effect of comments in #1073 as much as possible. I only touched places where the only change is placement of comments. There were some reorderings and renamings that could probably also be reverted, but I didn't touch them, I don't understand it that well.

.github/workflows/test-behaviour.yml
cicr99: Yes, I agree. I would also remove the `build` job (which is before this one), I don't see the point on having the old version here as well. Once `cairo-1.0` branch is moved to `develop`, that job would be deprecated, I mean, you would only run the one you wrote for this branch

.github/workflows/test-behaviour.yml
cicr99, line 1: At the beginning of the file (lines 15-16) there is:
```
env:
  CAIRO_LANG_VERSION: 0.10.3
```
Shouldn't the version be changed?

tests/behaviour/expectations/behaviour.ts
cicr99, line 1: Couldn't we have something like with the compilation tests? I mean, instead of having this file commented, we could leave this one as it is, and create a new file `passingContracts` the same as [this one](https://github.com/NethermindEth/warp/blob/cairo-1.0/tests/compilation/passingContracts.ts).
When you collect all of the files to be tested you would have to add a filter to remove those that are commented in this file. I think this would help wit readability and it would make the process of uncommenting a lot easier.

tests/behaviour/behaviour.test.ts
AlejandroLabourdette: It's only one folder to create, seems better to use a more restrict approach

tests/behaviour/expectations/behaviour.ts
AlejandroLabourdette: will be solved in a future pr

tests/behaviour/behaviour.test.ts
temyurchenko: Unfortunately, FS is a naturally parallelly-accessed system, so checks like that are prone to data races. It might not be to much of a concern in CI at this moment, but why pessimize)

piwonskp (approved):

tests/behaviour/expectations/behaviour.ts
cicr99: I added an issue for this #1074

cicr99 (approved):

PR #1041, opened 2023-04-26
replace function calls by operators and removed imports and input checks

esdras-santos:
This  PR do three things:
1 - deal with unsafe operations inside unchecked blocks.
2 - remove unnecessary warplib functions and replace it for their respective operations.
3 - remove input checks.

given the following function:

```
function add(uint x, uint y) external pure returns (uint) {
    unchecked {
        return x + y;
    }
}
```

old transpilation:

```
#[view]
fn add_771602f7(__warp_0_x : u256, __warp_1_y : u256)-> u256{
    let __warp_se_0 = warp_add_unsafe256(__warp_0_x, __warp_1_y);
    return __warp_se_0;
}
```

new transpilation:

```
#[view]
fn add_771602f7(__warp_0_x : u256, __warp_1_y : u256)-> u256{
        let (__warp_se_0, overflow) = u256_overflowing_add(__warp_0_x, __warp_1_y);
        assert(overflow == false, 'Overflow in unchecked block');
        return __warp_se_0;
}
```

new transpilation for when the operation is bellow than 256 bits

```
#[view]
fn add_feb99390(__warp_0_x : u128, __warp_1_y : u128)-> u128{
        let (__warp_se_0, overflow) = matchu128_overflowing_add(__warp_0_x, __warp_1_y){
            Result::Ok(__warp_se_0) => (__warp_se_0, false),
            Result::Err(__warp_se_0) => (__warp_se_0, true),
        };
        assert(overflow == false, 'Overflow in unchecked block');
        return __warp_se_0;
}
```

In the case of an operation outside an unchecked block like the following:

```
function add(uint8 x, uint8 y)public returns(uint8){
        return x + y;
}
```

old transpilation:

```
------------imports---------------
use warplib::maths::add::warp_add8;
use warplib::maths::external_input_check_ints::warp_external_input_check_int8;

--------------------------------------

#[external]
fn add_bb4e3f4d(__warp_0_x : felt252, __warp_1_y : felt252)-> felt252{

    
    warp_external_input_check_int8(__warp_1_y);
    
    warp_external_input_check_int8(__warp_0_x);
    
    let __warp_se_0 = warp_add8(__warp_0_x, __warp_1_y);
    
    
    return __warp_se_0;
}
```
new transpilation:
```
------------No warplib imports-------------------------
 
#[external]
fn add_bb4e3f4d(__warp_0_x : felt252, __warp_1_y : felt252)-> felt252{
    
    return __warp_0_x + __warp_1_y;
}
```

PR #1057, opened 2023-05-17
replaced CairoAssert by FunctionCall

esdras-santos:
replaced CairoAssert node by FunctionCall.


src/passes/builtinHandler/require.ts
```diff
@@ -50,33 +48,29 @@ export class Require extends ASTMapper {
...
 
     if (expression.vIdentifier === 'require' || expression.vIdentifier === 'assert') {
-      const requireMessage =
-        expression.vArguments[1] instanceof Literal ? expression.vArguments[1].value : null;
-
       return new ExpressionStatement(
         ast.reserveId(),
         expression.src,
-        new CairoAssert(
+        new FunctionCall(
```
cicr99: This new FunctionCall node is exactly the same as `expression` but with a new id; instead of creating this node manually consider using `cloneASTNode(expression, ast)`

src/passes/builtinHandler/require.ts
```diff
@@ -50,33 +48,29 @@ export class Require extends ASTMapper {
...
           ast.reserveId(),
           expression.src,
+          expression.typeString,
+          expression.kind,
           createBoolLiteral(false, ast),
```
cicr99: This boolean literal can't be the expression of the Function Call node, this argument is supposed to be the expression that defines the callee, in this case it should be an Identifier node "revert". You can just pass `expression.vExpression` for that as you did in the previous case.

You still need to add `false` as the first parameter of this function though, so it gets printed correctly when it gets to the writer. For that you will need to modify the list of arguments you pass to create this FunctionCall node. Instead of passing `expression.vArguments`, it could be:
 `[createBoolLiteral(false, ast), ...expression.vArguments]`

src/utils/astChecking.ts
cicr99, line 1: Comment on line 74. You can remove CairoAssert from there

cicr99 (changes requested):

src/passes/builtinHandler/require.ts
cicr99: You should also add some examples to compilation tests to check the transpilation was correct.

src/passes/references/expectedLocationAnalyser.ts
```diff
@@ -320,7 +319,7 @@ export class ExpectedLocationAnalyser extends ASTMapper {
     this.visitStatement(node, ast);
   }
 
-  visitCairoAssert(node: CairoAssert, ast: AST): void {
```
piwonskp, line 323: You've removed `visitCairoAssert` in `visitor.ts` so the function won't be called

tests/compilation/contracts/errorHandling/assertHandling.sol
cicr99, line 1: The name of the file was changed in order for it to successfully compile. Refer to this issue for further details https://github.com/NethermindEth/warp/issues/1086

tests/compilation/passingContracts.ts
```diff
@@ -34,9 +34,9 @@ export const passingContracts = [
   // 'tests/compilation/contracts/deleteUses.sol',
   'tests/compilation/contracts/enums.sol',
   // 'tests/compilation/contracts/enums7.sol',
-  // 'tests/compilation/contracts/errorHandling/assert.sol',
-  // 'tests/compilation/contracts/errorHandling/require.sol',
-  // 'tests/compilation/contracts/errorHandling/revert.sol',
+  'tests/compilation/contracts/errorHandling/assertHandling.sol',
+  'tests/compilation/contracts/errorHandling/requireHandling.sol',
+  'tests/compilation/contracts/errorHandling/revertHandling.sol',
```
cicr99: This test is failing due to errors related to input checks which were being solved in a separate PR https://github.com/NethermindEth/warp/pull/1041
Let's comment this test case here and uncomment it once the error is handled

cicr99 (approved):

src/cairoWriter/writers/functionCallWriter.ts
```diff
@@ -29,7 +29,9 @@ export class FunctionCallWriter extends CairoASTNodeWriter {
     const func = writer.write(node.vExpression);
     switch (node.kind) {
       case FunctionCallKind.FunctionCall: {
-        if (node.vExpression instanceof MemberAccess) {
+        if (['assert', 'require', 'revert'].includes(func)) {
+          return [`assert(${args}${node.vArguments.length === 1 ? ", 'Assertion error'" : ''})`];
```
piwonskp: Imo it should be done on the AST level. Could we create assert function call in builtin handler instead?

piwonskp (approved):
Minor issue of the place where arguments are handled and functions like require are transformed into `assert`. Apart from that lgtm

PR #1060, opened 2023-05-18
fix to-address conversion issues

temyurchenko:
- conversions from int variables used to call a Cairo 0 function. It
   was replaced with a new, Cairo 1,
   function---`unsafe_contract_address_from_felt`.

- conversions from int literals were represented in Cairo as raw int
   literals, which fails the type system.

- conversions from int literals to address wouldn't trigger typeString
   change from 'int_const bla-bla' to 'address', which caused
   primitiveTypeToCairo to fail.

- added tests

- the regex had to be changed, as it accidentally was catching methods
   of traits. Anyway, we might need to consider a solution not
   involving regexes, although it's not that bad, given that we only
   impose it on the code written by us


src/passes/builtinHandler/explicitConversionToFunc.ts
```diff
@@ -99,24 +99,31 @@ export class ExplicitConversionToFunc extends ASTMapper {
...
           ),
-          [node.vArguments[0]],
+          [operand],
           ast,
         );
         ast.replaceNode(node, replacementCall);
       } else {
-        ast.replaceNode(node, node.vArguments[0]);
+        throw new TranspileFailedError(
+          `Don't know how to convert ${argType.constructor.name} to address`,
```
rjnrohit: Unknown type `${argType.constructor.name}` received for addressType conversion

warplib/src/conversions/integer_conversions.cairo
```diff
@@ -33,3 +36,34 @@ fn bool_into_felt252(val: bool) -> felt252 {
...
+                return Option::None(());
+            }
+        }
+        Option::Some(
+            self.high.into() * 0x100000000000000000000000000000000_felt252 + self.low.into()
+        )
+    }
+}
+
+fn unsafe_contract_address_from_u256(x: u256) -> ContractAddress {
```
rjnrohit, line 63: can keep this function in the `utils.cairo` file

rjnrohit (approved):
lgtm

warplib/src/conversions/integer_conversions.cairo
```diff
@@ -33,3 +36,34 @@ fn bool_into_felt252(val: bool) -> felt252 {
...
+    match x.try_into() {
```
rjnrohit, line 64: `u256_to_felt252` is already defined in `utils.cairo` file

src/passes/builtinHandler/explicitConversionToFunc.ts
temyurchenko: Oh god, took me a minute to understand that you weren't describing a warp error :D 

Will change

warplib/src/conversions/integer_conversions.cairo
temyurchenko: What is the distinction logic? It seems like integer conversion. Well, definitely from integer, and arguably into integer.
I'm asking to make a better choice in the future

warplib/src/conversions/integer_conversions.cairo
temyurchenko: Which `utils.cairo` are we talking about? The only `utils.cairo` I see is `warplib/src/maths/utils.cairo` and it's Cairo 0

temyurchenko: I did a global search of the project with ripgrep and there is no such function, it seems?

rjnrohit: sorry I got confused with https://github.com/NethermindEth/warp/blob/5a0728efab84613fb06d546cab82129db927258d/warplib/src/maths/utils.cairo#L70

since it is in the pipeline

warplib/src/conversions/integer_conversions.cairo
rjnrohit: Sorry again confused with the file https://github.com/NethermindEth/warp/blob/5a0728efab84613fb06d546cab82129db927258d/warplib/src/maths/utils.cairo

piwonskp:
Is there a behaviour test we could uncomment after implementing the PR?

temyurchenko:
> Is there a behaviour test we could uncomment after implementing the PR?

I'm not sure, especially given that I was told we don't use behaviour tests at this very moment.

temyurchenko:
> Is there a behaviour test we could uncomment after implementing the PR?

What I'm saying is, it might be better to do once we enable such tests in PR, because right now we have no way of tracking regressions. Once we do, we can uncomment all tests, see which pass, and leave those uncommented. Are you okay with such an approach?

warplib/src/conversions/integer_conversions.cairo
```diff
@@ -36,3 +39,34 @@ fn bool_into_felt252(val: bool) -> felt252 {
         0
     }
 }
+
+// Cairo-1.0 corelib code from the RC
```
piwonskp: RC is now merged, could it be deleted then?

temyurchenko: Oh, not that RC :( 1.1 RC

temyurchenko: Oh wait, maybe that one! Let me check once again

warplib/src/conversions/integer_conversions.cairo
```diff
@@ -36,3 +39,34 @@ fn bool_into_felt252(val: bool) -> felt252 {
...
+
+fn unsafe_contract_address_from_u256(x: u256) -> ContractAddress {
+    match x.try_into() {
+        Option::Some(felt_x) =>
+            match starknet::contract_address_try_from_felt252(felt_x) {
+                Option::Some(address) => address,
+                Option::None(_) =>
+                    panic_with_felt252('the felt cannot be an address'),
+            },
+        Option::None(_) => panic_with_felt252('the felt cannot be an address'),
```
piwonskp: The error message is misleading, the problem is the conversion `u256 -> felt` here, not `felt -> address`

warplib/src/conversions/integer_conversions.cairo
temyurchenko: Yeah, not that one( I'll change the version to 1.1

warplib/src/conversions/integer_conversions.cairo
temyurchenko: Absolutely right! Fixed this

warplib/src/conversions/integer_conversions.cairo
piwonskp: Please add `TODO: remove [...]` just to be explicit

temyurchenko: Here you go

piwonskp (approved):

PR #1070, opened 2023-05-23
translate int conversions to Cairo 1

temyurchenko:
Instead of generating Cairo functions, we use generic capabilities of
   Cairo 1 to write a single static function in warplib.

Additionally we remove some usages of old functions, but not all of
   them, since they are tied to much into some subsystems, like
   calldata and storage. This will be finished in future work.

### DIFF
--- a/src/passes/cairoUtilImporter.ts
+++ b/src/passes/cairoUtilImporter.ts
@@ -1,6 +1,7 @@
 import {
   AddressType,
   ElementaryTypeName,
+  FunctionCall,
   IntType,
   Literal,
   MemberAccess,
@@ -12,7 +13,14 @@ import {
 import { AST } from '../ast/ast';
 import { ASTMapper } from '../ast/mapper';
 import { createImport } from '../utils/importFuncGenerator';
-import { INTO, U256_FROM_FELTS, U128_FROM_FELT, CONTRACT_ADDRESS } from '../utils/importPaths';
+import {
+  INTO,
+  U256_FROM_FELTS,
+  U128_FROM_FELT,
+  CONTRACT_ADDRESS,
+  CUTOFF_DOWNCAST,
+  WARPLIB_INTEGER,
+} from '../utils/importPaths';
 import { safeGetNodeType } from '../utils/nodeTypeProcessing';
 import { getContainingSourceUnit, primitiveTypeToCairo } from '../utils/utils';
 
@@ -39,6 +47,17 @@ export class CairoUtilImporter extends ASTMapper {
     } else if (cairoType === 'ContractAddress') {
       createImport(...CONTRACT_ADDRESS, this.dummySourceUnit ?? node, ast);
     }
+  }
+
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    // FIXME: remove when BitAnd is available for integer types in
+    // corelib
+    if (node.vFunctionName === CUTOFF_DOWNCAST[1]) {
+      const path = WARPLIB_INTEGER.slice(0, -1);
+      const name = WARPLIB_INTEGER.at(-1)!;
+      createImport(path, name, this.dummySourceUnit ?? node, ast);
+    }
+    super.visitFunctionCall(node, ast);
   }
 
   visitLiteral(node: Literal, ast: AST): void {
--- a/src/passes/cairoUtilImporter.ts
+++ b/src/passes/cairoUtilImporter.ts
@@ -12,6 +12,7 @@ import {
 } from 'solc-typed-ast';
 import { AST } from '../ast/ast';
 import { ASTMapper } from '../ast/mapper';
+import { requireNonNullish } from '../export';
 import { createImport } from '../utils/importFuncGenerator';
 import {
   INTO,
@@ -54,7 +55,7 @@ export class CairoUtilImporter extends ASTMapper {
     // corelib
     if (node.vFunctionName === CUTOFF_DOWNCAST[1]) {
       const path = WARPLIB_INTEGER.slice(0, -1);
-      const name = WARPLIB_INTEGER.at(-1)!;
+      const name = requireNonNullish(WARPLIB_INTEGER.at(-1));
       createImport(path, name, this.dummySourceUnit ?? node, ast);
     }
     super.visitFunctionCall(node, ast);
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -82,3 +82,13 @@ export interface ExecSyncError {
 export function instanceOfExecSyncError(object: any): object is ExecSyncError {
   return 'stderr' in object;
 }
+
+export function requireNonNullish<T>(x: T | null | undefined): T {
+  if (x === null) {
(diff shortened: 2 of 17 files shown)


temyurchenko:
I see two warnings of this kind: https://typescript-eslint.io/rules/no-non-null-assertion/. 

The non-null-check operator "!" is used in two, seemingly to me, perfectly reasonable places. It is basically equivalent to saying «the max int type is 256 bytes wide». Furthermore, I can see this operator being useful in other static contexts, like accessing elements of a static array. 
It might be reasonable to remove this option from eslint, and check validity of using such an operator on a case-by-case basis.

src/warplib/implementations/conversions/int.ts
```diff
@@ -116,19 +25,24 @@ export function functionaliseIntConversion(conversion: FunctionCall, ast: AST):
...
-    IntFunction(conversion, conversion.vArguments[0], 'uint', 'int_conversions', ast);
-    return;
-  } else if (
-    fromType.nBits === toType.nBits ||
-    (fromType.nBits < toType.nBits && !fromType.signed && !toType.signed)
-  ) {
-    arg.typeString = conversion.typeString;
-    ast.replaceNode(conversion, arg);
+  const fromCairoWidth = cairoWidths.find((x) => x >= fromType.nBits)!;
+  const toCairoWidth = cairoWidths.find((x) => x >= toType.nBits)!;
```
rodrodros: I understand what you say about the `!` operator. Nonetheless you're assuming that both `cairoWidths` and the `fromType.nBits` will be always correct, something that may not be true.

I think it would be a good idea also to add a failing compilation test for the unsupported `nBits`

src/passes/cairoUtilImporter.ts
```diff
@@ -41,6 +49,17 @@ export class CairoUtilImporter extends ASTMapper {
     }
   }
 
+  visitFunctionCall(node: FunctionCall, ast: AST): void {
+    // FIXME: remove when BitAnd is available for integer types in
+    // corelib
+    if (node.vFunctionName === CUTOFF_DOWNCAST[1]) {
+      const path = WARPLIB_INTEGER.slice(0, -1);
+      const name = WARPLIB_INTEGER.at(-1)!;
```
rodrodros: The `!` can be misused, I would advise against it in every case even if it means writing a little more code. Even if it's safe here (and makes sense), I would prefer a different handling that wouldn't require it.

I believe there is a general risk of using `!` through out the code, JavaScript is unsafe enough for us to start adding the possibility of  human error during Typescript compilation

src/warplib/generateWarplib.ts
```diff
@@ -62,14 +61,13 @@ const mathWarplibFunctions: WarplibFunctionInfo[] = [
...
 ];
-const conversionWarplibFunctions: WarplibFunctionInfo[] = [int_conversions()];
 const inputCheckWarplibFunctions: WarplibFunctionInfo[] = [
   external_input_check_ints(),
   // external_input_check_address - handwritten
 ];
 
 generateWarplibFor('maths', mathWarplibFunctions);
-generateWarplibFor('conversions', conversionWarplibFunctions);
+generateWarplibFor('conversions', []);
```
rodrodros, line 70: I think it's safe to delete this line

rodrodros (commented):
Great work! Some minor comments

src/warplib/generateWarplib.ts
temyurchenko: Unfortunately, it isn't, otherwise the `conversions.cairo` file isn't going to get generated

src/warplib/implementations/conversions/int.ts
temyurchenko: Genuinely asking, doesn't `nBits` come from `solc-typed-ast`, thus, from the solidity AST. Given that no Solidity code can have types wider than 256 bits, what test would show that?

src/passes/cairoUtilImporter.ts
temyurchenko: I suggest allowing non-null operator for checks for internal consistencies. In other words, situations where we would otherwise write `assert (x !== null)`. It doesn't include handling business logic, business logic requires explicit handling and possibly throwing an appropriate exception.
We solve misuses with reviewing.

temyurchenko: We asked the team about their opinion, but no one answered, so we are keeping the restriction on non-null operators. I've added a commit removing them

src/warplib/implementations/conversions/int.ts
rodrodros: Right now there can be `nBits` that do not exisist in cairoWidth. For example 248, 240, ... . It should throw a `NotSupportedYet` error saying that conversion from x to y is not supported yet.

temyurchenko: Oh no, it doesn't do that. It gets the least Cairo type able to fit that number. So, for u248 it's u256, for u24 it's u32

src/utils/errors.ts
```diff
@@ -82,3 +82,13 @@ export interface ExecSyncError {
...
   return 'stderr' in object;
 }
+
+export function requireNonNullish<T>(x: T | null | undefined): T {
+  if (x === null) {
+    throw new Error('Unexpected null');
+  }
+  if (x === undefined) {
+    throw new Error('Unexpected undefined');
+  }
```
piwonskp, line 92: Just out of curiosity, why not assert?

temyurchenko: This method is generic, so we don't know if `x` coming into it is an internal assumption (which deserves an assert and might be removed in the release version, automatically) or an assumption related to business logic (we can't ignore it). Thus, an error)

src/warplib/implementations/conversions/int.ts
```diff
@@ -116,19 +25,24 @@ export function functionaliseIntConversion(conversion: FunctionCall, ast: AST):
...
-    IntFunction(conversion, conversion.vArguments[0], 'uint', 'int_conversions', ast);
-    return;
-  } else if (
-    fromType.nBits === toType.nBits ||
-    (fromType.nBits < toType.nBits && !fromType.signed && !toType.signed)
-  ) {
-    arg.typeString = conversion.typeString;
-    ast.replaceNode(conversion, arg);
+  const fromCairoWidth = requireNonNullish(cairoWidths.find((x) => x >= fromType.nBits));
+  const toCairoWidth = requireNonNullish(cairoWidths.find((x) => x >= toType.nBits));
```
piwonskp, line 29: Since Cairo implements all uint sizes existing in solidity is there a reason to do it that way?

temyurchenko: It only supports uN where N = 8 * 2^i, for i in [0, 5], while Solidity supports all divisors of 8

piwonskp: True, I forgot `uint24` is a valid type in Solidity

piwonskp (approved):
lgtm

rodrodros (approved):

PR #1073, opened 2023-05-24
enable behaviour tests in CI, again

temyurchenko:


temyurchenko:
The branch was incorporated in #1055

Issue #1064, opened 2023-05-22
Fix function naming of functions inside traits


rodrodros:
Functions inside WarpMemory, are being called this way:
```rust
let x = warp_memory.get_or_create_id(...)
```
Because, currently warp_memory is invisible to the solidity AST, the `warp_memory.` prefix is included in the naming. So for the example above, `get_or_create_id` is actually called `warp_memory.get_or_create_id`.

Two possible solutions are including WarpMemory to the AST, and making all function calls part of a member access, or keep it hidden. The second one is fastest, and easy (it would be cool to implement it in a non hacky way)

PR #1075, opened 2023-05-25
Fix warpmemory function naming and implicits + other minor stuff

### LINKED ISSUE #1063: Correctly annotate implicits inside traits
Functions inside the WarpMemory trait are not being set to require the warp_memory implicit.

rodrodros:
Fixes #1063  #1064 

Also fix other minor bugs found in Warp while updating

Add a compilation test

----

The reason this PR haven't been merged is because the existence of an unexpected behaviour on the cairo compiler after the compilation stage. By compiling the test at tests/compilation/contracts/warpMemorySystems/implicitPropagation.sol we get:

```
error: Illegal bigint value during serialization.
```

It is odd because all the independent components used by the transpiled contract have been tested.
We should look help from Starknet to see what is causing this error, or expect that a future release of Cairo won't have this behaviour


rodrigo-pino:
The reason this PR haven't been merged is because the existence of an unexpected behaviour on the cairo compiler after the compilation stage. By compiling the test at _tests/compilation/contracts/warpMemorySystems/implicitPropagation.sol_ we get:
```bash
error: Illegal bigint value during serialization.
```
It is odd because all the independent components used by the transpiled contract have been tested. 
We should look help from Starknet to see what is causing this error, or expect that a future release of Cairo won't have this behaviour

PR #1077, opened 2023-05-30
Do not include starknet accounts file in commits

piwonskp:
Changed its location to `tests/cli/` since the file is in `.gitignore`

### DIFF
--- a/tests/behaviour/behaviour.test.ts
+++ b/tests/behaviour/behaviour.test.ts
@@ -33,16 +33,18 @@ interface AsyncTestCluster {
   dependencies: Map<string, string[]>;
 }
 
-const accountDir = '.starknet_accounts_devnet';
-fs.mkdirSync(accountDir, { recursive: true });
-fs.writeFileSync(path.join(BASE_PATH, accountDir, 'starknet_open_zeppelin_accounts.json'), '{}');
-const address = execSync(`${starknetCliCall('new_account', '')} | awk 'NR==1 {printf $3}'`);
+fs.writeFileSync(path.join(BASE_PATH, 'tests', 'cli', 'starknet_open_zeppelin_accounts.json'), '{}');
+const new_account = `${starknetCliCall('new_account', '')} | awk 'NR==1 {printf $3}'`;
+console.log(new_account);
+const address = execSync(new_account);
 console.log(`New address: ${address}`);
-const addFunds = `curl ${DEVNET_URL}/mint -H "Content-Type: application/json" -d '{"address": "${address}", "amount": 1000000000000000000}'`;
+const addFunds = `curl ${DEVNET_URL}/mint -H "Content-Type: application/json" -d '{"address": "${address}", "amount": 1000000000000000000000000}'`;
 console.log(addFunds);
 console.log(execSync(addFunds).toString());
 console.log('Deploying account:');
-const account_deployed = execSync(starknetCliCall('deploy_account', ''));
+const deployAcc = starknetCliCall('deploy_account', '')
+console.log(deployAcc)
+const account_deployed = execSync(deployAcc);
 console.log(account_deployed.toString());
 
 // Transpiling the solidity files using the `bin/warp transpile` CLI command.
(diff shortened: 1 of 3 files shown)


tests/behaviour/behaviour.test.ts
```diff
@@ -33,16 +33,18 @@ interface AsyncTestCluster {
   dependencies: Map<string, string[]>;
 }
 
-const accountDir = '.starknet_accounts_devnet';
-fs.mkdirSync(accountDir, { recursive: true });
-fs.writeFileSync(path.join(BASE_PATH, accountDir, 'starknet_open_zeppelin_accounts.json'), '{}');
-const address = execSync(`${starknetCliCall('new_account', '')} | awk 'NR==1 {printf $3}'`);
+fs.writeFileSync(path.join(BASE_PATH, 'tests', 'cli', 'starknet_open_zeppelin_accounts.json'), '{}');
+const new_account = `${starknetCliCall('new_account', '')} | awk 'NR==1 {printf $3}'`;
+console.log(new_account);
```
temyurchenko, line 38: Do we want that? If we show it for user information, do we want to add some text explaining what it is?

temyurchenko (approved):

tests/behaviour/behaviour.test.ts
cicr99, line 38: Is this necessary? I think you can remove it

tests/behaviour/behaviour.test.ts
```diff
@@ -33,16 +33,18 @@ interface AsyncTestCluster {
...
+const address = execSync(new_account);
 console.log(`New address: ${address}`);
-const addFunds = `curl ${DEVNET_URL}/mint -H "Content-Type: application/json" -d '{"address": "${address}", "amount": 1000000000000000000}'`;
+const addFunds = `curl ${DEVNET_URL}/mint -H "Content-Type: application/json" -d '{"address": "${address}", "amount": 1000000000000000000000000}'`;
 console.log(addFunds);
 console.log(execSync(addFunds).toString());
 console.log('Deploying account:');
-const account_deployed = execSync(starknetCliCall('deploy_account', ''));
+const deployAcc = starknetCliCall('deploy_account', '')
+console.log(deployAcc)
```
cicr99, line 46: Same with this one

tests/behaviour/behaviour.test.ts
```diff
@@ -33,16 +33,18 @@ interface AsyncTestCluster {
   dependencies: Map<string, string[]>;
 }
 
-const accountDir = '.starknet_accounts_devnet';
-fs.mkdirSync(accountDir, { recursive: true });
```
cicr99, line 37: Is this not needed anymore?

piwonskp:
@cicr99 @murcake I've added `console.log` because I constantly keep getting errors on those steps so it's useful for debugging. Let's keep it for a while, since the solution is hackish it is better to be verbose

tests/behaviour/behaviour.test.ts
piwonskp: Yes, the location is `tests/cli` so until someone deletes the directory from repository it's not needed

temyurchenko:
> @cicr99 @murcake I've added `console.log` because I constantly keep getting errors on those steps so it's useful for debugging. Let's keep it for a while, since the solution is hackish it is better to be verbose

I'm suggesting to add some context to the debug output, as opposed to just a raw number, or whatever they are. I imagine it would help in debugging or at least reduce initial confusion.

piwonskp:
> > @cicr99 @murcake I've added `console.log` because I constantly keep getting errors on those steps so it's useful for debugging. Let's keep it for a while, since the solution is hackish it is better to be verbose
> 
> I'm suggesting to add some context to the debug output, as opposed to just a raw number, or whatever they are. I imagine it would help in debugging or at least reduce initial confusion.

Actually the output is just a command so you know what command was run thus know which has failed

temyurchenko:
> > > @cicr99 @murcake I've added `console.log` because I constantly keep getting errors on those steps so it's useful for debugging. Let's keep it for a while, since the solution is hackish it is better to be verbose
> > 
> > 
> > I'm suggesting to add some context to the debug output, as opposed to just a raw number, or whatever they are. I imagine it would help in debugging or at least reduce initial confusion.
> 
> Actually the output is just a command so you know what command was run thus know which has failed

Oh yeah, I parsed it as actual bash evaluation accidentally

rodrodros (approved):

PR #1084, opened 2023-06-05
translate int conversions to Cairo 1

temyurchenko:
Instead of generating Cairo functions, we use generic capabilities of
   Cairo 1 to write a single static function in warplib.

Additionally we remove some usages of old functions, but not all of
   them, since they are tied to much into some subsystems, like
   calldata and storage. This will be finished in future work.


temyurchenko:
continuation of #1070

piwonskp (approved):

piwonskp:
Was previously approved in in #1070

PR #1085, opened 2023-06-08
Update cairo 1.1

rodrodros:
This PR updates every compiler used by Warp to Cairo 1.1

In order for the plugin to be executed and to work with Scarb 0.4, the way the Scarb.toml was written was modified in order to include both Starknet and Warp plugins as its dependencies.


warplib/src/conversions/integer_conversions.cairo
```diff
@@ -51,26 +51,6 @@ fn bool_into_felt252(val: bool) -> felt252 {
...
-            // since FELT252_PRIME_LOW is 1.
-            if self.low != 0 {
-                return Option::None(());
-            }
-        }
-        Option::Some(
-            self.high.into() * 0x100000000000000000000000000000000_felt252 + self.low.into()
-        )
-    }
-}
```
piwonskp, line 72: Do warplib tests work? It seems they're not run by the CI. I believe an import of try into implementation should be added

rodrodros: Yes the warplib test run during the compilation stage. The implementation of the `Try` is now on the corelib by default, so need to have it here as well

piwonskp (approved):

